PR #26, opened 2023-06-15
Document interfaces

mlof:
Moved interfaces into Abstractions, added xmldoc for declared interfaces


AsakusaRinne:
That's great! Thank you!

PR #32, opened 2023-06-19
Remove native libraries from LLama.csproj and replace it with a targets file.

drasticactions:
The `LLama.csproj` deploys its native runtime dlls and dylibs into the output directory of whatever project references is. This differs from the NuGet, where the runtime files are not packaged. If you wanted to use this repo as a submodule into another project, you can't reference LLama.csproj directly because it's packaging those native libraries there.

Ideally, the same targets files you're shipping on nuget would be in this repo. Until then, I have a PR that can help fix this for those wanting to use the source directly. I removed the native libraries from `LLama.csproj` and put them into a targets file, and changed the example projects to reference that instead. If you don't want to use the runtime files, you can set `IncludeBuiltInRuntimes` to `False` and they won't be bundled, letting you bundle your own.

This does not affect the NuGet, as I can't find how you're building them in this repo. I'm assuming you're doing that locally with your own scripts.


AsakusaRinne:
Great work! Thanks a lot for your contribution!

> This does not affect the NuGet, as I can't find how you're building them in this repo. I'm assuming you're doing that locally with your own scripts.

Yes, I build them locally by manually changing the files with `Nuget Package Explorer`. BTW, could you please give some guides about how to config the backend package settings so that it would be generated when building the project? (I'm not familiar with the MSBuild but I'll learn it if you'd like to give some guides, many thanks!)

Issue #37, opened 2023-06-22
build error on web api


manipuraco:
Hi
I just took latest code from main branch and getting this build error. Can anyone advice. Thanks in advance

Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0311	The type 'LLama.LLamaModel' cannot be used as type parameter 'T' in the generic type or method 'ChatSession<T>'. There is no implicit reference conversion from 'LLama.LLamaModel' to 'LLama.OldVersion.IChatModel'.	LLama.WebAPI	C:\Users\conta\source\repos\LLamaSharp\LLama.WebAPI\Services\ChatService.cs	8	Active

This is the code impacted
```
LLamaModel model = new(new LLamaParams(model: @"ggml-model-q4_0.bin", n_ctx: 512, interactive: true, repeat_penalty: 1.0f, verbose_prompt: false));
_session = new ChatSession<LLamaModel>(model)

```

AsakusaRinne:
Hi, that's because v0.4.0 introduced break changes. I haven't been free to upgrade the web api project yet. Could you please have an another try with lower version?

martindevans:
Was this resolved by #39?

PR #39, opened 2023-06-24
update webapi example

xbotter:
- fix webapi example 
- add history chat example 
- add stream example


xbotter:
It can probably solve issue #37.

PR #47, opened 2023-07-08
MacOS metal support

SignalRT:
That replaces de libraries with the current version with Metal support.


AsakusaRinne:
Thank you for your contribution!

PR #51, opened 2023-07-20
fix breaking change in llama.cpp; bind to latest version llama.cpp to…

fwaris:
… support new quantization methods

latest version of llama.cpp has a breaking api change.

Added windows versions of llama.cpp libs for cpu and cuda11.

Tested with llama2 model weights (see updated readme for model link) - by manually copying dlls to .nuget package locations. 

Need instructions/scripts for building 'backend' packages.


AsakusaRinne:
Thanks a lot for your contribution!

PR #53, opened 2023-07-20
XML docs fixes

martindevans:
Various minor XML docs fixes:
 - Removed parameters from XML that do not exist
 - Fixed invalid XML syntax (e.g. replaced  `<=` with `&gt;=`)
 - Added some comments on uncommented items (lots and lots of these left to do)


AsakusaRinne:
Thank you for the contribution!

PR #54, opened 2023-07-20
Fixed Multiple Enumeration

martindevans:
Fixed multiple cases where an `IEnumerable<T>` was enumerated multiple times.


LLama/Common/FixedSizeQueue.cs
```diff
@@ -30,14 +30,24 @@ public FixedSizeQueue(int size)
         /// <param name="data"></param>
         public FixedSizeQueue(int size, IEnumerable<T> data)
         {
+            // Try an early check on the amount of data supplied (if possible)
+            if (data.TryGetNonEnumeratedCount(out var count))
+                throw new ArgumentException($"The max size set for the quene is {size}, but got {count} initial values.");
+
+            // Size of "data" is unknown, copy it all into a list
```
AsakusaRinne, line 37: Hi, thank you for very much for your contribution. Could you please tell me about why size of `data` is unknown here? I thought the size was got via `data.count()`.

martindevans: Every time you enumerate the `IEnumerable` it's doing all the work necessary to generate the results.

When you call `.Count()` it does all that work to generate the results, counts them, and then immediately throws away all the other results. Then when you do `_storage = new(data);` it has to do all that work to enumerate it _again_ (this time using the results).

Calling `data.TryGetNonEnumeratedCount()` is an optimisation to _try_ and get the count _without_ doing the enumeration (for example if it's an array you can get the count directly). If that fails then rather than doing all the work twice it does it once (to create the list) and then checks the count.

You can see an example of that [here](https://dotnetfiddle.net/ZTmtT6). If you run `OldMethod` you will see the enumeration done twice, if you swpa to `NewMethod` it is only done once.

AsakusaRinne: I got it! Thank you for your explanation! After merging another PR, there's a conflict now. Could you please help to resolve it?

martindevans: It should be fixed now :)

PR #56, opened 2023-07-21
Memory Mapped LoadState/SaveState

martindevans:
Replaced `SaveState` and `LoadState` implementations. These new implementations map the file into memory and then pass the pointer directly into the native API. This improves things in a few ways:
 - Most importantly it fixes a bug: A C# array cannot exceed 2,147,483,591 bytes. In my own use of LlamaSharp I encountered this limit causing a crash.
 - This saves an extra copy of the entire state data into a C# `byte[]`, so it should be faster.
 - The file is trimmed to size, so it saves disk space.

I've included wrappers which expose the old NativeApi method signatures, so this should be a completely backwards compatible change.

This does _not_ fix some other places where `GetStateData` is used. I'll look at those in a separate PR (#57)

### DIFF
--- a/LLama/LLamaModel.cs
+++ b/LLama/LLamaModel.cs
@@ -1,13 +1,11 @@
 ﻿using LLama.Exceptions;
 using LLama.Native;
-using LLama.OldVersion;
-using LLama.Extensions;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


AsakusaRinne:
Great work! Thanks you!

PR #59, opened 2023-07-23
Instruct & Stateless web example implemented

saddam213:
Instruct chat page
Stateless chat page

Max concurrent model now enforced (per model)


AsakusaRinne:
Thanks a lot for your contribution!

PR #63, opened 2023-07-24
Fixed More Multiple Enumeration

martindevans:
Fixed 2 more "multi enumeration" issues.

There are many more of these same issues highlighted by code analysis, but I think all of the remaining ones come from the same root source (tokenization returns an `IEnumerable<int>`) so I'm probably going to try to tackle that in a separate PR.


AsakusaRinne:
Great! Thank you for your contribution!

PR #70, opened 2023-07-28
Multiple context per model support

saddam213:
This builds on top of PR #64 as that brought in the new native API changes needed to allow multiple contexts per model instance

Due to some unfortunate naming there are quite a few files changes, but I did that ins a seperate commit to make review a bit less painless

LLamaModel was renamed to LLamaModelContext and has very little changes, it now just takes a model as a parameter

Basic flow now
LLamaModel -> LLamaModelContext -> ILLammaExecutor

All examples should be updated and working, and the Web interface is activly using mutiple contexts, just open more tabs


martindevans:
Thanks for doing this, I was planning to investigate something similar once my other PR was merged. LGTM 👍 

In future PRs I'm hoping to push some of what's done in the `LLamaModelContext` down into the `SafeLlamaContextSafeHandle`, with the intent that anything you can do with a `llama_context` in llama.cpp can be done with a `SafeLlamaContextSafeHandle` in LlamaSharp. I don't think this PR blocks that change though.

PR #73, opened 2023-07-29
Fixed netstandard2 build

martindevans:
Previous changes to FixedSizeQueue were not compatible with netstandard2.0, fixed that.


martindevans:
Included this change to #74

PR #74, opened 2023-07-30
GitHub actions

martindevans:
Added some simple CI which builds the project and runs tests in debug and release mode on Linux, Windows and MacOS.

I've added an action to download a model into the test project as part of the build. This file is **not** comitted to the repo, but it's cached in the CI so it should be fast after the first time.

This also incorporates a fix to the netstandard2.0 build (formerly #73) so that the CI doesn't immediately fail!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,26 +17,32 @@ jobs:
           - build: linux-debug
             os: ubuntu-latest
             config: debug
+            dotnet: 6.0.x
           - build: linux-release
             os: ubuntu-latest
             config: release
+            dotnet: 6.0.x
           - build: macos-debug
             os: macos-latest
             config: debug
+            dotnet: 6.0.x
           - build: macos-release
             os: macos-latest
             config: release
+            dotnet: 6.0.x
           - build: windows-debug
             os: windows-2019
             config: debug
+            dotnet: 6.0.x
           - build: windows-release
             os: windows-2019
             config: release
+            dotnet: 6.0.x
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


martindevans:
I've added a single test which simply loads and disposes a model (previously there were no tests). I'm not sure why this is failing on Linux and MacOS, probably an issue with the way the test project was set up but I'm not sure what.

martindevans:
Someone on Discord (jlsantiago) did some investigation into why this is failing on Mac. No conclusion yet.

Here's the conversation for reference: https://discord.com/channels/1106946823282761851/1106947264938790972/1137050602711560252

saddam213:
It looks like there was some issues Marsaling the tensor_split affecting Linux and Windows

I have commited a fix [2b5efa0](https://github.com/saddam213/LLamaSharp/commit/2b5efa060dd798cd0be22c2bffd1f9a1c95cba73) that should work on all platforms, can confirm Winows and Linux, but I dont have a recent compilation of the Mac lib to test

However 99% sure this will also work for mac also

martindevans:
@saddam213 thanks for investigating this 👍 

Would you mind opening a PR into my form for this fix?

saddam213:
I think this branch predates the tensor_split change in llama.cpp

martindevans:
> I think this branch predates the tensor_split change in llama.cpp

@saddam213 I've rebased this branch onto master, so it should be fully up to date now.

martindevans:
I'm going to merge this PR and the MacOS binaries update by @SignalRT (#77). After that we'll have CI running and we can see what the status is with cross platform support and binary versions etc.

PR #76, opened 2023-08-05
Multiple context layers per model

saddam213:
This gives the ability to open multiple contexts ontop of a loaded model

The model/context split was done with the least amount of changes to the current implemtation so this may not be the best/final solution, but is a good drop in place update to keep in-line with llama.cpp

- Console examples have been updated however they do not make use of multiple contexts
- Web interface has been updated to use multiple contexts, one per websocket

I accedentally merged #74 into this branch, so might be worth waiting for that to be closed (sorry about that)


martindevans:
Now that all the other PRs are merged could you rebase this onto master? Thanks

Also I noticed this includes some binaries, which version are they exactly?

AsakusaRinne:
It's really a big work. Thank you for your contribution! However I think adding an abstraction `LLamaModelContext` may be a little confusing. A dictionary is introduced to keep multiple models in a `LLamaModel` object instead of reducing the memory cost. In the current master branch, `LLamaExecutor` is used as a context to run the inference of the model. Therefore, if I wasn't misled, there're some overlaps between the concept of `LLamaModelContext` and `LLamaExecutor`. Could you please tell us more about your idea?

saddam213:
Yeah modifying the Executor may be a better option.

I will close this off so  better implemetation can be done via the Executor base class

PR #77, opened 2023-08-05
Update llama.cpp binaries to 5f631c2 and align the LlamaContext 

SignalRT:
It also includes some changes to compile in NetStandard and some typo.


martindevans:
Thanks for this. I resolved the conflicts and merged it. With this the CI should now pass on MacOS and fail on Windows+Linux (due to incorrect binaries for those platforms).

PR #79, opened 2023-08-05
ModelParams & InferenceParams abstractions

saddam213:
Abstacting out these 2 classes will make it a lot easier for UI front ends to handle

- Projects like the web will need to add validation attributes etc, base classing wont work here
- An interface will ensure new properties are propagated when added.

I think this is a cleaner solution than duplicating models and using mappers etc


martindevans:
@saddam213 could you rebase this onto master? That should pull in the new CI stuff (and will hopefully fix the CI failures).

martindevans:
Thanks 👍

PR #80, opened 2023-08-06
Expose some native classes

saddam213:
Having access to these classes outside the hosted project would be very helpful while we are going though rapid deveopment stages.


martindevans:
Thanks for this 👍 

In the longer term I think leaving the sampling API exposed is fine but I'd like to clean up `Utils` a bit. Which parts of that are you wanting to use?

saddam213:
Im just working on some multi context stuff, just need access to everything right now :p

PR #81, opened 2023-08-06
Improved Tensor Splits

martindevans:
Replaced `nint` with `float[]?` in Model params, which is much more user friendly!

This is an improvement on work done in #64 to make it compatible with MacOS. @SignalRT would you mind testing this again on MacOS?


SignalRT:
At least in a smoke test it works OK.

martindevans:
Thanks for testing that.

PR #92, opened 2023-08-08
Abstract executor state saving

saddam213:
Abstact out the executor state saving code, will allow state saving to other media like SQL, Redis

- Basic Interfaces added
- JsonFileStateService set as default
- Left old methods in placve marked as obsolete

This only takles the executor state for now, the model state format will need more looking into so I left the server naming a bit more generic as this will also be added to this service


martindevans (commented):
I didn't mean to start this review

martindevans:
Rather than storing the "StateService" in the executor could it be passed into the Load/Save methods?

That means you wouldn't need to pass the "stateFile" or the "stateService" into the constructor and it's one less bit of state inside the executors to worry about. More importantly it also means you could choose to save an executor to a different place later, rather than binding it in the constructor.

Something like:

```csharp
var executor = new InstructionExecutor(etc...);

executor.Save(new FolderStateSave(a_file_path));
executor.Save(new DbStateSave(db_connection));
executor.Save(new StreamSave(new MemoryStream());

etc
```

saddam213:
Yeah, After sleeping on this I think its probably better to leave this to a higher level library.

The ExeutorBase already exposes ExecutorBaseState load and save, so that can just be wrapped in a class outside LLamaShsrp

Most people will probably use JSON anyway and database is probably more of an edge case for vector query nerds like me, lol

PR #93, opened 2023-08-08
feat: add the api to get the embedding length of the model.

AsakusaRinne:
Add an api to get the embedding length of different models.


martindevans:
I have already made a similar change in my proposed changes to support multi-context (https://github.com/SciSharp/LLamaSharp/pull/90) btw.

AsakusaRinne:
> I have already made a similar change in my proposed changes to support multi-context (#90) btw.

That's great! I'll close this PR.

PR #95, opened 2023-08-10
Add native logging output

saddam213:
llama.cpp commit added output of the log to a callback 
[ea04a4c](https://github.com/ggerganov/llama.cpp/commit/ea04a4ca1940d92becc0ee26523aa2c4a18cf938)

Added native methods and a way to register though the existing Logger, although probably a temp solution

Order the current LogEnum to match to save making a duplicate. 

Edit:
Ugggggh.. I think we have different tab sizes, do we have a preference?

### DIFF
--- a/LLama/Common/Logger.cs
+++ b/LLama/Common/Logger.cs
@@ -16,7 +16,7 @@ public interface ILLamaLogger
         Info = 4
     }
     /// <summary>
-    /// Write the log in cosutomized way
+    /// Write the log in customized way
     /// </summary>
     /// <param name="source">The source of the log. It may be a method name or class name.</param>
     /// <param name="message">The message.</param>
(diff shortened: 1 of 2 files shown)


LLama/Common/Logger.cs
```diff
@@ -9,10 +10,10 @@ public interface ILLamaLogger
...
-        Debug,
-        Warning,
-        Error
+        Debug = 1,
+        Error = 2,
+        Warning = 3,
+        Info = 4
     }
     /// <summary>
     /// Write the log in cosutomized way
```
martindevans, line 19:
```suggestion
    /// Write the log in customized way
```

martindevans:
Oh that's great, I was actually planning to investigate what it would take to replace the llama.cpp logging with a callback exactly like this!

> Ugggggh.. I think we have different tab sizes, do we have a preference?

Yeah indentation seems to be very messed up in this PR.

Visual studio uses 4 spaces, so let's stick to that.

martindevans:
Thanks!

PR #97, opened 2023-08-10
Embedder Test

martindevans:
Added a test for the LLamaEmbedder.

This currently shows different results on Linux, Windows and MacOS.


martindevans:
Something is wrong with embeddings at the moment. Here are the embeddings for "cat" generated by CI:

 - Linux: -0.09917596,-0.71790683,-0.008531962,-0.9898389,-0.66339684...
 - Windows: -0.12730388,-0.67805725,-0.08524404,-0.9569152,-0.6386326...

They're not the same on different platforms!

martindevans:
I don't really have any good ideas for debugging/fixing this :/

martindevans:
On my PC, using the [embedding.exe](https://github.com/ggerganov/llama.cpp/tree/master/examples/embedding) from llama.cpp, on [this version](https://github.com/ggerganov/llama.cpp/releases/tag/master-e59fcb2):

Windows: -0.127304, -0.678057, -0.085244, -0.956915, -0.638633 (same as CI)
Linux (WSL): -0.127304, -0.678057, -0.085244, -0.956915, -0.638633 (**different to CI**)

So this is either a bug with LLamaSharp, or a bug with llama.cpp which has been fixed since we last updated the LLamaSharp binaries.

saddam213:
Did some tests on my Linux server

UBUNTU-22 - Packaged llama.cpp
Expected: -0.0991759598
Actual:   -0.127304003


UBUNTU-22 - Latest llama.cpp
Expected: -0.0991759598
Actual:   -0.127304003

Let me know if you would like tests or anything run on native Linux

saddam213:
I have also double checked the Marshaling of LLamaContextParams fields on both platforms to make sure its not a model paremeters parsing issue. Both platforms are identical

saddam213:
Looks like it may be an upstream issue, running llama.cpp embeddings on native linux produces different results to windows

Ubuntu-22 - llama.cpp - embeddings - cat
`./embedding -m ~/llama-2-7b-chat.ggmlv3.q3_K_S.bin -p cat`
-0.099176 -0.717907 -0.008532 -0.989839 -0.663397

Windows Server 2022 - llama.cpp - cat
`embedding.exe -m llama-2-7b-chat.ggmlv3.q3_K_S.bin -p cat`
-0.127304 -0.678057 -0.085244 -0.956915 -0.638633

martindevans:
Aha, thanks for the investigation. I'm glad it's not our bug in the end! Can you report that to them and link the issue here so we can track the fix?

saddam213:
yeah, just digging though the open issues now to make sure its not already known

martindevans:
I just realised my test was totally broken for a couple of reasons:
 - I used `+` instead of `*` in the dot product 🤦‍♂️ 
 - The comparison was backwards (similar = 1, dissimilar = 0) 🤪

martindevans:
I've loosened up the requirements on the test (epsilon of `0.08f`!) to make it pass on CI, just so we can get this merged in for now.

PR #108, opened 2023-08-19
renamed some arguments in ModelParams constructor so that class can be serialized easily

erinloy:


martindevans:
I assume the serializer needs the constructor parameters to be the same name as the property?

erinloy:
> I assume the serializer needs the constructor parameters to be the same name as the property?

Yes. I'm reading from appsettings.json using standard IConfiguration.GetSection(...)

martindevans:
Thanks for the PR!

PR #115, opened 2023-08-22
ModelsParams record class

martindevans:
Converted ModelParams into a `record` class. This has several advantanges:
 - Equality, hashing etc all implemented automatically
 - Default values are defined in just one place (the properties) instead of the constructor as well
 - Added test to ensure that serialization works properly

Also added tests to ensure the ModelParams class can be round tripped through JSON using both System.Text.Json and Newtonsoft.

### DIFF
--- a/LLama.Examples/NewVersion/ChatSessionStripRoleName.cs
+++ b/LLama.Examples/NewVersion/ChatSessionStripRoleName.cs
@@ -10,7 +10,12 @@ namespace LLama.Examples.NewVersion
             var modelPath = Console.ReadLine();
             var prompt = File.ReadAllText("Assets/chat-with-bob.txt").Trim();
 
-            var parameters = new ModelParams(modelPath, contextSize: 1024, seed: 1337, gpuLayerCount: 5);
+            var parameters = new ModelParams(modelPath)
+            {
+                ContextSize = 1024,
+                Seed = 1337,
+                GpuLayerCount = 5
+            };
             using var model = LLamaWeights.LoadFromFile(parameters);
             using var context = model.CreateContext(parameters);
             var executor = new InteractiveExecutor(context);
--- a/LLama.Examples/NewVersion/ChatSessionWithRoleName.cs
+++ b/LLama.Examples/NewVersion/ChatSessionWithRoleName.cs
@@ -10,7 +10,12 @@ namespace LLama.Examples.NewVersion
             var modelPath = Console.ReadLine();
             var prompt = File.ReadAllText("Assets/chat-with-bob.txt").Trim();
 
-            var parameters = new ModelParams(modelPath, contextSize: 1024, seed: 1337, gpuLayerCount: 5);
+            var parameters = new ModelParams(modelPath)
+            {
+                ContextSize = 1024,
+                Seed = 1337,
+                GpuLayerCount = 5
+            };
             using var model = LLamaWeights.LoadFromFile(parameters);
             using var context = model.CreateContext(parameters);
             var executor = new InteractiveExecutor(context);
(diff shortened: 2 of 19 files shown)


martindevans:
@erinloy I know you've made a couple of PRs to do with serialization recently, just wanted to check this doesn't break anything you're doing? I think it should be more robust long term, since serialization no longer relies on the naming of constructor parameters.

saddam213:
Im not sure we should be adding anything Json serialize specific to the default `ModelParams` class, if your serializing/deserializing to json you should be creating a record/class that implements the appropriate interface like `IModelParams`.

This way allows you to add the necessary Json attributes for validation etc, because its inevitable a converter or validation rule will be needed at some point, for example IInferenceParams contains non serializable properties, so would need a wrapper class to add ignore attributes.

erinloy:
Thans for thinking of me.  I've been loading the ModelParams from appsettings.json, but it's just for convenience to avoid some typing.  It's not a problem to make an adapter.

martindevans:
@saddam213 Are you thinking of this from a future-proofing perspective? i.e. just in case we ever add something non-serializable to `ModelsParams` we shouldn't support serialization right now? If so that is a fair point.

If we decide we don't want to support encoding the `ModelParams` I could actually make use of that right now by replacing the `string Encoding` property with an `Encoding encoding` property.

saddam213:
I think the default class implementation of the Interfaces in the Common namespace should be dumb classes, and should just be there for people to use if they are using base functionality.

This will allow us to add proper typed functionality like you mentioned with `Encoding` without needing to use a Json, Yaml, Xml of DB friendly types, the interfaces are accepted everywhere so the dumb classes are really just convenience,

This is all just a personal opinion on the design, I just feel it could get messy in the main nuget package if we thy support a range of file formats for the end user.

Flip side of this would be to move to make all setting/config type classes totaly serilizable. (ModelParams, InferenceParams etc)

martindevans:
I do take your point, and I'm really not sure where I stand on it at the moment to be honest so take the rest of this as me "thinking out loud"...

I do agree it would be a mess to try and support JSON+YAML+XML+DB etc etc. On top of that supporting serialization ties our hands in the future since we need to make sure any properties that are added (such as `Encoding`) are serializable. My original thinking was if we provide serialization to JSON that realistically probably covers 80% of the use cases, and since it's so easy to do (just 1 attribute and 1 private constructor) then it would be worth it.

Just as an experiment I've updated the PR to have a properly typed `Encoding` property. Since that's thoretically one of the advantages of dropping serialization support. As part of that I updated the tests to see how difficult it would be to serialize in this new setup - it's actually pretty simple to write the converters.

saddam213:
At the end of the day its not a big deal either way, just trying to think of ways this nuget package will be consumed and how far the core pakage should go in terms of stuff like this

wondering if more projcts are required to create more nugets like for example `LLamaSharp,Json`  which could contain all the relevant classes/services for Json, then packages could be made by us or others for savages that use stuff like Yaml

I guess this is more of a question for @AsakusaRinne reagrding how LLamaSharp will be presented nuget wise.

martindevans:
Just modified this PR again with some more experiments with converters. This shows that serialization with `System.Text.Json` support can be made totally seamles (**if we want that!**).

I didn't want to go so far as to ship custom serialization converters as public classes etc (potentially adding lots of extra "API surface"). But if we can do it like this (an internal JsonConverter for types which need it) I think it's worth doing, it's a small amount of code (and no change to the public API) to add support for the most common serializer.

Also I just wanted to check - the real main change in this PR is the change to a `record class` and all the better initialization/equality/hashing etc that brings (serialization was just a side thing!). Does that particular stuff all seem good to you @saddam213?

saddam213:
yeah no issue with the change to record

martindevans:
I'm going to merge this for now, to get the `record` improvements merged in.

With regards to serialization this keeps things roughly as they are - it its current state it is _possible_ to serialize `ModelParams` pretty much as before. There's definitely a worthwhile debate over what we want our longer term serialization story to be, and of course we can follow up with other PRs to remove that if necessary.

PR #116, opened 2023-08-23
Changed type conversion

zombieguy98:
Removed the Utils methods for converting types from the [.Net Framework Fix Commit](https://github.com/SciSharp/LLamaSharp/commit/10f88ebd0e982fab8349d9816631bfaae7e179ce), on further reflection I think it makes more sense to do this where it was done before and using the System.Convert class ensures consistency, this also ensures any systems that set a negative 8bit int for true are handled correctly.

### DIFF
--- a/LLama/Native/LLamaContextParams.cs
+++ b/LLama/Native/LLamaContextParams.cs
@@ -77,8 +77,8 @@ namespace LLama.Native
         /// </summary>
         public bool low_vram
         {
-            get => Utils.SignedByteToBool(_low_vram);
-            set => _low_vram = Utils.BoolToSignedByte(value);
+            get => Convert.ToBoolean(_low_vram);
+            set => _low_vram = Convert.ToSByte(value);
         }
         private sbyte _low_vram;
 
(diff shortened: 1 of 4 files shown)


martindevans:
Good idea, I think this is definitely neater than before 👍 

Is there a reason this is using signed bytes instead of bytes? You mentioned some systems setting negative values for true, which is... odd!

zombieguy98:
> Is there a reason this is using signed bytes instead of bytes? You mentioned some systems setting negative values for true, which is... odd!

I'm mainly just been making sure to stick to the original marshalling version which was applying the [MarshalAs(UnmanagedType.I1)] attribute, which in the docs states: `A 1-byte signed integer. You can use this member to transform a Boolean value into a 1-byte, C-style bool (true = 1, false = 0).`.

Bools are a weird type that are "implementation-defined" in the C++ standard, afaik most modern compilers use the false = 0, true = 1 format converting the other way the C++ standard says: `A zero value, null pointer value, or null member pointer value is converted to false;
any other value is converted to true` so sbyte should be a best fit.

martindevans:
Thanks for the explanation 👍

Issue #123, opened 2023-08-25
GitHub Action Compiling Dependencies


martindevans:
Rather than hand building the dependencies for every release we could set up a system that builds them all in a GitHub action.

We could:
1. Have a file that defines the commit id
2. Have an action that pulls the main llama.cpp repo at that commit
3. Runs builds for Windows, Linux and MacOS
4. Packages the binaries into the right location and opens a PR on this repo
5. Auto run that action every week

This would mean that updating to a new version of llama.cpp would be much easier and done more frequently. The PR would be auto opened and CI would run to check if it works, if it doesn't then it's obvious someone needs to modify the PR to fix things.

martindevans:
WIP: https://github.com/martindevans/LLamaSharp/tree/cron_job

AsakusaRinne:
I'm late to see this good idea. Is there any help needed for the binary release? BTW, I remember that github action only has intel MAC instead of M1/M2 MAC. Is there already a fix for it?

martindevans:
> I'm late to see this good idea. Is there any help needed for the binary release?

I've done most of the work for the compiling of binaries in the new GitHub action (currently only manually triggered). This pulls in the master branch from llama.cpp and builds for:
 - Windows
 - Windows+CUBLAS
 - Linux
 - MacOS (ARM64)
 - MacOS (x86_64)

I have _not_ yet done a metal build for MacOS. I have no idea how that works, so any help there would be really good.

> BTW, I remember that github action only has intel MAC instead of M1/M2 MAC. Is there already a fix for it?

Unfortunately not, Mac runners are still all x86_64. I've set up the new action to cross compile, so it can create the ARM64 binaries, it just can't run them. That does mean we still have no way to test ARM64 though :(

martindevans:
I just noticed the GitHub action doesn't include cuda .so files either, so that needs doing as well.

martindevans:
I've managed to work out the Linux cuda build, turns out it was basically the same as the Windows one. Fortunately we don't need a GPU for compiling CUDA, although it might be a good idea to run some CI on a GPU server in the future.

Someone [gave me some tips](https://github.com/SciSharp/LLamaSharp/pull/122#issuecomment-1697338329) on how to build for metal, so I will investigate adding that to the GitHub action. But that probably won't be until next weekend.

SignalRT:
[...]

PR #126, opened 2023-08-28
Dependency Building In Github Action

martindevans:
This is the first stage of my idea proposed in #123.

This adds an action that builds all of the dependencies. At the moment the action does not run automatically, but can be manually triggered to produce all binaries.


AsakusaRinne:
I've added a repo secret key `LLAMA_SHARP_NUGET_KEY` which is the key to push packages to nuget. It will support automatically building releases and push them to nuget in the future. (as a memorandum here)

PR #138, opened 2023-08-31
Enable Semantic kernel support

drasticactions:
This PR adds base level [Semantic Kernel](https://github.com/microsoft/semantic-kernel) support. This was partially based on https://github.com/xbotter/semantic-kernel-LLamaSharp with some refactoring from me (the original implementation would create new models for run of the commands. That would cause each command to load a new instance, which wouldn't work)

I also added a new entry point for LLamaEmbedder to accept an existing context, so you wouldn't need to load a new one to get embeddings.

I ported two of the three APIs the original repo had, TextCompletion and ChatCompletion, as well as samples from Semantic Kernel for both. They should function as their examples in the original repo do, although depending on the model used the TextCompletion sample may produce weird results, since the prompt is kinda weak and doesn't have a stop statement. I ran it against OpenAI too and it had similar results depending on the settings. The Chat example is more consistant.

I did not port the embedding support yet. I tried it and it "seemed" to work but when I tried porting the MemorySkill sample that implements it, it couldn't recall any of the embedded documents, even though it was putting "something" into the context. I'm not sure if I did something wrong, or the  `LLamaEmbedder` is incorrect. It's not necessary to at least implement some skills though, so IMO it can wait.


LLama.SemanticKernel/TextCompletion/LLamaSharpTextCompletion.cs
```diff
@@ -0,0 +1,27 @@
+﻿using LLama;
+using LLama.Abstractions;
+using Microsoft.SemanticKernel.AI.TextCompletion;
+
+namespace Microsoft.SemanticKernel.Connectors.AI.LLama.TextCompletion;
```
martindevans, line 5: Should this really be in a `Microsoft.SemanticKernel` namespace instead of something like `LLamaSharp.SenanticKernel`? This seems a bit weird to me, but maybe it's the convention in SemanticKernel.

martindevans:
To be honest I don't know much about SemanticKernel, but I've given this a quick look through and added a couple of comments.

I think @AsakusaRinne did some work on SemanticKernel and may be able to provide a better review than me?

LLama/LLamaEmbedder.cs
```diff
@@ -29,6 +29,11 @@ public LLamaEmbedder(IModelParams @params)
             _ctx = weights.CreateContext(@params);
         }
 
+        public LLamaEmbedder(LLamaContext ctx)
```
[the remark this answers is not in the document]
drasticactions: Changed with https://github.com/SciSharp/LLamaSharp/pull/138/commits/326c802be7694e6d820080a845ebeb7a0fa42ca9

LLama.SemanticKernel/TextCompletion/LLamaSharpTextCompletion.cs
drasticactions: Changed with https://github.com/SciSharp/LLamaSharp/pull/138/commits/2bde188c64cd8b42568aaface7b7cfeb3d4e1648,

I didn't like it either, if ya'll want it back I can change it, but having it be inside this namespace does make more sense and shouldn't affect SemanticKernel itself.

Arlodotexe:
> I did not port the embedding support yet. I tried it and it "seemed" to work but when I tried porting the MemorySkill sample that implements it, it couldn't recall any of the embedded documents, even though it was putting "something" into the context. I'm not sure if I did something wrong, or the LLamaEmbedder is incorrect. It's not necessary to at least implement some skills though, so IMO it can wait.

If you have the code, could you open a draft PR once this one closes? I can take a look 🙂

drasticactions:
> > I did not port the embedding support yet. I tried it and it "seemed" to work but when I tried porting the MemorySkill sample that implements it, it couldn't recall any of the embedded documents, even though it was putting "something" into the context. I'm not sure if I did something wrong, or the LLamaEmbedder is incorrect. It's not necessary to at least implement some skills though, so IMO it can wait.
> 
> If you have the code, could you open a draft PR once this one closes? I can take a look 🙂

Yeah. If you want to play around with it right now,

https://github.com/SciSharp/LLamaSharp/tree/98bfbe1d50dfb124d2c593214d6a48f72ecee0f3/LLama.SemanticKernel/TextEmbedding

https://github.com/SciSharp/LLamaSharp/tree/98bfbe1d50dfb124d2c593214d6a48f72ecee0f3/LLama.SemanticKernel/ChatCompletion

https://github.com/SciSharp/LLamaSharp/blob/98bfbe1d50dfb124d2c593214d6a48f72ecee0f3/LLama.Examples/NewVersion/SemanticKernelMemorySkill.cs

Adding these back to my branch should get it working.

martindevans:
We did find a while ago that llama.cpp returns different results on different platforms (https://github.com/ggerganov/llama.cpp/issues/2582). That might be causing your embedding to act weirdly?

drasticactions:
> We did find a while ago that llama.cpp returns different results on different platforms ([ggerganov/llama.cpp#2582](https://github.com/ggerganov/llama.cpp/issues/2582)). That might be causing your embedding to act weirdly?

Ahhh, that's good to know. I was running this on Windows but I can try it on Linux and Mac and see if it's different.

AsakusaRinne:
LGTM, thank you for this good job! @xbotter any suggestion please?

AsakusaRinne (approved):

AsakusaRinne:
One more thing, could you please add a README file under the folder of `LLama.SemanticKernel`? A brief description of the relationship between LLamaSharp and Semantic Kernel and a link to the examples is enough.

drasticactions:
> One more thing, could you please add a README file under the folder of `LLama.SemanticKernel`? A brief description of the relationship between LLamaSharp and Semantic Kernel and a link to the examples is enough.

Added with https://github.com/SciSharp/LLamaSharp/pull/138/commits/d4a57fffef453014311ca5e2ed6ee122aa168ee8

xbotter:
This is so great! I believe that embedding support can be ported. The effect of embedding is more closely related to the model. We can try using different models to compare the results.

martindevans (approved):

martindevans:
@drasticactions I noticed that the `ChatHistory` in LLamaSharp is marked with the comment `// copy from semantic-kernel`. Now that your semantic kernel support is merged in do you think it makes sense to remove these things from LLamaSharp?

drasticactions:
@martindevans Looking at its usage, I think it's only used for https://github.com/SciSharp/LLamaSharp/blob/master/LLama.WebAPI/Services/StatelessChatService.cs#L25

That could be updated to use SK by default and use its implementation. If that's the only use, that yeah, removing it and using SK instead makes sense to me.

PR #139, opened 2023-09-01
Add TextEmbedding for Semantic Kernel

drasticactions:
Going off of my last PR, this tried to add TextEmbedding for Semantic Kernel to LLamaSharp.

<s>I am having a few issues with this so far: I ported two examples that use Embedding, but they don't work right. None of the memories can be recalled, although I think they are saving. I'm not sure if the issue is with the code I've added, issues with LLamaEmbedder, or llama.cpp itself, or the model I'm using, or a combination of them.</s>

Upon bumping to the newest master code and using a GGUF LLama2 model, the embedding code started to (mostly) work and I could get results from the memory example. It does return results I would expect (Although not one to one with their output, that's fine since these are not the same and the numbers seem to be correct for what they are).

https://gist.github.com/drasticactions/0c90cac41ef890d507fe584f94e3b1a2

I removed the memory skill example as that still doesn't work, but I think that's due to the executor what what's being done behind the scenes in SK for the OpenAI/Azure ones with setting the correct prompts to let it work right. That may require more wrappers to be written, but since this underlying implementation should work, it's fine to add it in now.

cc @Arlodotexe


drasticactions:
@AsakusaRinne @martindevans This should be ready for review.

AsakusaRinne:
Good job! Do you think publishing a package named `LLamaSharp.SK` a good idea? To be honest I only used semantic-kernel for one or two times and am not familiar with its practices. It seems that users would like to build their apps based on this semantic-kernel integration rather than take it as an example.

drasticactions:
> Good job! Do you think publishing a package named `LLamaSharp.SK` a good idea? To be honest I only used semantic-kernel for one or two times and am not familiar with its practices. It seems that users would like to build their apps based on this semantic-kernel integration rather than take it as an example.

I think so. I didn't add the information to the csproj for it to be packaged (I assumed you would want to call it whatever you liked and I didn't want to impose) but with ReadMe I wrote you should be able to publish it as a nuget.

PR #142, opened 2023-09-02
refactor: remove old version files.

AsakusaRinne:


martindevans:
Just a note that we haven't released an update with the `[Obsolete]` markers in place yet. So we may want to delay this until the next version?

AsakusaRinne:
Yes removing old APIs without a notification is not a good choice at most of the stages. However now the old APIs cannot load the model at all because of the updating. The old version hasn't been updated for 3 months but I haven't seen an issue about it. Therefore I believe removing them is a safe behaviour and won't annoy users. Thank you for the reminder😊

martindevans:
Ah I didn't realise the updates had broken the old version. In that case go ahead!

PR #143, opened 2023-09-02
Update MacOS Binaries

SignalRT:
This should work to compile MacOS binaries (CPU and Metal).

Pending to test the binaries.


.github/workflows/main.yml
```diff
@@ -12,29 +12,29 @@ jobs:
...
             os: ubuntu-latest
             config: release
-          # - build: macos-debug
-          #   os: macos-latest
-          #   config: debug
-          # - build: macos-release
-          #   os: macos-latest
-          #   config: release
+          - build: macos-debug
+            os: self-hosted
```
martindevans, line 24: what does this mean?

.github/workflows/compile.yml
```diff
@@ -198,7 +200,12 @@ jobs:
       - name: Rearrange MacOS files
         if: ${{ github.event.inputs.macos }}
         run: |
-          cp artifacts/llama-bin-macos-arm64.dylib/libllama.dylib deps/libllama.dylib
+          mkdir deps/macos-cpu
+          cp artifacts/llama-bin-macos-cpu.dylib/libllama.dylib deps/macos-cpu/libllama-cpu.dylib
```
martindevans, line 204: I don't _think_ this naming will work.

As I understand it since we've marked our methods with `[DllImport("libllama")]` the file that gets loaded needs to be called `libllama`. That's why all of the other files are distinguished by their _folder_ path but have identical names (except for the file extension of course).

martindevans:
Thanks very much for doing this! I wasn't looking forward to doing it myself 😆

.github/workflows/main.yml
martindevans: Did some googling, looks like this means you can self host your own github job runners, which I guess you were using for testing and don't actually want to merge in here?

.github/workflows/compile.yml
drasticactions: If it's for GitHub artifacts, that could be fine. You would need to rename it for the Nuget since it would have to be the same as the DLLImport, but that could be done in a script whenever that gets generated.

.github/workflows/main.yml
SignalRT: Yes it's what I use in my fork and is something that I don't want to merge

.github/workflows/compile.yml
SignalRT: That's right it was the intention

SignalRT:
It seems to be ok now.

martindevans:
Thanks very much!

PR #145, opened 2023-09-03
Disable test parallelism

martindevans:
Disable test parallelism to prevent flakey CI


martindevans:
Unfortunately this doesn't fix CI (windows-release still seems flakey), but it did help locally (running tests could cause a total lockup occasionally)

PR #149, opened 2023-09-06
Removed unused properties of `InferenceParams` & `ModelParams`

martindevans:
I noticed that these properties are unused.

### DIFF
--- a/LLama/Abstractions/IInferenceParams.cs
+++ b/LLama/Abstractions/IInferenceParams.cs
@@ -32,21 +32,6 @@ namespace LLama.Abstractions
 		public IEnumerable<string> AntiPrompts { get; set; }
 
 		/// <summary>
-		/// path to file for saving/loading model eval state
-		/// </summary>
-		public string PathSession { get; set; }
-
-		/// <summary>
-		/// string to suffix user inputs with
-		/// </summary>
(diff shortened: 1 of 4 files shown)


martindevans:
I'm leaving this open for a while, just in case anyone wants to say that actually we should keep these properties and implement them properly.

saddam213:
Wondering if renaming `ModelAlias` to `Name` would be beneficial

Not an issuse for LLamaSharp itself, but if you are dealing with multiple models it could be handy to have a identifier/display name other than filepath

This is also something end users can esily add themselfs, so maybe not

martindevans:
I'm not against adding something like that. I'm only proposing this PR because right now the properties are totally unused, but if you want to open a PR to implement any of these things I'll happily review it.

For the proposed `Name` property are you thinking it would tag the `LLamaWeights` object with the name (public readonly property), to make it easier to identify? That's not a bad idea.

saddam213:
That what I was thinking, but users could just name thier model files a bit clearer and that would also work

And thinking about it more, `Name` would really be more of a field requiered if dealing with multiple models, and in that case users would most likely need an `Id` field too, so they would most likely create thier own classess from the interfaces anyway

Ignore my old man ramblings :p

PR #157, opened 2023-09-08
Replace ILLamaLogger for ILogger implementation

saddam213:
Looking to improve the logging, I think using the Microsoft ILogger framework may be a good move, so just opening this PR to begin discussion.

As this will be used in .NET apps and there are many great ILogger packages out there I thought maintaing our own may be just extra work.

I think just adding a clean way for users to register thier logger, and a way to make logging a bit friendlier around the lib could be good too, not DI or anything but passing loggers in constructors is annoying so perhaps a sneaky LogManager class could help, bit of an anti-pattern but would encourage more devs to log if was easy to setup.

This commit is just a as-is replacement, does nothing to address the current DefaultLogger or how its used.

Thoughts?


saddam213:
Thinking outloud...

If there is a `Configure` method coming like was discussed in one of the Issues, an `ILoggerFactory` could be passed in there

We could then setup default `ILoggerFactory` to fall back too if none is supplied, and that would just return a simple console logger, and drop all the internal FileLog stuff

saddam213:
This commit is the basic idea we could use, like I said the static factory may be a bit ugly, but it does allow for easy inlitalization of a logger thoughout the codebase

Now users can just pass down any ILoggerFactory into the configure and use their own logging framework.

SignalRT:
I was thinking on making the same. I will be happier using ILogger than a specific logger per library (that makes log a mess). 
I'm more happy using DI for the Log and not a factory. 

In my opinion to move the code base to use ILogger is a very good idea.

saddam213:
That scenario is even easier, just pass in the ILogger to the constructor, if you have one, no need for a Logger class at all

LLama.SemanticKernel/LLamaSharp.SemanticKernel.csproj
```diff
@@ -34,6 +34,7 @@
 
 	<ItemGroup>
 	  <PackageReference Include="Microsoft.SemanticKernel.Abstractions" Version="0.21.230828.2-preview" />
+	  <PackageReference Include="System.Memory" Version="4.5.5" />
```
martindevans, line 37: Should `System.Memory` be in a `<ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">` group, same as LLamaSharp?

saddam213:
If we are just going to delete the Logger class and just carry on with passing in loggers into the constructors then I might close this and open another PR to remove all the commit mess of the ILoggerFactory, I only went that route so we have a "Log Anywhere" solution in the codebase

martindevans:
I can just make sure to `Squash and merge` if you like

PR #158, opened 2023-09-08
ILogger implementation

saddam213:
As talked about in PR #157 replace the in-built logger for the Microsoft ILogger framework

1. Update the few places using a logger
2. Move LogLevel enum to native and update callback
3. Remove old logger class

Some nuget updates were required along the way


martindevans:
Should we be taking an `ILogger<Something>` everywhere instead of a plain `ILogger`?

I'm not very familiar with `ILogger`, but it looks like the convention is to take `ILogger<ThisClassName>` in the constructor for `ThisClassName`? I assume that would mean that you don't have to include `[LLamaExecutor]` in the text of the message and it would be automatically categorised by type for you.

saddam213:
We should, however that would force the end user to pass down those typed loggers, that was one of the concerns in my first PR.

The problem is without DI or a Factory pattern the typed loggers are clunky to use.

martindevans:
Ah sorry, I missed that in your first PR.

How "clunky" are we talking? Could you show an example?

saddam213:
> Ah sorry, I missed that in your first PR.
> 
> How "clunky" are we talking? Could you show an example?

pudo code, but example, say context and executor both had typed loggers in the constructor, you would need to create to loggers to do simple tasks

``` 
var contextLogger = MyLoggerFactory.Create<LLamaContext>();
var executeLogger= MyLoggerFactory.Create<InstructExecutor>();

using var model = LLamaWeights.LoadFromFile(parameters);
using var context = model.CreateContext(parameters, contextLogger);
var executor = new InstructExecutor(context, executeLogger);```

using ILogger they can just pass down any logger they have

martindevans:
Oh I was imagining way worse than that! I think that's fine personally.

If people really hate then can easily work around it in the future (e.g. add another constructor that just accepts non-generic ILogger, or one which accepts an ILoggerFactory as you suggested), but I don't think that's likely to be needed.

saddam213:
> Oh I was imagining way worse than that! I think that's fine personally.
> 
> If people really hate then can easily work around it in the future (e.g. add another constructor that just accepts non-generic ILogger, or one which accepts an ILoggerFactory as you suggested), but I don't think that's likely to be needed.

Then nesting also becomes an issue, if Executor creates a class that also has its typed logger, how do we pass it down?

I think it has to be ILogger? or we create the ILogger<T>s and the user provides us thiet ILoggerFacory if they want to

saddam213:
The we could add a global `Configure(ILoggerFactory, Numa, Cuda)` an the user can just register in one place

martindevans:
> Regret opening this can of worms :p

Agreed 😆 

Let's just stick with `ILogger` as you have it in this PR - it's nice and simple and covers most of what we need for now.

I'll merge this if you agree?

saddam213:
yep, happy to megre and run, lol

PR #163, opened 2023-09-10
MacOS default build now is metal llama.cpp #2901

SignalRT:
With this changes I think that it's enought to have only a binary. In this moment I keep the name off the DLL with metal as suffix.

https://github.com/SignalRT/llama.cpp/commit/e36ecdccc8754783f93ad3ac8a09e540101f2ca0

There is other change because .metal file should be in the binaries folder of it doesn't work.


martindevans:
Looks good to me! I can't test this, but if you can confirm it works with and without metal I'm happy to merge it.

geffzhang:
I just verified it on a Mac machine 
218f44183da691f2c244166a76caf7f

drasticactions:
Is the idea here to get rid of the Metal specific Nuget and make this the default in the runtime package?

martindevans:
I think that's the idea yes, since there's only one MacOS binary after this there's no point splitting it into a separate package.

drasticactions:
> I think that's the idea yes, since there's only one MacOS binary after this there's no point splitting it into a separate package.

It's still supported in their codebase. The underlying implementations are still there, but they switched the CMake files to build with metal by default. So there are still two binaries... that said, if the default is Metal, then that would probably make the most sense to ship by default too, so I'm okay with that.

SignalRT:
There is another argument as explained in the documentation:

_**When built with Metal support, you can explicitly disable GPU inference with the `--gpu-layers|-ngl 0` command-line
argument.**_

Oceania2018 (approved):

Issue #184, opened 2023-09-28
Major llama.cpp Change Notes


martindevans:
llama.cpp [3328](https://github.com/ggerganov/llama.cpp/pull/3228#event-10501459551) just got merged. This contains a major change to the API which will require some big changes to LLamaSharp to support.

 - [x] Make a release before this change, with current API
 - [x] Make basic changes to low level API
 - [ ] Adapt current executors to work with new API (minimal changes)
   - I think we're maybe _not_ going to do this, since there's general agreement the executors need completely replacing anyway.
 - [ ] Work out how to expose new capabilities (entirely new executor)

AshD:
I ran into this issue when I wanted to try the new mistral model. 
I updated the llama.cpp dlls and it failed the token generation with current main branch llamasharp branch.

It's tough to keep up with Llama.cpp breaking changes and all the new models coming in daily :-)

martindevans:
That's odd, I actually tried out Mistral-7B-Instruct earlier today using 0.5.1 and it worked perfectly.

AshD:
Thanks Martin. I used the current Llamasharp main branch with the Cuda 11 dll in runtime folder and it worked!

I thought Mistral required the new llama dlls and when I had tried it with yesterday's Llama.cpp dlls, it threw the tokenization error. 

Thanks again, 
Ash

PR #185, opened 2023-09-29
Major llama.cpp API Change

martindevans:
llama.cpp https://github.com/ggerganov/llama.cpp/pull/3228#event-10501459551 just got merged with major changes to the API. See #184 with some notes on this.

This PR has basic changes to match the underlying API changes. Things to do:
 - [x] `LLamaContextParams` and `LLamaModelParams` have been split into 2, we should split `IModelParams` in the same way
 - [x] Fix cloning contexts (removed support)
 - [x] Fix `OutOfContext` test
 - [x] Add binaries for all platforms
   - ~~Built by [this run](https://github.com/SciSharp/LLamaSharp/actions/runs/6496667993) based on [this commit](https://github.com/ggerganov/llama.cpp/commit/9e24cc6e2e589d405bd1720c400f5b0b9d0ca3ee)~~
   - Build by [this run](https://github.com/SciSharp/LLamaSharp/actions/runs/6553124555) based on [this commit](https://github.com/ggerganov/llama.cpp/commit/cb33f43a2a9f5a5a5f8d290dd97c625d9ba97a2f)
   - [ ] Test on all platforms
     - [x] Windows (Martin)
     - [x] Linux
     - [ ] MacOS
       - @SignalRT 🙏 
   - [x] Update binaries and fix tokenizer tests


martindevans:
@saddam213 I've made some changes which have broken some of the LLama.Web stuff at the moment, it looks like it's all places which construct a single context by loading an entirely new set of model weights (probably legacy code?). I can probably throw in some workarounds if you want, or do you want to do something more complete and push it into this PR?

saddam213:
> @saddam213 I've made some changes which have broken some of the LLama.Web stuff at the moment, it looks like it's all places which construct a single context by loading an entirely new set of model weights (probably legacy code?). I can probably throw in some workarounds if you want, or do you want to do something more complete and push it into this PR?

No probs man, I'll get it done tonight, wont take long to get it up to date with your changes :)

martindevans:
@saddam213 I actually just pushed up some workarounds to at least get it building, but I left some todos in place.

martindevans:
> The web project has fallen behind a bit, I need to get it up to date, got lost in the world of stable diffusion the last few weeks, lol

haha I know how that is. I bounce back and forth between LLMs and stable diffusion all the time 😆

saddam213:
wondering if we should merge the Web and WebApi into one project, there is a lot of code that will need to be shared to get them both up to date, and we can just use an appsetting config to decide what it strats up, API or WEB

Or add a shared project, but that seems overkill for examples

I cant even port the LLamaStack api or web cause I went and make custom executor line an idiot :/

saddam213:
Ok, so the changes to get the web up to date with version 0.5.1 are more than I thought. 
(have done a untested prelim change on my master branch) [Commits](https://github.com/SciSharp/LLamaSharp/compare/master...saddam213:LLamaSharp:master)

I think its too big to mash into this PR, so I will create its own PR then we can merge that in, then we can merge this PR's changes overtop

That work for you?

martindevans:
> That work for you?

That's fine by me, there are workarounds in place in this PR so that it at least builds and we can get your PR in right after this one is done.

I'll try to get this one finished and merged soon so we can get on with that :)

martindevans:
The only thing blocking this PR is proper testing on MacOS.

SignalRT:
I will be at home tomorrow. This week I could check this PR

SignalRT:
@martindevans 

At least with today llama.cpp binaries It fail in two test. All related with the tokenizer. 

This could be related with the latest changes in the tokenizer: https://github.com/ggerganov/llama.cpp/pull/3538
, but I cannot test it in Windows today

These are the errors:

Xunit.Sdk.EqualException
Assert.Equal() Failure: Collections differ
              ↓ (pos 1)
Expected: [1, 450, 4996, 17354, 1701, ···]
Actual:   [1, 1576, 4996, 17354, 1701, ···]
              ↑ (pos 1)
   at LLama.Unittest.LLamaContextTests.Tokenize() in ...LLamaContextTests.cs:line 40

And

Xunit.Sdk.EqualException
Assert.Equal() Failure: Collections differ
           ↓ (pos 0)
Expected: [450, 4996, 17354, 1701, 29916]
Actual:   [1576, 4996, 17354, 1701, 29916]
           ↑ (pos 0)
   at LLama.Unittest.LLamaContextTests.TokenizeWithoutBOS() in ../LLamaContextTests.cs:line 48

martindevans:
Thanks for testing that, I'll retrigger the github action to recompile all of the binaries and fix the tokenizer tests.

Did you test with the included binaries at all to confirm that everything else works?

SignalRT:
I didn’t tested it yet with the included binaries, I used my daily build binaries. Tomorrow I will test the binaries generated by the github build and I will make some test with additional projects and some Semantic Kernel examples.

staviq:
> This could be related with the latest changes in the tokenizer: [ggerganov/llama.cpp#3538](https://github.com/ggerganov/llama.cpp/pull/3538) , but I cannot test it in Windows today


Hi, that PR only applies if you explicitly pass "special=true" to `llama_tokenize`, otherwise tokenizer behavior remains unchanged.

Using "special=true" is meant only for "internal" use, and should not be exposed to the end user ( application user )

Enabling special token processing causes the following:
- No space prefix is added to the input string. This is important when prompt format expects `word` instead of ` word`, for example when system prompt expects `system` instead of ` system` which are different tokens.
- Special ( control ) tokens, such as `<s>`, `<|im_start|>` etc. are explicitly matched first and directly converted to appropriate token id.
- Remaining string segments, between previously matched special tokens are tokenized normally.
- Results are merged into one chain of token ids.

This addition to the API is meant for processing things like prompt template, and allows including bos/eos directly in the input string.

**User text should be tokenized normally, with "special=false".**

---
Clarification on "special" tokens:

Model vocabulary naturally has a tree like structure, every normal token can be represented by two other shorter tokens, down to single character.

Tokenizer picks neighboring characters from input string, finds a token that represents that substring, and repeats merging neighbors until no valid longer token can be matched.

Special tokens break that rule, and are chosen to never be "reachable" by merging normal tokens, preventing the user from inserting control tokens in the prompt.

Previously, special tokens were simply tokenized as plaintext ( `<s>` would become something like `<`, `s`, `>` ), and various prompt formats weren't exactly working with llama.cpp.

This change, allows use of special ( control tokens ), directly in the input string, removing the need to fiddle with querying vocabulary manually to make prompt templates work.

martindevans:
@SignalRT I've added new binaries and fixed a few small things (e.g. I think out-of-context handling should be about 100x faster now).

@staviq thanks very much for the extra info! I think the problem in this case was because SignalRT was using newer binaries than this PR was built for I hadn't added that new parameter yet on the C# side! I'm not quite sure what happens in this case (the call was still successful, so I guess whatever junk was on the stack gets used). Once I added the parameter and passed `special=false` the results went back to the usual values.

SignalRT:
It's seems to work but I changed the packages in the Example to be able to run SK examples. This is the change:

  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
    <PackageReference Include="Microsoft.SemanticKernel" Version="0.21.230828.2-preview" />
    <PackageReference Include="Microsoft.SemanticKernel.Reliability.Basic" Version="0.24.230918.1-preview" />
  </ItemGroup>

So for me is OK.

martindevans:
Thanks for testing that!

AsakusaRinne:
It's a great work! Is it a good time to publish a new release?

AsakusaRinne:
And maybe I should add an auto-release to ci, I'll do it this weekend.

martindevans:
> It's a great work! Is it a good time to publish a new release?

I think so yeah, this has made a few breaking changes to the API but nothing massive. Good time to push out a new version 👍

AsakusaRinne:
> I think so yeah, this has made a few breaking changes to the API but nothing massive. Good time to push out a new version 👍

I'm making the auto-release ci now and I think there's something confusing me. There are two supports for MAC, CPU and Metal. However only one `libllama.dylib` is included in `LLama/runtimes`, is it both used for CPU and Metal?

martindevans:
@AsakusaRinne It was removed in this PR: https://github.com/SciSharp/LLamaSharp/pull/163

From the discussion there it looks like one binary supports CPU and Metal inference (but I don't have a Mac so I can't confirm that).

SignalRT:
@AsakusaRinne, one binary supports CPU and GPU.

AsakusaRinne:
@martindevans @SignalRT Thank you for your help! I've opened a PR #204 of the auto-release ci. 😊

PR #193, opened 2023-10-13
Support SemanticKernel 1.0.0-beta1

DVaughan:
This PR brings in support for SemanticKernel 1.0.0-beta1.
The `ChatRequestSettings` class has been removed from SemanticKernel as per https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0008-support-generic-llm-request-settings.md
I've reinstated `ChatRequestSettings` as part of the LLamaSharp.SemanticKernel project and bumped up the nuget package number.
Tests pass.
I'm uncertain if using a `ChatRequestSettings` class is the best strategy. The author may have a better idea about how best to support the new base `AIRequestSettings` class.

### DIFF
--- a/LLama.Examples/LLama.Examples.csproj
+++ b/LLama.Examples/LLama.Examples.csproj
@@ -28,7 +28,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
-    <PackageReference Include="Microsoft.SemanticKernel" Version="0.21.230828.2-preview" />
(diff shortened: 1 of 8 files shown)


martindevans:
@drasticactions since you originally developed the SemanticKernel support would you mind reviewing this? Thanks.

drasticactions (approved):
LGTM, seems to work fine. Hopefully their API stabilizes when it hits 1.0 so we wouldn't need to catch up like this.

martindevans:
Thanks for the PR @DVaughan and thanks for the review @drasticactions :)

Issue #189, opened 2023-10-09
Running LLamaSharp on gpu 


jonty-esterhuizen:
so i am currently using LLamaSharp like this 

```
    public StatefulChatService ( IConfiguration configuration )
        {
            _configuration = configuration;

            modelPath = Path.Combine(_configuration [ "ModelPath" ], _configuration [ "ModelName" ]);
            systemPrompt = File.ReadAllText(_configuration [ "ChatWithBase" ]);
            _context = new LLamaContext(new LLama.Common.ModelParams(modelPath)
            {
                ContextSize = 2048,
                GpuLayerCount = 50,
                MainGpu=0
            });

            _chatSession = new ChatSession(new InteractiveExecutor(_context));
            _chatSession.AddInputTransform(new MyInputTransform1());
            _chatSession.AddInputTransform(new MyInputTransform2());
            _chatSession.AddInputTransform(new MyInputTransform3());
        } 
        public void Dispose ()
        {
            _chatSession.SaveSession(_configuration [ "SavedSessionPath" ]);
            _context?.Dispose();
        }

        public async Task<string> ChatAsync ( SendMessageInput input )
        {
            var userInput = input.Text;
            if (!_continue)
            {
                userInput = systemPrompt + userInput;
                Console.Write(systemPrompt);
                _continue = true;
            }

            Console.ForegroundColor = ConsoleColor.Green;
            Console.Write(input.Text);

            Console.ForegroundColor = ConsoleColor.White;
            var outputs = _chatSession.ChatAsync(userInput, new LLama.Common.InferenceParams()
            {
                RepeatPenalty = 1.0f,
                AntiPrompts = new string [ ] { "User:" },
                FrequencyPenalty = 1.0f,
                Temperature=0.5f
            });
            var result = "";
            await foreach (var output in outputs)
            {
                Console.Write(output);
                result += output;
            }

            return result;
        }
````

but the issue i am encountering is that i cant seem to let this run on my gpu it only uses my cpu and ram

AsakusaRinne:
Hi, which backend package did you install for your project? There's three kinds of backends: cpu, cuda and Metal (MAC), be care that cpu and cuda backend shouldn't be installed at the same time.

jonty-esterhuizen:
so when using LLamaSharp.Backend.cuda12 and LLamaSharp both versions 0.5.1
i encountered this error 
RuntimeError: The native library cannot be found. It could be one of the following reasons: 
[...]

PR #202, opened 2023-10-20
Multi GPU

martindevans:
Added support for multi GPU in the model config.

~~This is a litte rough at the moment. The `IModelParams.TensorSplits` array length must be <= `NativeApi.llama_max_devices()`. If it's too small it's padded up to size (with zeroes). If it's too large the `ToLlamaModelParams` method throws. I'd rather not require the user to query the native API directly, and I'd rather the error were caught at config time.~~

This is exposed through a `TensorSplitsCollection` on the `IModelParams` interface. The collection can be indexed into (like an array) and automatically has the correct size. There is no way to change the size, preventing incorrect usage.

 - [ ] Tested with multiple GPUs

### DIFF
--- a/LLama.Unittest/ModelsParamsTests.cs
+++ b/LLama.Unittest/ModelsParamsTests.cs
@@ -12,37 +12,49 @@ namespace LLama.Unittest
                 BatchSize = 17,
                 ContextSize = 42,
                 Seed = 42,
-                GpuLayerCount = 111
+                GpuLayerCount = 111,
+                TensorSplits = { [0] = 3 }
             };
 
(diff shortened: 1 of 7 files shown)


martindevans:
Note that this PR breaks Newtonsoft.Json serialization of `IModelParams`. `System.Text.Json` still works fine.

Last time we discussed this I don't think anyone was particularly interested in supporting JSON serialization _at all_ and I only added it because it was very easy to do.

martindevans:
Some discussion about this PR over in #189.

martindevans:
Tested by swiftress on Discord and confirmed working (https://discord.com/channels/1106946823282761851/1106947264938790972/1166937319664795689)

Issue #203, opened 2023-10-22
Using LLamaSharp with Korean langauge


sinusinu:
If I try to use LLamaSharp with Korean language, the output gets garbled. This does not happen in llama.cpp, so I suspect this is a problem with LLamaSharp.

- LLamaSharp 0.5.1
- Model: `llama-2-7b-guanaco-qlora.Q4_K_M.gguf`
- Prompt (Translated version of chat-with-bob): `사용자가 철수라는 어시스턴트와 상호작용하는 대화 내용입니다. 철수는 도움이 되고 친절하며 정직하고 글쓰기에 능숙하며 사용자의 요청에 즉각적이고 정확하게 답변하는 데 실패한 적이 없습니다.\r\n\r\n사용자: 안녕하세요, 철수.\r\n철수: 안녕하세요. 오늘은 무엇을 도와드릴까요?\r\n사용자: 유럽에서 가장 큰 도시를 알려주세요.\r\n철수: 네. 유럽에서 가장 큰 도시는 러시아의 수도인 모스크바입니다.\r\n사용자:`
- Anti-prompt is not used, so text is all generated.

llama.cpp:
![20231022-172856-WindowsTerminal](https://github.com/SciSharp/LLamaSharp/assets/1446561/9fe04878-fcbb-4c64-9157-8bc2b380e190)

LLamaSharp:
![20231022-172909-WindowsTerminal](https://github.com/SciSharp/LLamaSharp/assets/1446561/baff687a-ec1d-4827-8ddc-3424e04c7898)

My gut feeling is that llama.cpp generates tokens in UTF-8, and Korean characters' surrogate pairs are getting sliced into separate tokens, and LLamaSharp does `Encoding.UTF8.GetString(byte[])` with sliced surrogate pairs in somewhere, converting them into U+FFFD. A korean character takes 3 bytes in UTF-8, and the lengths of all U+FFFD blocks are divisible by 3...

It's just my guess, I have no evidence, I looked at the code and couldn't understand how it works enough to tell if my theory is correct or not.

Can this be fixed? Am I doing something wrong?

martindevans:
> Am I doing something wrong?

I don't think you're doing anything wrong. There have been some other reports of bad character encoding with languages such as Chinese and Russian. I've even seen some broken characters when using English (I think the LLM was trying to return emoticons in that case, which are also multi-byte characters).

I've created PR #205 which I think reproduces the problem.

martindevans:
[...]

PR #205, opened 2023-10-22
RoundTrip Tokenization Errors

martindevans:
See #203 for original bug report.

After some investigation I discovered that the problem is how tokens are converted into text. LLamaSharp has been built with the assumption that **one token -> one or more characters** (i.e. one to many). However, this is not true! Some characters require multiple tokens to encode:
 - The character `철` maps to multiple tokens: `[29871, 239, 181, 163]`
 - Test it by pasting that token into [this tool](https://huggingface.co/spaces/Xenova/the-tokenizer-playground) or [this other tool](https://belladoreai.github.io/llama-tokenizer-js/example-demo/build/)
 - Both of those external tools show a broken decoding of that character, they have the same bug!

---

Fixing this required quite an extensive redesign of how tokens are converted into text!

Additions:
 - Introduced a new `StreamingTokenDecoder` which accepts tokens and accumulates a buffer of characters. If a single token does not decode into a valid character that's fine, it remembers the state internally.
 - Also introduced a new `AntipromptProcessor` which keeps a buffer of previously decoded text. This is safer than using the `TokensEndsWithAnyString` methods.
 - Modified `StatelessExecutor` to use these, it is totally fixed.

Removals:
 - The various `DeTokenize` and `TokenToString` methods are all incorrect to use. They have been marked as Obsolete or removed.
 - The `TokensEndsWithAnyString` methods are also incorrect to use and have been marked as Obsolete.

Work remaining for future PRs:
 - Remove all the obsolete methods.
 - Fix the other executors in the same way.

### DIFF
--- a/LLama.Unittest/BeamTests.cs
+++ b/LLama.Unittest/BeamTests.cs
@@ -32,13 +32,14 @@ public sealed class BeamTests
     {
         const int num_beams = 2;
         const int n_predict = 3;
+        const string prompt = "The cat sat on";
 
         var context = _model.CreateContext(_params);
 
         var result = new StringBuilder();
 
-        var initial_tokens = context.Tokenize("The cat sat on");
-        result.Append(context.DeTokenize(initial_tokens.ToArray()));
+        var initial_tokens = context.Tokenize(prompt);
+        result.Append(prompt);
         context.Eval(initial_tokens, 0);
 
         NativeApi.llama_beam_search(context.NativeHandle, (data, state) =>
--- a/LLama.Unittest/TokenTests.cs
+++ b/LLama.Unittest/TokenTests.cs
@@ -74,13 +74,79 @@ public sealed class TokenTests
     }
 
     [Fact]
+    public void TokensEndWith2()
+    {
+        var tokens = _model.NativeHandle.Tokenize("The cat sat on the edge of the mat", false, true, Encoding.UTF8);
+
+        var decoder = new StreamingTokenDecoder(Encoding.UTF8, _model);
+        decoder.AddRange(tokens);
+
+        var processor = new AntipromptProcessor(new[]
+        {
+            "a fish",
+            "the mat",
+            "this is an improbably long query to be using for this method"
+        });
+        var result = processor.Add(decoder.Read());
+
+        Assert.True(result);
+    }
+
+    [Fact]
+    public void TokensEndSubstring2()
+    {
+        var tokens = _model.NativeHandle.Tokenize("The cat sat on the edge of the mat", false, true, Encoding.UTF8);
+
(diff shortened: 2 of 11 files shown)


martindevans:
Investigation Notes:
 - Using this test string: `철수라는`
 - In UTF8 bytes that is: `[ 236, 178, 160, 236, 136, 152, 235, 157, 188, 235, 138, 148 ]`, (each triplet of bytes is one character)
 - Using [this tool](https://huggingface.co/spaces/Xenova/the-tokenizer-playground) that should be these tokens: `[1, 29871, 239, 181, 163, 30970, 31197, 31081]`.
   - LLamaSharp **does** generates the correct tokens.
   - Note that the tool linked above also generates broken characters!
   - So does [this other tool](https://belladoreai.github.io/llama-tokenizer-js/example-demo/build/)!
 - The character `철` maps to multiple tokens: `[29871, 239, 181, 163]`
   - The LLamaSharp tokenizer assumes that there is a **one to many** relationship between tokens and characters; i.e. one token can become multiple characters. However in this case it is **many to one**; i.e. multiple tokens are becoming one character!

martindevans:
Fix Notes:
 - We have `TokenToString` methods - those seem to be broken conceptually since we now know a single token doesn't always map to a valid string.

martindevans:
I've made a start on fixing this, but it's going to require a bit of a rework of how we handle some detokenization stuff.

martindevans:
Latest commit remove all `TokenToString` methods. This is a breaking change, but I think it's needed here. Since we now know that a single character may require several tokens it would never be correct to use the `TokenToString` methods.

martindevans:
@sinusinu this _should_ fix the issue you reported in #203, as long as you use the `StatelessExecutor`. There's still work needed to fix the other executors.

Would you mind testing it to confirm if it works for you?

sinusinu:
Well, except for the output being a little wacky, looks like StatelessModeExecute example is working fine with Korean output!
20231023-100701-WindowsTerminal

martindevans:
> except for the output being a little wacky

Is this a significantly _different_ response than it generates without this PR? This should only affect character encoding and antiprompt detection, but other than that it shouldn't change the output!

sinusinu:
Output text is being...accumulated? Full text is repeatedly printed each time instead of just new tokens. Same thing happens with English.

Without PR:
20231023-233413-WindowsTerminal

With PR:
20231023-233206-WindowsTerminal

martindevans:
Aha ok, I'll have a look into that later. I'm surprised my unit tests didn't pick that up!

martindevans:
@sinusinu It should be fixed now.

sinusinu:
Yes, now it works as expected. Very nice!
20231024-010137-WindowsTerminal

PR #223, opened 2023-10-26
New Binaries & Improved Sampling API

martindevans:
A lot of work has ended up bundled into this one PR!

My initial aim with the work in this PR was to build a batched decoding prototype. That's included in this PR (as example #15).

Working with batched decoding required updates to the binaries again, since some of the batched decoding API has changed since we added support in #185.

Updating those binaries required changes to the sampling API because some sampling methods were changed in `llama.cpp`. I moved all the sampling methods over from being static methods in `SamplingAPI` to being instance methods on `LLamaTokenDataArray`. In the process I fixed a bug (properly copying the `sorted` flag back from the C++ side), I think this would have been causing a small performance reduction with sampling.

New Binaries from: [this commit](https://github.com/ggerganov/llama.cpp/commit/207b51900e15cc7f89763a3bb1c565fe11cbb45d)


martindevans:
@AsakusaRinne just checking to make sure I understand the new release process. Since this now has the `patch-release` tag that means it will automatically push out a new `0.6.1` package when merged?

lexxsoft:
@martindevans, in regards to testing https://github.com/SciSharp/LLamaSharp/issues/225 - it seems the performance issue has been resolved, it works with the same speed as with v0.5.1.

martindevans:
Thanks for confirming that @lexxsoft

Issue #224, opened 2023-10-27
osx-arm64 platform


tonual:
First thing first thank you for the project and effort enabling llama model inference with .NET!

I use Mac M2 and was trying various combinations from the compatibility table (llamasharp-backend-model) with the basic console example.
Not sure if this is the arm64 issue alone but I'am always getting same error about missing assembly (no matter which version I use)

> Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'LLamaSharp, Version=0.5.0.0, >Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.

>File name: 'LLamaSharp, Version=0.5.0.0, Culture=neutral, PublicKeyToken=null'
>   at Program.<Main>$(String[] args)
>   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
>   at Program.<Main>$(String[] args)
>   at Program.<Main>(String[] args)

**.csproj**

```
<PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net7.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <Platforms>osx-arm64</Platforms>

  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="LLamaSharp" Version="0.5.1" />
    <PackageReference Include="LLamaSharp.Backend.Cpu" Version="0.5.1" />
  </ItemGroup>

```

SignalRT:
It works without problems on Mac M2. Can you try to execute the test or examples in the Master or LLamaSharp?

philippjbauer:
@tonual Try the following csrpoj file contents:

```csproj
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net7.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="LLamaSharp" Version="0.7.0" />
    <PackageReference Include="LLamaSharp.Backend.MacMetal" Version="0.7.0" />
  </ItemGroup>

  <Target Name="CopyNativeLibs" AfterTargets="Build" Condition="$([MSBuild]::IsOsPlatform('OSX'))">
[...]

PR #226, opened 2023-10-30
Adapts to SK Kernel Memory 

xbotter:
Adapts to SK Kernel Memory (Semantic Memory)
closes #200

PR #238, opened 2023-11-03
fix: binary not copied on MAC platform.

AsakusaRinne:


AsakusaRinne:
I'll merge this PR since it's confirmed to fix #224 . The next patch release including this fix, however, will be delayed until #226 merged, along with a new package `LLamaSharp.kernel-memory`.

PR #242, opened 2023-11-04
MacOs intel

SignalRT:
This is a try to generate intel binaries. Just waiting for confirmation that it works.

Pending:

- [ ] Verification that It works.
- [ ] Delete metal files as is generate with metal off


martindevans:
If we have two sets of MacOS binaries we're going to need some way to determine which set to load at runtime.

The runtime feature detection system can be used for that with some minor extensions.

SignalRT:
I will close this PR because the complete work is in #258

PR #244, opened 2023-11-04
build: add package for kernel-memory integration.

AsakusaRinne:


LLama.KernelMemory/LLamaSharp.KernelMemory.csproj
```diff
@@ -1,9 +1,30 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
```
martindevans, line 4: Are we intentionally dropping/not including netstandard2.0 support for kernel memory?

I'm not against that, just wanted to make sure it's deliberate!

AsakusaRinne: This a line that was removed! 🤣 What I was doing is adding the frameworks including netstandard2.0. I think we won't drop netstandard2.0 because it has many users.

AsakusaRinne: Update: Though we want to support netstandard2.0, kernel-memory does not... I removed it and will add it back in the future if kernel-memory developers change the mind.

LLama.Examples/LLama.Examples.csproj
```diff
@@ -29,7 +29,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
-    <PackageReference Include="Microsoft.SemanticKernel" Version="1.0.0-beta4" />
+    <PackageReference Include="Microsoft.SemanticKernel" Version="1.0.0-beta1" />
```
martindevans, line 32: Downgrading the SemanticKernel version?

martindevans: (from `beta4` to `beta1`)

AsakusaRinne: Yes because the LLamaSharp.semantic-kernel package depends on beta1, though using beta4 seems to have no error.

PR #248, opened 2023-11-05
feat(kernel-memory): avoid loading model twice.

AsakusaRinne:
@xbotter Please help to review it. Thank you!

related with: https://github.com/SciSharp/LLamaSharp/issues/35#issuecomment-1793540192

### DIFF
--- a/LLama.KernelMemory/BuilderExtensions.cs
+++ b/LLama.KernelMemory/BuilderExtensions.cs
@@ -4,6 +4,9 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using LLama;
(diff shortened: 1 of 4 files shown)


xbotter (approved):
LGTM. 
I noticed that there is an EmbeddingMode in ModelParams. Will it have any impact on the embedding?

martindevans:
I'm not certain but I think embedding mode loads the model in a way that can _only_ do embedding.

AsakusaRinne:
Martin's right. Maybe one of the reasons behind it is the kv cache. If only using embedding mode, it seems no need for kv-cache. However what confused me is that why a model for inference cannot be used to generate embeddings.

PR #259, opened 2023-11-06
Enhance framework compatibility

Uralstech:
**Overview:**
This pull request enhances framework compatibility of LLamaSharp, addressing some issues discussed in SciSharp/LLamaSharp#178.

**Changes Made:**
- Updated preprocessor directives in the LLamaSharp C# project to ensure compatibility with .NET Standard 2.0 and higher and .NET 6.0 and higher.
- Additionally, preprocessor errors have been introduced to provide clear feedback when attempting to target a possibly unsupported framework.

### DIFF
--- a/LLama/Common/FixedSizeQueue.cs
+++ b/LLama/Common/FixedSizeQueue.cs
@@ -48,8 +48,6 @@ namespace LLama.Common
             // in which case we'll have to check later
             if (data.TryGetNonEnumeratedCount(out var dataCount) && dataCount > size)
                 throw new ArgumentException($"The max size set for the quene is {size}, but got {dataCount} initial values.");
-#elif !NETSTANDARD2_0_OR_GREATER
-#error Target framework not supported!
 #endif
 
             // Size of "data" is unknown, copy it all into a list
--- a/LLama/Common/FixedSizeQueue.cs
+++ b/LLama/Common/FixedSizeQueue.cs
@@ -43,11 +43,13 @@ namespace LLama.Common
         /// <param name="data"></param>
         public FixedSizeQueue(int size, IEnumerable<T> data)
         {
-#if !NETSTANDARD2_0 
+#if NET6_0_OR_GREATER
             // Try to check the size without enumerating the entire IEnumerable. This may not be able to get the count,
             // in which case we'll have to check later
             if (data.TryGetNonEnumeratedCount(out var dataCount) && dataCount > size)
                 throw new ArgumentException($"The max size set for the quene is {size}, but got {dataCount} initial values.");
+#elif !NETSTANDARD2_0_OR_GREATER
+#error Target framework not supported!
 #endif
 
             // Size of "data" is unknown, copy it all into a list
--- a/LLama/Extensions/ListExtensions.cs
+++ b/LLama/Extensions/ListExtensions.cs
@@ -5,12 +5,14 @@ namespace LLama.Extensions
 {
     internal static class ListExtensions
     {
-#if NETSTANDARD2_0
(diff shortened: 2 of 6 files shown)


martindevans:
Thanks for the PR!

I don't think the approach taken here it quite right though. The extension methods covered with `#if NETSTANDARD2_0` are usually polyfills for methods which were introduced in later versions of dotnet. That means they should be defined on platforms where those methods don't exist, but must not exist on platforms where those methods already exist.

You could probably use `#if !NETSTANDARD2_1` for most of the blocks I would guess?

Uralstech:
> I don't think the approach taken here it quite right though. The extension methods covered with `#if NETSTANDARD2_0` are usually polyfills for methods which were introduced in later versions of dotnet. That means they should be defined on platforms where those methods don't exist, but must not exist on platforms where those methods already exist.
> 
> You could probably use `#if !NETSTANDARD2_1` for most of the blocks I would guess?

I understand. In my previous testing with Unity, I had found `EnsureCapacity<T>(this List<T> list, int capacity)` is not defined, so I thought it would be better to update all instances of `NETSTANDARD2_0`.

I'll just have `EnsureCapacity` with `NETSTANDARD2_0 || NETSTANDARD2_1` then!

LLama/Common/FixedSizeQueue.cs
```diff
@@ -43,11 +43,13 @@ public FixedSizeQueue(int size)
         /// <param name="data"></param>
         public FixedSizeQueue(int size, IEnumerable<T> data)
         {
-#if !NETSTANDARD2_0 
+#if NET6_0_OR_GREATER
             // Try to check the size without enumerating the entire IEnumerable. This may not be able to get the count,
             // in which case we'll have to check later
             if (data.TryGetNonEnumeratedCount(out var dataCount) && dataCount > size)
                 throw new ArgumentException($"The max size set for the quene is {size}, but got {dataCount} initial values.");
+#elif !NETSTANDARD2_0_OR_GREATER
```
martindevans, line 51: I don't think we need the failure case here. This block of code is just an (extremely small) optimisation, if it doesn't run it doesn't matter.

LLama/Extensions/ListExtensions.cs
```diff
@@ -5,12 +5,14 @@ namespace LLama.Extensions
 {
     internal static class ListExtensions
     {
-#if NETSTANDARD2_0
+#if NETSTANDARD2_0 || NETSTANDARD2_1
```
martindevans, line 8: `EnsureCapacity` was introduced with `NET6` (docs: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.ensurecapacity?view=net-7.0). So I think this can just be:

`#if !NET6_OR_GREATER`

martindevans (changes requested):

martindevans:
Thanks for fixing that so quickly, just a couple of other minor nits I spotted in review :)

LLama/Common/FixedSizeQueue.cs
Uralstech: Ok, I've removed the case!

LLama/Extensions/ListExtensions.cs
Uralstech: I've updated it to `#if !NET6_0_OR_GREATER`!

martindevans (approved):

martindevans:
Looks good to me! Just waiting for those tests to run and I'll merge it. Thanks very much :)

Uralstech:
Thank you! You're welcome!

PR #268, opened 2023-11-08
feat: add detection template for cuda and avx.

AsakusaRinne:
Just for review, please don't merge it.

### DIFF
--- a/LLama/Native/NativeApi.cs
+++ b/LLama/Native/NativeApi.cs
@@ -1,8 +1,13 @@
 ﻿using System;
 using System.Buffers;
+using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text;
 using LLama.Exceptions;
+using ManagedCuda;
+#if NET6_0_OR_GREATER
+using System.Runtime.Intrinsics.X86;
+#endif
 
 #pragma warning disable IDE1006 // Naming Styles
 
@@ -24,8 +29,9 @@ namespace LLama.Native
     {
         static NativeApi()
         {
-            // Try to load a preferred library, based on CPU feature detection
-            TryLoadLibrary();
+#if NET6_0_OR_GREATER
+            NativeLibrary.SetDllImportResolver(typeof(NativeApi).Assembly, LLamaImportResolver);
+#endif
 
             try
             {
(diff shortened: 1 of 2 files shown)


LLama/Native/NativeApi.cs
```diff
@@ -44,61 +50,120 @@ static NativeApi()
...
+            AvxLevel level = AvxLevel.None;
+#if NET6_0_OR_GREATER
+            if (Avx.IsSupported) level = AvxLevel.Avx;
+            if (Avx2.IsSupported) level = AvxLevel.Avx2;
 #if NET8_0_OR_GREATER
-                    ?? TryLoad("avx512/libllama.so", System.Runtime.Intrinsics.X86.Avx512.IsSupported)
+             if(Avx512F.IsSupported) level = AvxLevel.Avx512;
 #endif
-                    ?? TryLoad("avx2/libllama.so", System.Runtime.Intrinsics.X86.Avx2.IsSupported)
-                    ?? TryLoad("avx/libllama.so", System.Runtime.Intrinsics.X86.Avx.IsSupported)
```
martindevans, line 76: AVX/AVX2/AVX512 detection should already be handled by this

LLama/Native/NativeApi.cs
```diff
@@ -24,8 +29,9 @@ public unsafe partial class NativeApi
     {
         static NativeApi()
         {
-            // Try to load a preferred library, based on CPU feature detection
-            TryLoadLibrary();
+#if NET6_0_OR_GREATER
+            NativeLibrary.SetDllImportResolver(typeof(NativeApi).Assembly, LLamaImportResolver);
```
martindevans, line 33: I deliberately avoided using `NativeLibrary.SetDllImportResolver` when I originally set this up, since it overwrites global configuration that may already be in use by the application.

martindevans:
The intention when I originally introduced this was that new dependencies could easily be added (in order of preference, most preferred first) something like this:

```csharp
return TryLoad("magic/libllama.dll", IsMagicSupported())
       ?? TryLoad("unicorns/libllama.dll", IsUnicornsSupported())
       ?? TryLoad("rainbows/libllama.dll", IsRainbowsSupported())
       ?? etc...
```

i.e. any new platform can simply be chained in, so long as you can write a boolean method that prechecks if that particular thing is supported.

I deliberately avoided using `NativeLibrary.SetDllImportResolver` since in theory it could have already been set externally on this assembly, and that would cause an exception the second time. I guess it's not a huge risk really, it just feels like a poorly designed API really (feels like it should be chainable).

PR #274, opened 2023-11-11
Update ToLLamaSharpChatHistory extension method to be public and support semantic-kernel author roles

kidkych:
Currently when semantic-kernel ChatHistory is converted to LLamaSharp ChatHistory through the ToLLamaSharpChatHistory extension method in https://github.com/SciSharp/LLamaSharp/blob/master/LLama.SemanticKernel/ExtensionMethods.cs the AuthorRole in the resulting LLamaSharp ChatHistory is set to AuthorRoles.Unknown as semantic-kernel uses lowercase author names while LLamaSharp capitalizes the first character. This PR tells the Enum.TryParse method to ignore letter case when matching.

Further, it exposes the ToLLamaSharpChatHistory extension method publicly since it would be useful for debugging in downstream applications.

### DIFF
--- a/LLama.SemanticKernel/ExtensionMethods.cs
+++ b/LLama.SemanticKernel/ExtensionMethods.cs
@@ -5,7 +5,7 @@ namespace LLamaSharp.SemanticKernel;
 
 public static class ExtensionMethods
 {
-    public static global::LLama.Common.ChatHistory ToLLamaSharpChatHistory(this ChatHistory chatHistory)
+    public static global::LLama.Common.ChatHistory ToLLamaSharpChatHistory(this ChatHistory chatHistory, bool ignoreCase = true)
     {
         if (chatHistory is null)
         {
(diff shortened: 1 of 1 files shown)


LLama.SemanticKernel/ExtensionMethods.cs
```diff
@@ -3,9 +3,9 @@
 
 namespace LLamaSharp.SemanticKernel;
 
-internal static class ExtensionMethods
+public static class ExtensionMethods
 {
-    internal static global::LLama.Common.ChatHistory ToLLamaSharpChatHistory(this ChatHistory chatHistory)
+    public static global::LLama.Common.ChatHistory ToLLamaSharpChatHistory(this ChatHistory chatHistory)
```
AsakusaRinne, line 8: I think adding a parameter `bool ignoreCase = true` to allow control on it will be better.

AsakusaRinne (changes requested):
LGTM except one detail, thank you for this contribution!

LLama.SemanticKernel/ExtensionMethods.cs
kidkych: Makes sense to me, just added the parameter and thank you for the awesome library! This is actually my first contribution to an open source project :smile:

AsakusaRinne: You're always welcome! I'm happy that our work makes sense to you. If you're interested in developing this library, there's a TODO list in [LLamaSharp project](https://github.com/orgs/SciSharp/projects/5). 😊

AsakusaRinne (approved):

LLama.SemanticKernel/ExtensionMethods.cs
kidkych: Sure, I'd love to help! It looks like the project board is set to private though, I'm getting a 404 error

AsakusaRinne: Sorry for that, I've now changed it to public. :)

Issue #270, opened 2023-11-09
Linux cuda crash


ChengYen-Tang:
I'm getting a crash using LLamaSharp.Backend.Cuda11
This problem occurs in 0.6.0 and 0.7.0, and is normal in 0.5.1
![image](https://github.com/SciSharp/LLamaSharp/assets/32535494/aa01f453-2bc8-4ecf-91ae-cb2eba55117c)

Env:
Ubuntu 22.04
![image](https://github.com/SciSharp/LLamaSharp/assets/32535494/d7af4df3-da69-49fe-8eab-49854b287bff)

martindevans:
We have a new set of binaries ready to release in the next version (very soon). Could you try pulling the master branch and testing if it fixes this issue?

ChengYen-Tang:
Same error

martindevans:
What CPU and GPU are you using?

ChengYen-Tang:
CPU: I7-8700
GPU: Titan RTX 24G

AsakusaRinne:
It's quite weird because the only difference on compilation of native library since v0.6.0 is avx. 

We've published v0.8.0 just now, with a fix for cuda library (but not sure if it's helpful for this issue) and the cuda feature detection in #275. Could you please have a try with it?

If it still doesn't work, you could add `NativeLibraryConfig.WithLibrary({Your_DLL_PATH})` at the very beginning of your code, to load a cuda library compiled yourself. That could be a temporary fix.

ChengYen-Tang:
Oh, it’s normal
Another question, how should I stop printing logs?

AsakusaRinne:
> Oh, it’s normal

Is the v0.8.0 normal or your self-compiled library normal?

> Another question, how should I stop printing logs?

Stopping logs from llama.cpp is an another issue. Currently you could try to redirect the outputs to other where.

AsakusaRinne:
Update: [this comment](https://github.com/SciSharp/LLamaSharp/issues/33#issuecomment-1714736219) will allow you to capture the outputs from native library and output them anywhere else.

ChengYen-Tang:
> > Oh, it’s normal
> 
> Is the v0.8.0 normal or your self-compiled library normal?
> 
> > Another question, how should I stop printing logs?
> 
> Stopping logs from llama.cpp is an another issue. Currently you could try to redirect the outputs to other where.

v0.8.0

[...]

PR #275, opened 2023-11-11
feat: cuda feature detection.

AsakusaRinne:
This PR is supposed to be merged after #258 

Note that it adds two APIs for users, which are `NativeLibraryConfig.WithLibrary` and `NativeLibraryConfig.WithMatchRule`.  To be honest I think the naming is not good but I can't come up with a better one. Any idea?

TODO: 

- [x] Check on Windows
- [x] Check on Linux
- [x] Check on MacOS
- [x] Add necessary logs
- [x] Change file structure in nuget package and verify it


AsakusaRinne:
Is there a good way for us to print some logs in `NativeApis` now? I think directly outputting to console is not a good idea.

martindevans:
I _think_ you can do something like `Logger.GetLogger()` now?

LLama/NativeLibraryConfig.cs
```diff
@@ -0,0 +1,113 @@
...
+    /// </summary>
+    public class NativeLibraryConfig
+    {
+        private static NativeLibraryConfig? instance;
+        private static readonly object lockObject = new object();
+
+        /// <summary>
+        /// Whether there's already a config for native library.
+        /// </summary>
+        public bool Initialied { get; private set; }
```
martindevans, line 18: Should be `Initialized`

LLama/NativeLibraryConfig.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            var config = GetInstance();
+            if (config.Initialied)
+            {
+                throw new InvalidOperationException("NativeLibraryConfig could be configured only once before any call to llama model apis.");
+            }
+            config.Desc = new Description(libraryPath);
+        }
+
+        /// <summary>
```
martindevans, line 51: I think it'd be better to split this into various `WithX` methods rather than having them all in one method (more extensible in the future).

e.g.

```csharp
WithCuda(bool cuda)
WithAvx(AvxLevel avx)
WithAutoFallback(bool allow)
WithSkipCheck(bool skip)
```

LLama/NativeLibraryConfig.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// Ass rules to match a suitable library from installed LLamaSharp backend.
+        /// </summary>
+        /// <param name="useCuda"></param>
+        /// <param name="avxLevel"></param>
+        /// <param name="allowFallback">Whether to allow fall-back when your hardware doesn't support your configuration.</param>
+        /// <param name="skipCheck">Whether to skip the check when fallback is allowed.
```
martindevans, line 57: I'm not really sure what `skipCheck` does from this description. it's "skipping the check", but what check?

martindevans:
Just a few review nits, but overall looks good.

My only criticism is it's less clear than before which binaries are preferred over others. Previously you could just read down the list of `TryLoad` calls and it was very obvious. Not a fatal flaw, just an observation!

LLama/NativeLibraryConfig.cs
AsakusaRinne: It's to check if the hardware support the user's choice. Let's ignore the `skipCheck` parameter first, then there're two basic conditions, allowing fallback and the opposite.
If fallback is allowed, we could choose another library if the user's configuration does not match the hardware. For example, the user specified cuda but there's actually no supported cuda version at that machine.

If fallback is disabled, however, we're supposed to load exactly the library specified. However the real condition may be complex and our logic may not cover all the cases (for example, linux cuda detection). Therefore if sometimes loading a library won't actually have problem but we take it as invalid by mistake, the user could force-load it without any check. 

I think by this way we could leave a solution for some potential problems in the future.

LLama/NativeLibraryConfig.cs
AsakusaRinne: This format seems good, especially for the ability of chained call. There's one problem that it couldn't notice user the default configuration well. If we just set default value to the parameter, it may looks strange when calling `WithAutoFallback().WithSkipCheck(false)`. 😹

Anyway, I prefer this format, too. However is there any solution about the case above?

AsakusaRinne:
The linux check has passed but the binaries seemed not to work. I can only load the library when I replaced the binaries with self-compiled ones. Not sure if it's related with #270

AsakusaRinne:
@SignalRT Could you please help to test it on MAC? For nuget package test please download [this file](https://public-1306037490.cos.ap-beijing.myqcloud.com/LLamaSharp_v0.8.0.zip).

SignalRT:
@AsakusaRinne I get this exception trying to load LLamaSharp:

System.BadImageFormatException: Could not load file or assembly 'LLamaSharp, Version=0.8.0.0, Culture=neutral, PublicKeyToken=null'. An attempt was ma...

System.BadImageFormatException
Could not load file or assembly 'LLamaSharp, Version=0.8.0.0, Culture=neutral, PublicKeyToken=null'. An attempt was made to load a program with an incorrect format.

   at LLama.Unittest.BasicTest..ctor()
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean publicOnly, Boolean wrapExceptions)

AsakusaRinne:
@SignalRT It seems to be a problem of LLamaSharp package instead of backend packages. Is your system 32-bit or 64-bit?

SignalRT:
@AsakusaRinne, Arm64 (M2 computer). I can run the test in the project, execute the examples, but with your packages or with the packages generated manually in my computer with prepare_release.sh it fails with the same error.

I'm trying to understand the problem....that seems a little weird.

SignalRT:
@AsakusaRinne After a lot of problems, it seems that it's related to some problem with nuget cache. I clear the cache and it works with the packages generated in the CI pipeline

AsakusaRinne:
> @AsakusaRinne After a lot of problems, it seems that it's related to some problem with nuget cache. I clear the cache and it works with the packages generated in the CI pipeline

Wow, that's nice! I think we could release it tonight.

@martindevans The CUDA libraries have been replaced the ones with avx 2, is that right?

martindevans:
> The CUDA libraries have been replaced the ones with avx 2, is that right?

Yep, the CUDA binaries should now compile with AVX2 by default.

PR #292, opened 2023-11-15
dotnet8.0

martindevans:
Added dotnet8.0 as a build target


martindevans:
@AsakusaRinne do you think we should disable the CodeQL for now? It doesn't seem to have ever worked properly and now it's breaking with net8.0 for some reason.

AsakusaRinne:
> @AsakusaRinne do you think we should disable the CodeQL for now? It doesn't seem to have ever worked properly and now it's breaking with net8.0 for some reason.

Makes sense to me, I think removing `codeql.yml` is ok

PR #297, opened 2023-11-15
Ubuntu 20.04 Compile

martindevans:
Moved compile job to run on `ubuntu-20.04`, hopefully picking up an older version of glibc with broader support.


AsakusaRinne:
Just a question, when will the `compile.yml` be triggered? It seems that the ci workflow doesn't include it.

Issue #308, opened 2023-11-17
'tensor splits length must equal 16'


JohnGalt1717:
I'm setting TensorSplits  = new LLama.Abstractions.TensorSplitsCollection([12F, 4F]);

Yet I get the above error despite them equaling 16.

I'm trying to distribute memory and load across 2 Cuda GPUs which is supposed to be supported by Llama.cpp now so that I can load a model with a large context length into GPU.

martindevans:
What is the _exact_ text of the error you're getting?

JohnGalt1717:
Sorry, fixed the title which was supposed to be the exact error until auto-correct got involved.

martindevans:
Aha, yes I thought that might be the problem. The error is telling you that the **length** of the array must be 16 (because that's the maximum number of devices supported by llama.cpp).

So you'd want to do something like this:

```csharp
TensorSplits = new LLama.Abstractions.TensorSplitsCollection([12F, 4F, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ]);
```

I've just opened a PR (#310) which would make your initial code work :)

JohnGalt1717:
Awesome! Thanks!

PR #310, opened 2023-11-17
Better TensorSplitsCollection Initialisation

martindevans:
The `TensorSplitsCollection` constructor allows passing an array of splits. At the moment this array must be exactly `NativeApi.llama_max_devices()` items long.

Modified it so that the supplied value can be less than that long, the rest will be set to zero.


martindevans:
Resolve #308

Issue #313, opened 2023-11-18
AIRequestSettings object can not be brute forced into a ChatRequestSettings object


futzy314:
Hi, I'm trying to get a sample from SemanticKernel working with LlamaSharp, and its failing when the SemanticKernel plugin function specifies any settings.  I took a look at the code, and it's trying to brute force convert the AIRequestSettings object into a ChatRequestSettings object, which won't work.  There needs to be an actual conversion function instead.  I can take a stab at it if you like, I'd like to help out with the SemanticKernel integration.

AsakusaRinne:
Hi, we'll appreciate it if you would like to help! Please feel free to ask me and @xbotter for help if there's any problem. Also, if you'd like to, you can join the discord channel to chat with us. :)

xbotter:
Thank you for pointing out this issue, feel free to submit a PR to fix it.

PR #315, opened 2023-11-19
Added a converter similar to the Open AI one

futzy314:
This is all based on the same classes in their OpenAI implementation.  I tested it with a couple of Microsoft's sample projects in Semantic Kernel, and it seems to work.  I didn't see any unit tests for the Semantic Kernel code yet, but if you want to start adding that, I can add some unit tests too.

Fixes SciSharp/LlamaSharp#313

### DIFF
--- a/LLama.SemanticKernel/ChatCompletion/ChatRequestSettings.cs
+++ b/LLama.SemanticKernel/ChatCompletion/ChatRequestSettings.cs
@@ -1,4 +1,6 @@
 ﻿using Microsoft.SemanticKernel.AI;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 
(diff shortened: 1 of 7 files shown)


AsakusaRinne:
Thank you for the contribution! If you would like to, we'll appreciate it if you could add some unit test about semantic-kernel. Note that this is not a required one, the overall looks very good to me. :)

xbotter (approved):
LGTM.

futzy314:
OK, I can work on some unit tests too, will add here.  One other suggestion, I think the ChatRequestSettings class is misnamed, because it doesn't just apply to Chat, it applies to Text as well.  I suggest renaming it to LlamaRequestSettings.

yukimakura:
I was struggling with this very issue just now. Thanks for the great pull request!

PR #316, opened 2023-11-19
November Binary Update

martindevans:
Added new binaries.
 - Built by this run: https://github.com/SciSharp/LLamaSharp/actions/runs/6921572568
 - llama.cpp commit: `e937066420b79a757bf80e9836eb12b88420a218`
 - Rearranged paths of binaries
 - Added all the various AVX binaries (successfully autodetects and loads AVX2 on my PC)
 - Added new symbols from llama.h (although no high level bindings)


LLama/Native/NativeApi.Load.cs
```diff
@@ -155,13 +155,13 @@ private static List<string> GetLibraryTryOrder(NativeLibraryConfig.Description c
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 platform = OSPlatform.Windows;
-                prefix = "runtimes/win-x64/native/";
```
AsakusaRinne, line 158: I don't think we should change this path. We used to put files under the `runtimes/{system}/native` folder because they could be detected without specifying the dll path. Though the feature detection actually specify the library path manually, I prefer to just keep them unless any bad impact is confirmed. 🤣

AsakusaRinne (commented):
Could you please also update the nuspec files to ensure the auto-release working? The only things need to be done is to append the paths of new libraries. :)

LLama/Native/NativeApi.Load.cs
martindevans: Done 👍

martindevans:
@AsakusaRinne I've updated the nuspec files and I think they build correct. I haven't really worked with nuspec much before though so please look at those closely in your review!

AsakusaRinne:
It looks good, I'll double-check the nuget package generation before next release. :)

AsakusaRinne (approved):

SignalRT:
@martindevans It looks good, and I pass all test locally with the binaries.

SignalRT (approved):
It's ok for me

Issue #264, opened 2023-11-07
Feature Request: Switch backends dynamically at runtime?


BrainSlugs83:
Right now, to switch backends, I have to uninstall one nuget package, and then reinstall another. -- This means that if I make an app, I would have to create different builds for every configuration.

Instead I'd rather just give someone a dropdown box in a UI and let them pick the backend, and have the app switch on the fly to whatever they have selected.

Ideally this would just be an enum in the ModelParams, something like `Backend = LLamaSharp.Cuda`, or `Backend = LLamaSharp.Cpu`, etc.  (Heck, even if it has to be a global static parameter, that's better than having to make separate builds.)

One possible way it could be implemented is by shipping all of the libllama.dll files in the output directory with a slightly different name (i.e. libllama-cuda.dll, libllama-avx.dll, etc.) and dynamically loading the DLL into memory based on which backend is being requested.

martindevans:
It sounds like there's two parts to this request:

 - Ability to configure the backend before it is loaded.
    - At the moment this isn't possible, because the DLL loading is in the static constructor for `NativeApi`, so the instant you access that class the DLL is loaded. We could introduce another class (e.g. `NativeApiConfig`) with static config methods/properties on it.
 - Ability to _unload_ the backend and load another.
    - This could probably be done by using `NativeLibrary.Load` and `NativeLibrary.Free` ([docs](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary?view=net-7.0)).

BrainSlugs83:
It looks like you're already trying to load them one at a time in the static API... what would be the harm in loading all of the ones that are found? and just putting the IntPtrs into an array.

We could set the one we want to use via a static enum property -- and the IntPtr could just be a read only property that returns the appropriate IntPtr based on which enum is selected.  (Or throws an exception if a missing one was selected.)

We could have another static method that is just like `IEnumerable<Backend> AvailableBackends { get; }` that does a `yield return Backend.xyz` if that backend is available...

martindevans:
Loading all of the backends at once wouldn't work with the current system because the native methods are written like this:

```csharp
[DllImport("libllama")]
public static extern void demo_method();
```

That will use the already loaded `libllama.dll` (if there is one) otherwise it will find and load it when called. Given that I'm not sure exactly what would happen it you loaded multiple, but I'm sure it wouldn't be good!

[...]

PR #333, opened 2023-11-28
feat: allow customized search path for native library loading.

AsakusaRinne:
1. Allow customized search path for native library loading. Related: #264 
2. Give a emergency bandaging to the chat session, to avoid anti-prompt missing because of the `\r\n` appended by `HistoryToText`. Related: #299

### DIFF
--- a/LLama/Native/NativeLibraryConfig.cs
+++ b/LLama/Native/NativeLibraryConfig.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace LLama.Native
 {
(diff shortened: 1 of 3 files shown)


LLama/Native/NativeLibraryConfig.cs
```diff
@@ -120,13 +129,62 @@ public NativeLibraryConfig WithLogs(bool enable = true)
...
         }
 
+        /// <summary>
+        /// Add self-defined search directories. Note that the file stucture of the added 
+        /// directories must be the same as the default directory. Besides, the directory 
+        /// won't be used recursively.
+        /// </summary>
+        /// <param name="directoriesAndPriorities">The directories and corresponding priorities, in which 0 is the lowest. The default path has priority 0.</param>
+        /// <returns></returns>
+        public NativeLibraryConfig WithSearchDirectories(IDictionary<string, int> directoriesAndPriorities)
```
martindevans, line 139: Would it makes more sense to use an array for the directories (here and for `_searchDirectories`) which is implicitly in priority order?

martindevans (approved):

PR #341, opened 2023-12-01
🔧 Refactor Semantic Kernel chat completion implementation

xbotter:
Breaking Changes
---
- Refactored the chat completion implementation in `LLamaSharpChatCompletion.cs` to use `StatelessExecutor` instead of `InteractiveExecutor`.

Changes
---
- Updated the chat history prompt in `LLamaSharpChatCompletion.cs` to include a conversation between the assistant and the user.
- Modified the `HistoryTransform` class in `HistoryTransform.cs` to append the assistant role to the chat history prompt.
- Updated the constructor of `LLamaSharpChatCompletion` to accept optional parameters for `historyTransform` and `outputTransform`.
- Modified the `GetChatCompletionsAsync` and `GetChatCompletions` methods in `LLamaSharpChatCompletion.cs` to use the new `StatelessExecutor` and `outputTransform`.
- Updated the `ExtensionMethods.cs` file to include the assistant and system roles in the list of anti-prompts.

Discussion
---
The current implementation refers to the handling method of History in LLamaSharp ChatSession, but the actual execution effect still needs to be combined with models and prompts.

### DIFF
--- a/LLama.Examples/Examples/SemanticKernelChat.cs
+++ b/LLama.Examples/Examples/SemanticKernelChat.cs
@@ -16,12 +16,11 @@ namespace LLama.Examples.Examples
             // Load weights into memory
             var parameters = new ModelParams(modelPath);
             using var model = LLamaWeights.LoadFromFile(parameters);
-            using var context = model.CreateContext(parameters);
-            var ex = new InteractiveExecutor(context);
+            var ex = new StatelessExecutor(model, parameters);
 
             var chatGPT = new LLamaSharpChatCompletion(ex);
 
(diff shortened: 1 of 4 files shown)


AsakusaRinne:
Just for curiosity, will replacing InteractiveExecutor with StatelessExecutor improve the performance of SK chat?

xbotter:
This will not bring any performance changes. It is just that using StatelessExecutor is more in line with the expectations of the current SK Chat interface.
The current SK ChatCompletion interface is more of a reference to OpenAI's API, providing the ability for stateless chat completion. Currently, there is indeed no suitable way to integrate stateful models.

PR #344, opened 2023-12-05
Rebuilt ChatSession class

philippjbauer:
- Saves with serialized ChatHistory of session
- Only allows use of ChatHistory.Message (instead of raw text) for easy post-processing with IHistoryTransform implementation
- Provides History Management methods
- Allows user to regenerate last assistant message


LLama/LLamaSharp.csproj
```diff
@@ -28,6 +28,7 @@
     <Platforms>AnyCPU;x64;Arm64</Platforms>
     <PackageId>LLamaSharp</PackageId>
     <Configurations>Debug;Release;GPU</Configurations>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
```
martindevans, line 31: What does this do?

martindevans:
Thanks for re-submitting this based on master! Overall looks good, just that conflict to resolve and one comment/question.

LLama/LLamaSharp.csproj
philippjbauer: I ran into issues during build, where AssemblyInfo attributes were flagged as duplicated. See: https://stackoverflow.com/questions/10311347/duplicate-assemblyversion-attribute

philippjbauer: Screenshot 2023-12-07 at 8 47 17 AM

philippjbauer: I tried to remove the attributes in the csproj files other than the `LLamaSharp.csproj` but that didn't have any effect. Then I tried the attribute to stop generating them all-together and it worked again for me.

martindevans:
> I don't have write access to the file to resolve the conflict.

I've resolved that for you, please check that I did it right!

philippjbauer:
> > I don't have write access to the file to resolve the conflict.
> 
> I've resolved that for you, please check that I did it right!

LGTM 👍

martindevans:
The tests seems to be failing due to a type checking error, looks like something isn't right with that merge?

AsakusaRinne:
This format of prompt looks good. Is it a commonly used format in LLM now?

philippjbauer:
> The tests seems to be failing due to a type checking error, looks like something isn't right with that merge?

Hmm, I didn't run into these issues before. I thought I had updated everything that depended on the `ChatSession` class before pushing this PR. I'll have a look tomorrow and fix them up.

philippjbauer:
> This format of prompt looks good. Is it a commonly used format in LLM now?

What exactly are you referring to? I haven't touched the default `IHistoryTransform` implementation so I'm not sure where you're referring to the prompt format.

AsakusaRinne:
> What exactly are you referring to? I haven't touched the default `IHistoryTransform` implementation so I'm not sure where you're referring to the prompt format.

I mean the following format in the example prompt:

```
{
    "messages": [
        {
            "author_role": "System",
            "content": "Transcript of a dialog, where the User interacts with an Assistant named Bob. Bob is helpful, kind, honest, good at writing, and never fails to answer the User's requests immediately and with precision."
        },
        {
            "author_role": "User",
            "content": "Hello, Bob."
        },
        {
            "author_role": "Assistant",
            "content": "Hello. How may I help you today?"
        },
        {
            "author_role": "User",
            "content": "Please tell me the largest city in Europe."
        },
        {
            "author_role": "Assistant",
            "content": "Sure. The largest city in Europe is Istanbul, Turkey."
        }
    ]
}
```

philippjbauer:
> > What exactly are you referring to? I haven't touched the default `IHistoryTransform` implementation so I'm not sure where you're referring to the prompt format.
> 
> I mean the following format in the example prompt:
> 
> ```
> {
>     "messages": [
>         {
>             "author_role": "System",
>             "content": "Transcript of a dialog, where the User interacts with an Assistant named Bob. Bob is helpful, kind, honest, good at writing, and never fails to answer the User's requests immediately and with precision."
>         },
>        { ... }
>     ]
> }
> ```

This is simply a json representation of the chat history. This will be ultimately transformed by the `IHistoryTransform.HistoryToText()` method into the prompt format that the user specifies for the model they're using. Saving it as json makes the system "prompt format agnostic". 

As such the user (developer) is able to introduce multiple implementations of the `IHistoryTransform` interface for multiple models and easily switch between them. I ran into the problem where I had to use regular expressions to extract the content for each turn from an already formatted chat history (like `chat-with-bob.txt`). 

This also allows us to package specific `IHistoryTransform` implementations for popular prompt formats, like ChatML, Alpaca, etc. The `DefaultHistoryTransform` is kinda-sorta working for most, but not everyone is aware that the correct template does make difference for most models.

philippjbauer:
> The tests seems to be failing due to a type checking error, looks like something isn't right with that merge?

I synced my master with the latest to get the files that caused issues and updated them to use the new ChatSession integration. The sample for the Chinese character encoding works, but I have a hard time evaluating whether its output is actually correct (I sadly don't know Mandarin). This was a best guess effort on my part.

AsakusaRinne:
@philippjbauer I've added some changes and the Chinese example is generating correct answer now. It's absolutely a good work!

philippjbauer:
> @philippjbauer I've added some changes and the Chinese example is generating correct answer now. It's absolutely a good work!

Thank you!

martindevans:
Thanks for fixing the merge conflict and the Chinese example. This all looks good to me, ready for merging?

martindevans:
Thanks for all your hard work on this PR!

PR #348, opened 2023-12-08
Custom Sampling Pipelines

martindevans:
Introduced an entirely new way to sample models.

The current mechanism is hardcoded (see [here](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/LLamaContext.cs#L235)) and can only be configured by tweaking values in InferenceParams.

This means a couple of things are not possible:
 - Re-ordering sampling (for example this [PR suggests](https://github.com/ggerganov/llama.cpp/pull/4267) applying temperature twice)
 - Developing entirely new sampling mechanisms (e.g. [MinP](https://github.com/ggerganov/llama.cpp/pull/3841) could not have been developed using LLamaSharp).

This PR introduces:
 - Low level interface for entirely customised sampling `ISamplingPipeline`. If the `SamplingPipeline` property in the inference params is non-null this pipeline will be used.
 - `BaseSamplingPipeline` is an abstract implementation if `ISamplingPipeline` which makes it a little easier to implement.
 - `DefualtSamplingPipeline` which is a demo implementation that does standard sampling.

You can see an example of it in use here: https://github.com/SciSharp/LLamaSharp/pull/348/files#diff-16c496b3d63b9606d57c5d2a5592059223a9e4bc7ab29416ea775d5356888df7R37


AsakusaRinne (approved):
Great works! Greedy sampling is also a common sampling method, which is good to be provided in LLamaSharp. It's not required in this PR, just a mark. :)

martindevans:
That reminds me of something related that I thought about and I might come back to in a future PR. At the moment the LLamaContext basically has several different modes, depending on what parameters you pass:

 - Pure Greedy (Temp <= 0)
 - MiroStat
 - MiroStat2
 - TopK + TailFree + LocallyTypical + TopP + MinP + Temperature + Sampling

So using this new pipeline system we could explicitly split it into 4 separate pipelines, construct them inside LLamaContext and use them as necessary. Would split some of the code out of LLamaContext into a more re-usable form.

PR #356, opened 2023-12-11
bump sk to 1.0.1 & km to 0.18

xbotter:


martindevans:
Looks like it's failing due to the serialization of `Encoding`. That _should_ be serialized as the name of the encoding, but looks like it's not here for some reason.

xbotter:
This looks like an error introduced by upgrading to System.Text.Json 8.0.0.
The sk-rc3 version uses System.Text.Json 8.0.0, and it is currently not possible to downgrade. I will go and see how to resolve this issue.

martindevans:
I don't really know much about semantic kernel, but from what I can see everything looks good 👍

martindevans (approved):

martindevans:
Shall we go ahead and merge this?

xbotter:
We can wait a little longer because SK may release version 1.0.0 soon, and I don't know if there will be any breaking changes.

xbotter:
@martindevans SK has directly jumped to version 1.0.1 and has been updated.

LLama.KernelMemory/LLamaSharpTextEmbeddingGenerator.cs
```diff
@@ -89,5 +92,15 @@ public Task<IList<ReadOnlyMemory<float>>> GenerateEmbeddingsAsync(IList<string>
...
+
+        /// <inheritdoc/>
+        public Task<Embedding> GenerateEmbeddingAsync(string text, CancellationToken cancellationToken = default)
+        {
+            var embeddings = _embedder.GetEmbeddings(text);
+            return Task.FromResult(new Embedding(embeddings));
+        }
+
+        /// <inheritdoc/>
+        public int CountTokens(string text) => _embedder.Tokenize(text).Length;
```
martindevans, line 104:
```suggestion
        public int CountTokens(string text) => _weights.NativeHandle.Tokenize(text).Length;
```

Doing this means the `Tokenize` method on embedder can be removed.

LLama/LLamaEmbedder.cs
```diff
@@ -117,5 +117,13 @@ public void Dispose()
         {
             _ctx.Dispose();
         }
+
+        /// <summary>
+        /// Tokenize a string.
+        /// </summary>
+        public int[] Tokenize(string text, bool addBos = true, bool special = false)
```
martindevans, line 124: This can be removed if the suggestion above is merged.

martindevans:
Just one small thing I noticed, sorry should have caught that in earlier reviews.

martindevans (changes requested):
See suggested changes in other comments. Just two small changes.

LLama.KernelMemory/LLamaSharpTextEmbeddingGenerator.cs
xbotter: There is a small issue, if only the embedder is present when creating the Generator, this _weights cannot be used.
https://github.com/SciSharp/LLamaSharp/blob/40ac944fb5b9abc3a3f59ea30d8c724ace8270e3/LLama.KernelMemory/LLamaSharpTextEmbeddingGenerator.cs#L62

martindevans: Ah I see. In that case how about exposing a `LLamaWeights` property on `LLamaEmbedder`? Then you can use that to do the tokenizing. If you do that you'll probably need to delete the two obsolete constructors, they've been marked as obsolete for a while so that's not a problem.

xbotter: > In that case how about exposing a LLamaWeights property on LLamaEmbedder?

It seems like a good way.

xbotter: Due to the absence of the Weights attribute in LLamaEmbedder, I have exposed the Context attribute. The deprecated constructors have also been removed.

martindevans (approved):

martindevans:
Thanks for fixing that. Looks good to merge to me 👍

PR #358, opened 2023-12-12
Upgrade unittest target framework to .NET 8.0

xbotter:
This pull request upgrades the target framework of the unittest project to .NET 8.0.
 It also updates the serialization and deserialization code to use JsonSerializer with custom options for ModelsParamsTests .

It fixed the issue on https://github.com/SciSharp/LLamaSharp/pull/356#issuecomment-1851185829

Ref: https://github.com/dotnet/runtime/issues/95893?notification_referrer_id=NT_kwDOADd2vbI4NzMxNTMzMTE0OjM2MzQ4Nzc#issuecomment-1851644867

### DIFF
--- a/LLama.Unittest/ModelsParamsTests.cs
+++ b/LLama.Unittest/ModelsParamsTests.cs
@@ -1,4 +1,5 @@
 ﻿using LLama.Common;
+using System.Text.Json;
 
 namespace LLama.Unittest
 {
@@ -7,6 +8,8 @@ namespace LLama.Unittest
         [Fact]
         public void SerializeRoundTripSystemTextJson()
         {
+            var options = new JsonSerializerOptions();
+            options.Converters.Add(new EncodingConverter());
             var expected = new ModelParams("abc/123")
             {
                 BatchSize = 17,
@@ -16,8 +19,8 @@ namespace LLama.Unittest
                 TensorSplits = { [0] = 3 }
             };
 
-            var json = System.Text.Json.JsonSerializer.Serialize(expected);
-            var actual = System.Text.Json.JsonSerializer.Deserialize<ModelParams>(json)!;
+            var json = JsonSerializer.Serialize(expected, options);
+            var actual = JsonSerializer.Deserialize<ModelParams>(json, options)!;
 
             // Cannot compare splits with default equality, check they are sequence equal and then set to null
             Assert.Equal((IEnumerable<float>)expected.TensorSplits, expected.TensorSplits);
--- a/LLama.Unittest/ModelsParamsTests.cs
+++ b/LLama.Unittest/ModelsParamsTests.cs
@@ -8,8 +8,6 @@ namespace LLama.Unittest
         [Fact]
         public void SerializeRoundTripSystemTextJson()
         {
-            var options = new JsonSerializerOptions();
-            options.Converters.Add(new EncodingConverter());
             var expected = new ModelParams("abc/123")
             {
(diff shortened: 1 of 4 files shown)


LLama.Unittest/ModelsParamsTests.cs
```diff
@@ -7,6 +8,8 @@ public class ModelsParamsTests
         [Fact]
         public void SerializeRoundTripSystemTextJson()
         {
+            var options = new JsonSerializerOptions();
```
martindevans, line 11: The `JsonConverter` attribute ([here](https://github.com/SciSharp/LLamaSharp/blob/d87d654a342179edb32bf85db26c6bec203567df/LLama/Common/ModelParams.cs#L96)) is meant to do this automatically. I'm not sure why that's not working.

Although this change fixes the test failure it doesn't really fix the problem - the `EncodingConverter` is an `internal` type so no users of LLamaSharp would actually be able to do this (and the intention is that serialization should "just work" without any extra config).

xbotter: You are right, I overlooked the fact of "internal".

After .NET 8, JsonSerializer has some [breaking change](https://learn.microsoft.com/en-us/dotnet/core/compatibility/serialization/8.0/metadata-resolving) that caused custom JsonConverter on Encoding to become ineffective. You can refer to the [Ref above for more information.](https://github.com/dotnet/runtime/issues/95893#issuecomment-1851644867).

In this case, either convert EncodingConverter to public, but still need to remember to configure this converter during serialization. Or we need to find another way to serialize the Encoding property.

martindevans: Aha! Thank you, I've been looking through documentation trying to work out what was wrong for half an hour 🤦‍♂️

martindevans: I'll have a fiddle with trying other ways to do it

martindevans: @xbotter I've pushed up an alternative way to handle the `Encoding`. Eagerly converting it into a string, which can then be serialized directly.

martindevans (approved):

LLama/Common/ModelParams.cs
```diff
@@ -92,9 +92,19 @@ public record ModelParams
         /// <inheritdoc />
         public bool VocabOnly { get; set; }
 
+        /// <summary>
+        /// `Encoding` cannot be directly JSON serialized, instead store the name as a string which can
+        /// </summary>
+        [JsonPropertyName("Encoding")]
```
martindevans, line 98:
```suggestion
        [JsonPropertyName("Encoding")]
        [JsonInclude]
```

LLama.Unittest/ModelsParamsTests.cs
xbotter: It looks good. ❤️

martindevans: Over on the issue you linked eiriktsarpalis suggested that all converters should be made public to support the source generator serialization ([ref](https://github.com/dotnet/runtime/issues/95893#issuecomment-1852344280)). Do you want to make that change in this PR too (for the `TensorSplitsCollectionConverter`)?

xbotter: Updated

martindevans:
Looks good to me! I'll merge it once the last test passes

Issue #269, opened 2023-11-09
Create HTTP API server and provide API like OAI


xbotter:
like https://github.com/ggerganov/llama.cpp/tree/master/examples/server

AsakusaRinne:
It's a good idea, but I'm not sure if this is expected to be completed in this project. I have no idea about how to make a high-performance server now. Maybe @Oceania2018 and @saddam213 have some ideas about it? They are the authors for BotSharp and LLamaStack, respectively.

AsakusaRinne:
I think this issue could be separated into two parts. One is the improvement of text-embedding generation mentioned in #239 . The other is how to support using multi-models and multi-context and switch between them with high performance. Not sure how to achieve the second point yet.

saddam213:
The LLamaStack WebApi supports most of LLamaSharp features including state saving etc

Adding a new controller to map the different json inputs outputs should be trivial, assuming there is nothing out of the ordinary in there.

However LLamaStack uses custom executors and state management, so that could cause issues for people looking to use vanilla LLamaSharp.

So maybe best option is to make a version of the WebAPI for LLamaSharp, and add controllers for normal implementations and OAI implementations

1. Modify Web Example to be both Web and WebApi example
2. Modify existing WebApi to add the new Api

Option 2 would require a new project, a project to share commonalities, like the ModelService as that supports multi-models and multi-context

And if we go for option 2 we could even port the WPF app over

xbotter:
The current WebApi can only be used as a sample project for testing. I plan to move it to the Examples project. 
Then, create a new LLamaSharp.WebApi class project that provides webapi similar to OAI, serving as an endpoint  for asp.net core, making it easy and fast to integrate into webapi projects.
However, memory management for model loading is indeed a significant issue.

gianni-rg:
Hello, I'd like to share with you a .NET 8 Minimal API demo project, to locally host LlamaSharp supported models behind an OpenAI-compatible API. It's in very early stages, but it should work as a starting point. Feel free to get inspiration from it. It has been implemented to test open-source coding assistants in VSCode, like continue.dev.

https://github.com/gianni-rg/LlamaSharpApiServer

martindevans:
Since there are already two external projects that do this (and there are probably more out there), perhaps we should _not_ include in in this project and instead just update the readme to link to projects like LlamaStack and LlamaSharpApiServer?

sangyuxiaowu:
[...]

PR #362, opened 2023-12-13
[WIP] refactor: init some experimental refactoring.

AsakusaRinne:
This PR refactors the implementations of LLamaSharp, which is targeted for v1.0.0.

I pushed some changes but has not completed it yet. This PR now is more like a proposal, no promising for compiling and test.

 
### Currently I've made the following changes
1. Add `IGenerationControl` to allow customized control for the stop of generation, instead of just anti-prompts.
2. Add `ITokenizer` to allow customized tokenizer.
3. Allow call stateless executor with self-managed context.
4. Add `TextCompletion`, which is corresponding to `ChatSession`, to provide some further wrapping for text completion tasks. Related issue: #269  #178 

### Some other proposals
1. Make the `InstructExecutor` obsolete because it's only one of the mode of text completion.
2. Allow self-managed kv cache, providing APIs to allocate, update and release the cache. (maybe do it on master?)
3. Apply batch decoding, which will refactor the `StatelessExecutor` again

### For discussion
I'm wondering if we should introduce `Microsoft.SemanticKernel.Abstractions` to LLamaSharp, instead of just using it in SK integration. The reasons are listed as following:
1. Some features such as function call are complex and have been supported or partially supported in SK. We'll have a hard time to support them from scratch because of short of hands and the quick speed of LLM iteration.
2. Some useful abstractions are defined in SK. We could just take use these abstractions even though we have different implementations of the usages. For example, prompt template. It will provide convenience for both users and SK integration developers.
3. SK has been on the way to the standard of LLM in dotnet.

However, though, I also have some worrying about it:
1. We'll have to follow the changes of SK. If SK makes some break changes, we may have to rush to follow it.
2. Our directions of development will be restricted. I'm not sure how will SK be in the future. Though SK is opensource, it has many contents for Azure service because it's developed by MicroSoft. On the contrary, LLamaSharp absolutely does not has this preference.

---

@martindevans @SignalRT @saddam213 May I ask for some ideas from you? This is not a formal PR so please feel free to talk about anything.

### DIFF
--- a/LLama/Abstractions/IInferenceParams.cs
+++ b/LLama/Abstractions/IInferenceParams.cs
@@ -1,7 +1,9 @@
 ﻿using System.Collections.Generic;
 using LLama.Common;
+using LLama.Control;
 using LLama.Native;
 using LLama.Sampling;
+using LLama.Transform;
 
 namespace LLama.Abstractions
 {  
@@ -114,5 +116,15 @@ namespace LLama.Abstractions
 		/// Set a custom sampling pipeline to use. <b>If this is set All other sampling parameters are ignored!</b>
 		/// </summary>
 		ISamplingPipeline? SamplingPipeline { get; set; }
+
+		/// <summary>
+		/// Set a custom generation control to use. <b>If this is set antiprompt will be ignored!</b>
+		/// </summary>
+		IGenerationControl GenerationControl { get; set; }
+
+		/// <summary>
+		/// Set a custom tokenizer to use.
+		/// </summary>
+		ITokenizer Tokenizer { get; set; }
 	}
 }
--- a/LLama/Control/DefaultGenerationControl.cs
+++ b/LLama/Control/DefaultGenerationControl.cs
@@ -1,0 +1,42 @@
+﻿using LLama.Abstractions;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LLama.Control
+{
+    /// <summary>
+    /// The default generation control in LLamaSharp, using antiprompts. This class should not be inherited. 
+    /// <b>Note that this class has state. The previous outputs feeded to it will affect its control.</b>
+    /// If you use it in a session, please don't reuse it for another session unless you intend to do so.
+    /// </summary>
+    public sealed class DefaultGenerationControl: IGenerationControl
+    {
+        private AntipromptProcessor _antipromptProcessor;
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        public DefaultGenerationControl()
+        {
+            _antipromptProcessor = new AntipromptProcessor();
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        public bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, string lastOutputText)
+        {
+            _antipromptProcessor.SetAntiprompts(inferenceParams.AntiPrompts);
+            return _antipromptProcessor.Add(lastOutputText);
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        public bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, IEnumerable<int> lastOutputIds)
+        {
+            return false;
+        }
+    }
+}
--- a/LLama/Control/IGenerationControl.cs
+++ b/LLama/Control/IGenerationControl.cs
@@ -1,0 +1,31 @@
+﻿using LLama.Abstractions;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LLama.Control
+{
+    /// <summary>
+    /// Control the text generation of LLama Executors.
+    /// </summary>
+    public interface IGenerationControl
+    {
+        /// <summary>
+        /// Use the last output text to determine if the generation should stop.
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="inferenceParams"></param>
+        /// <param name="lastOutputText"></param>
+        /// <returns></returns>
+        bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, string lastOutputText);
+
+        /// <summary>
+        /// Use the last output ids to determine if the generation should stop.
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="inferenceParams"></param>
+        /// <param name="lastOutputIds"></param>
+        /// <returns></returns>
+        bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, IEnumerable<int> lastOutputIds);
+    }
+}
--- a/LLama/Transform/DefaultTokenizer.cs
+++ b/LLama/Transform/DefaultTokenizer.cs
@@ -1,0 +1,53 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LLama.Transform
+{
+    /// <summary>
+    /// The default tokenizer of LLamaSharp. This class should not be inherited.
+    /// <b>Note that this class has state. The previous outputs feeded to it will affect its control.</b>
+    /// If you use it in a session, please don't reuse it for another session unless you intend to do so.
+    /// </summary>
+    public sealed class DefaultTokenizer: ITokenizer
+    {
+        private Encoding _encoding;
+        private StreamingTokenDecoder _tokenDecoder;
+
+        /// <summary>
+        /// Initialize a new tokenizer with the specified encoding.
+        /// </summary>
+        /// <param name="encoding"></param>
+        public DefaultTokenizer(Encoding encoding)
+        {
+            _encoding = encoding;
+            _tokenDecoder = new StreamingTokenDecoder(encoding);
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        public IEnumerable<int> Tokenize(LLamaContext context, string text, bool addBos = true, bool special = false)
+        {
+            return context.Tokenize(text, addBos, special);
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        public string Detokenize(LLamaContext context, int token)
+        {
+            _tokenDecoder.Add(token, context.NativeHandle.ModelHandle);
+            return _tokenDecoder.Read();
+        }
+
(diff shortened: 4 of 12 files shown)


martindevans:
> Allow self-managed kv cache, providing APIs to allocate, update and release the cache. (maybe do it on master?)
> Apply batch decoding, which will refactor the StatelessExecutor again
These two will probably be handled by my ongoing work to redesign the executors. My plan is a new executor, built on a single batch, which can have multiple ongoing "conversations" that can all be updated in one single batch.

> I'm wondering if we should introduce Microsoft.SemanticKernel.Abstractions to LLamaSharp, instead of just using it in SK integration
Personally I would be against this, I think we should keep LLamaSharp as close to just being a wrapper around llama.cpp as possible. However, we should definitely make sure the library is flexible enough that something like SK can be integrated with it!

LLama/Abstractions/IInferenceParams.cs
```diff
@@ -114,5 +116,15 @@ public interface IInferenceParams
 		/// Set a custom sampling pipeline to use. <b>If this is set All other sampling parameters are ignored!</b>
 		/// </summary>
 		ISamplingPipeline? SamplingPipeline { get; set; }
+
+		/// <summary>
+		/// Set a custom generation control to use. <b>If this is set antiprompt will be ignored!</b>
+		/// </summary>
+		IGenerationControl GenerationControl { get; set; }
```
martindevans, line 123:
```suggestion
		IGenerationControl? GenerationControl { get; set; }
```

LLama/Abstractions/IInferenceParams.cs
```diff
@@ -114,5 +116,15 @@ public interface IInferenceParams
...
+
+		/// <summary>
+		/// Set a custom tokenizer to use.
+		/// </summary>
+		ITokenizer Tokenizer { get; set; }
```
martindevans, line 128:
```suggestion
		ITokenizer? Tokenizer { get; set; }
```

LLama/Control/DefaultGenerationControl.cs
```diff
@@ -0,0 +1,42 @@
+﻿using LLama.Abstractions;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LLama.Control
+{
+    /// <summary>
+    /// The default generation control in LLamaSharp, using antiprompts. This class should not be inherited.
```
martindevans, line 9:
```suggestion
    /// The default generation control in LLamaSharp, using antiprompts.
```

martindevans: It's sealed, so it's not possible to extend this class.

LLama/Control/DefaultGenerationControl.cs
```diff
@@ -0,0 +1,42 @@
+﻿using LLama.Abstractions;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LLama.Control
+{
+    /// <summary>
+    /// The default generation control in LLamaSharp, using antiprompts. This class should not be inherited. 
+    /// <b>Note that this class has state. The previous outputs feeded to it will affect its control.</b>
```
martindevans, line 10:
```suggestion
    /// <b>Note that this class has state. The previous outputs fed to it will affect its output.</b>
```

LLama/Control/DefaultGenerationControl.cs
```diff
@@ -0,0 +1,42 @@
...
+            _antipromptProcessor.SetAntiprompts(inferenceParams.AntiPrompts);
+            return _antipromptProcessor.Add(lastOutputText);
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        public bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, IEnumerable<int> lastOutputIds)
+        {
+            return false;
```
martindevans, line 39: should this be returning false?

LLama/Control/IGenerationControl.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <summary>
+        /// Use the last output text to determine if the generation should stop.
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="inferenceParams"></param>
+        /// <param name="lastOutputText"></param>
+        /// <returns></returns>
+        bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, string lastOutputText);
+
+        /// <summary>
```
martindevans, line 22: I'm not sure we should have this variant.

Any output tokens are going to be fed through a StreamingTokenDecoder somewhere else. To implement this method you would have to duplicate that work again, just for antiprompt processing!

LLama/Transform/DefaultTokenizer.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LLama.Transform
+{
+    /// <summary>
+    /// The default tokenizer of LLamaSharp. This class should not be inherited.
```
martindevans, line 8:
```suggestion
    /// The default tokenizer of LLamaSharp.
```

LLama/Transform/DefaultTokenizer.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <b>Note that this class has state. The previous outputs feeded to it will affect its control.</b>
```
martindevans, line 9:
```suggestion
    /// <b>Note that this class has state. The previous outputs fed to it will affect its output.</b>
```

LLama/Transform/DefaultTokenizer.cs
```diff
@@ -0,0 +1,53 @@
...
+        public DefaultTokenizer(Encoding encoding)
+        {
+            _encoding = encoding;
+            _tokenDecoder = new StreamingTokenDecoder(encoding);
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        public IEnumerable<int> Tokenize(LLamaContext context, string text, bool addBos = true, bool special = false)
```
martindevans, line 30: I think it would be better to accept a `LLamaWeights` in the constructor, instead of a `LLamaContext` in the Tokenize/Detokenize methids. That simplifies usage and allows you to use the tokenizer without creating an entire context (which is quite memory hungry).

LLama/Transform/StreamingTokenDecoder.cs
```diff
@@ -6,14 +6,14 @@
...
-namespace LLama
+namespace LLama.Transform
 {
     /// <summary>
     /// Decodes a stream of tokens into a stream of characters
     /// </summary>
     public sealed class StreamingTokenDecoder
     {
-        private readonly SafeLlamaModelHandle _weights;
+        private readonly SafeLlamaModelHandle? _weights;
```
martindevans, line 16: I don't think any of the changes in this class are necessary if we go with the suggested change to the `DefaultTokenizer`

martindevans:
Over all I really like the changes 1, 2 and 3. Can we split them into 3 separate PRs so they can proceed separately?

LLama/Transform/DefaultTokenizer.cs
AsakusaRinne: I'll change it, thank you!

LLama/Control/IGenerationControl.cs
AsakusaRinne: The default implementation of it inside LLamaSharp just takes use of `AntiPromptProcessor`. For users it may not be easy to have a good implementation of this interface without reading of source code. However I think the accessibility should be open to users for flexibility.

TBH I'm wondering if you mean to comment at `DefaultTokenizer`, which uses `StreamingTokenDecoder`?

AsakusaRinne:
> Over all I really like the changes 1, 2 and 3. Can we split them into 3 separate PRs so they can proceed separately?

Sure, I'll separate them into 3 PRs.

Besides there're some proposals I forgot to mention above:

1. Support cuda backends with different avx levels.
2. Mix the cuda11 backend and cuda12 backend to one because of the feature detection. Besides, allow users to specify the cuda version in `NativeLibraryConfig` to deal with some complex conditions.
3. Make the `withLog` in `NativeLibraryConfig` be true by defualt because according to the issues, the information of it is very important for users to debug.
4. Refactor the web service implementation after the support for batch decoding.

Also there're some features that makes me headache. I don't mean to complete them in v1.0.0 but just include them in the discussion.
1. Support fine-tune.
2. LLaVA support, or more generally, image2text support. 
3. Catch the error from llama.cpp. It's useful for automatic configuration of layers offloaded to VRAM.

LLama/Control/IGenerationControl.cs
martindevans: > TBH I'm wondering if you mean to comment at DefaultTokenizer, which uses StreamingTokenDecoder?

I really did mean here 😆 

What I'm getting at is **any** implementation of this method:

```csharp
bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, IEnumerable<int> lastOutputIds);
```

_Must_ feed the tokens through a `StreamingTextDecoder` internally (or something equivalent).

So thinking about the entire system you would be detokenizing inside this method (for antiprompting) and also somewhere else (to get the actual output), that's a lot of wasted duplicate work! That's why I'm saying we shouldn't have this overload and it's better to accept the strings directly.

martindevans: Looking at your new PR I've realised that I totally misunderstood how this interface is used, sorry for the confusion!

I _thought_ that `ShouldStopGeneration(tokens)` was an alternative to `ShouldStopGeneration(text)` and internally would be detokenizing. I didn't realise it was intended to process the tokens directly!

LLama/Transform/DefaultTokenizer.cs
AsakusaRinne: Is there already a way to tokenize without a context? I didn't find a such method

martindevans:
```csharp
LLamaWeights w;
w.NativeHandle.Tokenize("a string");
```

Should do it. There should really be higher level wrappers for tokenization in `LLamaWeights`, so that you don't have to access the NativeHandle, but I haven't built them yet.

AsakusaRinne: I see, I mistook the one in NativeApi with context as parameter as the lowest level api. I'll add a wrapper for it with LLamaWeights. :)

PR #364, opened 2023-12-14
feat: support custom generation control of executors.

AsakusaRinne:
Related discussion: #362


LLama/Control/DefaultGenerationControl.cs
```diff
@@ -0,0 +1,42 @@
...
+            _antipromptProcessor.SetAntiprompts(inferenceParams.AntiPrompts);
+            return _antipromptProcessor.Add(lastOutputText);
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        public bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, IEnumerable<int> lastOutputIds)
+        {
+            return false;
```
martindevans, line 39: maybe check for the EOS token here?

LLama/Control/IGenerationControl.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="inferenceParams"></param>
+        /// <param name="lastOutputText"></param>
+        /// <returns></returns>
+        bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, string lastOutputText);
+
+        /// <summary>
+        /// Use the last output ids to determine if the generation should stop.
+        /// </summary>
```
martindevans, line 24:
```suggestion
        /// </summary>
        /// <remarks>This is run before the token is coverted into text. The token will <b>not</b> be returned if this returns true</remarks>
```

LLama/Control/IGenerationControl.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <param name="context"></param>
+        /// <param name="inferenceParams"></param>
+        /// <param name="lastOutputIds"></param>
+        /// <returns></returns>
+        bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, IEnumerable<int> lastOutputIds);
```
martindevans, line 29: Can this accept a single token instead? That gives it finer control (i.e. it can choose exactly _which_ token to stop at) and also is more efficient (at the moment most call sites will only have one token and will need to allocate a single item array for it).

martindevans: (Maybe also add an extension method helper with the current signature, which just loops over all tokens calling `ShouldStopGeneration` for each one).

LLama/Control/IGenerationControl.cs
martindevans: See comment in LLamaStatelessExecutor before merging this!

LLama/LLamaStatelessExecutor.cs
```diff
@@ -105,13 +106,16 @@ await Task.Run(() => { Context.Eval(tokens, 1); }, cancellationToken)
...
+                    break;
+
                 // Decode this token into text
                 decoder.Add(id);
                 var decoded = decoder.Read();
                 yield return decoded;
 
                 // Check if any of the antiprompts have been generated
-                if (antiprocessor.Add(decoded))
+                if(control.ShouldStopGeneration(Context, inferenceParams, decoded))
```
martindevans, line 118: This block doesn't look quite right, although I'm not sure exactly what the behaviour should be!

If `ShouldStopGeneration(token)` (line 109) returns true then the generation immediately stops. The token is never decoded and is never returned to the user.

If `ShouldStopGeneration(string)` (line 118) returns true then the token has already been converted into a string and has been returned to the user. However, the token is **not** added to `lastTokens`.

I think `yield return decoded` and `lastTokens.Add` should always happen atomically (i.e. either both run, or neither do) -  i.e. should line 118 be moved down below `lastTokens.Add`?

This isn't really a problem in the stateless executor, but I'm concerned if the same thing is a problem in the other executors whree the control flow is more complex?

martindevans:
Overall looks good!

Over in the other PR I was confused about how the `ShouldStopGeneration(token)` method works and what I was talking about there was wrong. Desregard all that :)

LLama/Control/IGenerationControl.cs
AsakusaRinne: I used `IEnumerable<int>` because in some conditions one word is composed by two or more tokens. However after some investigation, I think it's not a common stage because when it happens, it means the model has few data of a certain language. Therefore the output of the model can't be good enough. The best choice for users is to use a model trained with their languages. I'll change it to use a single Id. :)

LLama/Control/DefaultGenerationControl.cs
AsakusaRinne: I've added it, thank you for the reminder.

LLama/LLamaStatelessExecutor.cs
AsakusaRinne: Yes you're right! I've changed the implementation here.

LLama/Control/IGenerationControl.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Use the last output text to determine if the generation should stop.
+        /// This method will be called after the overload with output id.
+        /// The text will be returned even if this returns true but the generation will be stopped.
+        /// </summary>
+        /// <param name="context">The LLamaContext used in the current generation.</param>
+        /// <param name="inferenceParams">The inference params used in the current generation.</param>
+        /// <param name="lastOutputText">The last output text generated by the model.</param>
+        /// <returns></returns>
+        bool ShouldStopGeneration(LLamaContext context, IInferenceParams inferenceParams, string lastOutputText);
```
martindevans, line 22: Just a thought - is it possible to decouple this interface from the `IInferenceParams` somehow? It feels a bit odd that `GenerationControl` is overriding parts of `IInferenceParams` and is _also_ configured by it.

The obvious way to do that would be to remove the `inferenceParams` parameter here, that would introduce problems in `DefaultGenerationControl` though.

martindevans (approved):

LLama/Control/IGenerationControl.cs
AsakusaRinne: Yeah I also thought of that and I could come up with the following alternatives:

1. Change the API of `ILLamaExecutor` to `InferAsync(string text, IInferenceParams? inferenceParams = null, IGenerationControl? control = null, CancellationToken token = default);` or similar formats. It will introduce some break changes though. 
2. Use generic type. Change the `ILLamaExecutor` to `ILLamaExecutor<TControl>`, or change the method `InferAsync`.
3. Use chained-call. For example, in the sight of users, they could do `var executor = new StatelessExecutor().WithControl()`.

Each of them has advantages and disadvantages. The most important one I think is that whether we should make it class level or method level. Actually I can't think up a case that the user must use different control strategies in different calls of an executor, though it does have such flexibility. If to get a compromise of it, I'd like to suggest the following proposal:

1. Use the option 3 above, the chained-call for `ILLamaExecutor`, taking `WithGenerationControl` as an extension method.
2. Define a static method to execute the generation, just like what I did in #362. In this static method, we could introduce any break change, like option 1 above. Since dotnet 7, we could define a static method in the interface but in older version it's not supported. Therefore if taking this approach, some parts of the design may look a little awkward.

BTW there's one thing related with it, the sampling pipeline. Currently `InferenceParams` contains many params of sampling, while specifying `SamplingPipeline` will totally override them. It's completely okay in the master branch because we don't want to introduce substantial break changes now. However I'm wondering if we should refactor it before v1.0.0.
For example, we could add many kinds of sampling pipelines as we discussed [here](https://github.com/SciSharp/LLamaSharp/pull/348#issuecomment-1848427031), along with a class named `CustomSamplingPipeline`, which accepts all the parameters related with samping in `InferenceParams` now.

martindevans: I was thinking more along the lines of removing the `IInferenceParams inferenceParams` from the `ShouldStopGeneration` methods.

martindevans: > It's completely okay in the master branch because we don't want to introduce substantial break changes now. However I'm wondering if we should refactor it before v1.0.0.

Yeah this is my thinking. It feels like we're moving in a direction where we can completely get rid of the inference params in the future (which is part of why I want to avoid depending on it in this new interface, if possible).

martindevans:
I've fixed the merge conflict. Waiting for tests now.

AsakusaRinne:
Sorry for leaving this PR incomplete... Since already two months past, the design of this PR may need to be re-considered. Therefore I convert it to draft PR now

Issue #350, opened 2023-12-09
Support cublas computation without requiring CUDA installed


AsakusaRinne:
related issue: #345

martindevans:
llama.cpp copies cudart, cublas and cublasLt64 into the release package.

See here: https://github.com/ggerganov/llama.cpp/blob/master/.github/workflows/build.yml#L497

AsakusaRinne:
What about splitting the packages into three kinds in the future?

- LLamaSharp.Core: only the main package of LLamaSharp, no native library.
- LLamaSharp: main package and all native libraries.
- LLamaSharp.Cuda: main package, native libraries and cublas dependencies, which does not require users to have cuda installed.

SignalRT:
@AsakusaRinne We should think also on LLaVa (or more generic any multimodal model). Right now this is not managed by the llama.cpp libraries but with an additional library. Until now there is no GPU so it's simple to go with only one package, but  I suppose that in the near future this will change.

AsakusaRinne:
@SignalRT I noticed that LLaVA requires another model named CLIP and llama.cpp doesn't provide C APIs for LLaVA but only an example. Do you have any idea about how to support it?

SignalRT:
@AsakusaRinne Yes, I have a version working, but not yet integrated cleanly on the currrent Executor architecture.

martindevans:
> but not yet integrated cleanly on the currrent Executor architecture.

Just a note on that, the current executors will probably be replaced at some point when we swap over to batch decoding. Just a heads up so you don't spend a tonne of effort on that!

> what about splitting the packages into three kinds in the future?

If we're going to split it I'd say there should be an option for CPU-only inference to be installed without any CUDA backend. Just because the CUDA binaries are huge.

Onkitova:
Hello again.
Was tinkering a little bit with https://github.com/sandrohanea/whisper.net which is also a c# binding, but for another ggerganov project -- [whisper.cpp](https://github.com/ggerganov/whisper.cpp) -- not llama.cpp. 
The thing is: it also utilizes cuda through cublas. At first, it looked like it won't work without CUDA Toolkit installed as well (our current state), but then I tried [a method from issue discussion](https://github.com/sandrohanea/whisper.net/issues/65#issuecomment-1610996912) and... it accidentally worked.

> ![image](https://github.com/SciSharp/LLamaSharp/assets/119720012/cbf0aa76-2433-4c24-9c6d-d0b0116d517d)

[...]

PR #365, opened 2023-12-14
Using CUDA while decoupling from the CUDA Toolkit as a hard-dependency

Onkitova:
Using CUDA while decoupling from the CUDA Toolkit as a hard-dependency

Possible solution for https://github.com/SciSharp/LLamaSharp/issues/350

Adding an alternative, fallback method of detection system-supported cuda version to make CUDA Toolkit installation optional. Technically, it uses output of the command line tool "nvidia-smi" (preinstalled with nvidia drivers), which also contains information about cuda version supported on system.

Can confirm it works only on Windows, but I suppose that similar approach can be utilized for Linux and MacOS as well. Didn't touch the code for these 2 platforms, nevertheless.

After that, cuda can be utilized simply by putting nvidia libraries from llama.cpp original repo, "bin-win-cublas-cu12.2.0-x64.zip" asset to the root folder of the built program. For example, to folder: "\LLama.Examples\bin\Debug\net8.0\".

### DIFF
--- a/LLama/Native/NativeApi.Load.cs
+++ b/LLama/Native/NativeApi.Load.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 
 namespace LLama.Native
 {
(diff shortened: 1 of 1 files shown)


AsakusaRinne (approved):
The overall looks good to me. Thank you for this contribution! Now it's much easier for us to support the case in #350

PR #373, opened 2023-12-16
Improved support for AVX512

martindevans:
- Enabled more features in build process (VBMI and VNNI)
   - I haven't rebuilt the binaries yet, so this improvement will be picked up in the next binary update
 - Added runtime checking for these new features
 - Improved runtime checking to no longer require dotnet8.0!


martindevans:
Running the example project with net6.0 now successfully autodetects and loads AVX512 for me 🥳

PR #378, opened 2023-12-18
Fix broken references in docs 

asmirnov82:
After commit "docs: update the docs to follow new version" references in web documentation under LLamaContext node  are not valid (404 file not found). Fix this


martindevans:
Thanks for this. The docs have fallen behind a bit due to the pace of updates in the last few updates, definitely some fixes needed there!

AsakusaRinne (changes requested):
Thank you so much for this contribution! There's an [index file](https://github.com/SciSharp/LLamaSharp/blob/master/mkdocs.yml) to specify the location of the markdown documents, could you please help to update them since you renamed some files?

asmirnov82:
> Thank you so much for this contribution! There's an [index file](https://github.com/SciSharp/LLamaSharp/blob/master/mkdocs.yml) to specify the location of the markdown documents, could you please help to update them since you renamed some files?

It looks like this file was changed in commit "docs: update the docs to follow new version" without changing the real location of MD files. That's why I renamed them to reflect the changes in mkdocs.yml. After this PR the path to MD files should be consistent with the content of mkdocs.yml

AsakusaRinne:
Sorry I didn't notice that before. Now all thing looks good! :)

AsakusaRinne (approved):

PR #404, opened 2024-01-02
LLamaToken Struct

martindevans:
This is a major change to the API. All uses of `int` (or `llama_token`, aliasing `int`) to represent a token have been replaced with a `LLamaToken` struct. This makes the API more strongly typed - it's less likely to mix up a token with some other non-token integer.

In general fixing any breakage is easy, you can just cast `(int)the_token` or `(LLamaToken)1`. Not many places in the basic API deal in raw tokens, and usually work on strings instead, so I'm actually not expecting this to seriously affect many users.


martindevans:
I'd like to get this one reviewed soon if possible, it touches a lot of files and so it will rapidly become a pain to merge if other PRs are merged.

saddam213:
Sorry didn't see this request, 

changes look fine to me :) however I don't know the lower levels that well

PR #460, opened 2024-01-27
Binary Update Jan 2024

martindevans:
Built with https://github.com/SciSharp/LLamaSharp/actions/runs/7674546118/job/20919306901, llama.cpp commit [`a1d6df129bcd3d42cda38c09217d8d4ec4ea3bdd`](https://github.com/ggerganov/llama.cpp/tree/a1d6df129bcd3d42cda38c09217d8d4ec4ea3bdd)

PR #465, opened 2024-01-30
Use llama instead of libllama in `[DllImport]`

### LINKED ISSUE #463: DllImport should not be prefixed with lib
The usage of `[DllImport("libllama")]` results in the windows binaries being prefixed with "lib".
The native libraries in the NuGet packages have renamed DLLs. This makes it annoying to copy in llama binary releases for windows, for testing purposes. If the DLL imports were replaced with just `[DllImport("llama")]`, this issue would not exist on windows, and on other platforms libllama.so and dylibs would work too, even with the lib prefix.

From: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/native-library-loading

> When running on Windows, the DLL is searched for in the following order:
>
> `nativedep`
> `nativedep.dll` (if the library name does not already end with .dll or .exe)
>
> When running on Linux or macOS, the runtime will try prepending lib and appending the canonical shared library extension. On these OSes, library name variations are tried in the following order:
>
> `nativedep.so` / `nativedep.dylib`
> `libnativedep.so` / `libnativedep.dylib`
> `nativedep`
> `libnativedep`

jasoncouture:
This results in windows users not needing to rename the DLL. This allows native llama builds to be dropped in, even on windows.

I also took the time to update the documentation, removing references to renaming the files, since the names now match.

Fixes #463


martindevans:
This looks like a great fix, thanks very much.

jasoncouture:
Looks like it needs a small tweak to the manual path search logic to work on OS X/Linux, I'll get that pushed soon.

jasoncouture:
Should be fixed in that commit:
image

martindevans:
Thanks!

I'll run a new build and update #460 with new binaries right now.

jasoncouture:
@martindevans when I run this in my fork, it always uses master anyway.
I'm looking into it.

Issue #464, opened 2024-01-30
Enable OpenCL/ROCm


jasoncouture:
llamacpp supports AMD GPUs with ROCm and OpenCL via clblas.
We should create an additional NuGet package for this, and update the native loader to support them.

martindevans:
Definitely a good idea, are you willing to work on the modifications to the Github actions pipeline to compile the necessary binaries? That's the first step to getting this supported.

jasoncouture:
I have already started working on it! 😄

PrestigeDevop:
I was trying to deploy llm unity to android , thought about : 
avx2 with nemo intrinsics  ( idk how ) 
vulkan backend with compute shader  ( this partially implemented by other )
onnx runtime with   custom model quantization   ( unity doesn't support .net core  bcl -clr yet)

martindevans:
Now that the CLBLAST binaries are included in the project (or at least, will be once #479 is merged) the next step is to create a [nuspec file](https://github.com/SciSharp/LLamaSharp/tree/master/LLama/runtimes/build) and make sure it's covered by the [deployment automation script](https://github.com/SciSharp/LLamaSharp/blob/master/.github/prepare_release.sh). Are you up for working on that @jasoncouture?

martindevans:
According to [this comment](https://github.com/SciSharp/LLamaSharp/pull/479#issuecomment-1925250945) clblast.dll may need to be included in the distributed package.

jasoncouture:
> Now that the CLBLAST binaries are included in the project (or at least, will be once #479 is merged) the next step is to create a [nuspec file](https://github.com/SciSharp/LLamaSharp/tree/master/LLama/runtimes/build) and make sure it's covered by the [deployment automation script](https://github.com/SciSharp/LLamaSharp/blob/master/.github/prepare_release.sh). Are you up for working on that @jasoncouture?

Indeed. I'm having some trouble getting it working, which is why no pr yet.


> According to this comment clblast.dll may need to be included in the distributed package.

Alright.

moozoo64:
Please consider LLamaSharp.Backend.Vulkan

I tried using the llama.cpp prebuilt vulkan dlls
The "new ModelParams(modelPath){....}" works and produces:

ggml_vulkan: Found 1 Vulkan devices:
Vulkan0: AMD Radeon VII | uma: 0 | fp16: 1 | warp size: 64

But "using var model = LLamaWeights.LoadFromFile(parameters);"
Gives
[...]

PR #468, opened 2024-01-30
Add CLBLAST native library to native libraries build

jasoncouture:
Build llama against CLBLAST (windows only for now)

Passing build: https://github.com/jasoncouture/LLamaSharp/actions/runs/7714965548/job/21028424704

Step 1 of #464 
Step 2 will be to update the binaries, and add a new NuGet package spec.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -199,17 +199,8 @@ jobs:
         run: |
           ls -R
 
-          mkdir deps
-
-          mkdir deps/avx
-          mkdir deps/avx2
-          mkdir deps/avx512
-
-          mkdir deps/osx-arm64
-          mkdir deps/osx-x64
-
-          mkdir deps/cu11.7.1
-          mkdir deps/cu12.1.0
+          # Make all directories at once
+          mkdir --parents deps/{avx,avx2,avx512,osx-arm64,osx-x64,cu11.7.1,cu12.1.0,clblast}
 
           cp artifacts/llama-bin-linux-noavx-x64.so/libllama.so  deps/libllama.so
           cp artifacts/llama-bin-linux-avx-x64.so/libllama.so    deps/avx/libllama.so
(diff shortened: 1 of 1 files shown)


.github/workflows/compile.yml
```diff
@@ -11,6 +11,7 @@ on:
 env:
   # Compiler defines common to all platforms
   COMMON_DEFINE: -DLLAMA_NATIVE=OFF -DLLAMA_BUILD_TESTS=OFF -DLLAMA_BUILD_EXAMPLES=OFF -DLLAMA_BUILD_SERVER=OFF -DBUILD_SHARED_LIBS=ON
+  LLAMA_CPP_COMMIT: a1d6df129bcd3d42cda38c09217d8d4ec4ea3bdd
```
martindevans, line 14: Good idea, I've thought about adding something like this but never quite got around to it.

Would it be possible to accept this as a `workflow_dispatch` input so it can be input from the UI instead of requiring a modification to this file?

jasoncouture: Split off into #470

martindevans: Merged :)

martindevans: Oh you'll probably need to rebase this onto master

jasoncouture: Yup, done.

jasoncouture: Probably want that github event variable fix first though #471

jasoncouture: My motiviation for #470 is that the current master in llama.cpp is broken, it causes the test host to segfault when loading a model.
I suspect they changed the model format again, making the current test model incompatible.

martindevans:
Is this good to merge now? I can see you've got the Linux stuff in there but commented out, do you want to include that in this PR, or merge as is and leave it for the future?

jasoncouture:
@martindevans it should be good to go.

Sorry for the close and reopen, fat fingers on my phone.

jasoncouture:
Llamacpp doesn't build clblast on Linux, I'll figure that out eventually, but fo now windows only.

martindevans:
Thanks again!

PR #469, opened 2024-01-30
Fix missing library name prefix for cuda

jasoncouture:
In #465 the native import, and library search path was updated, the native import had the lib prefix removed, and the search path code was updated to prepend `lib` to the name when not running on windows.

However, there was a place in the cuda path generation when not skipping checks that was not using the prefix. This PR Fixes that missing prefix.

### DIFF
--- a/LLama/Native/NativeApi.Load.cs
+++ b/LLama/Native/NativeApi.Load.cs
@@ -195,12 +195,12 @@ namespace LLama.Native
                 else if (cudaVersion == 11)
                 {
                     Log($"Detected cuda major version {cudaVersion}.", LogLevel.Information);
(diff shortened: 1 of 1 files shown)


martindevans:
Could you briefly explain what this is fixing/adding? Thanks.

jasoncouture:
> Could you briefly explain what this is fixing/adding? Thanks.

In #465 the native import, and library search path was updated, the native import had the lib prefix removed, and the search path code was updated to prepend `lib` to the name when not running on windows.

However, there was a place in the cuda path generation (Skip checks) that was not using the prefix. This PR Fixes that missing prefix.

jasoncouture:
Correction: Outside of Skip checks. SkipChecks was correct.

PR #470, opened 2024-01-30
Checkout specific ref for llamacpp when building native libs

jasoncouture:


martindevans:
Thanks

jasoncouture:
When I run this in my branch, it ends up always using master. I'm debugging it now.

martindevans:
Oops, I guess I was a bit too quick on the merge there 😆

martindevans:
@jasoncouture it looks like it should be `github.event.inputs.` instead of `github.input.events.` according to [docs](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs)

PR #471, opened 2024-01-30
Fix incorrect event input variable name

jasoncouture:
In the build workflow, when building deps, the github variable was specified as `input.events`, instead of `event.inputs`
So this PR fixes that issue.


martindevans:
By the way are you on the [Discord server](https://discord.gg/mDbPszsPzC)? I can give you the developer role there if you ping me (same username).

PR #485, opened 2024-02-04
KernelMemory EmbeddingMode bug correction

zsogitbe:


martindevans:
What's the purpose of this change?

AI-TOOLKIT:
> What's the purpose of this change?

It is a bug correction.  You can test why it is needed by running the KernelMemory example (it will crash without it).

zsogitbe:
> What's the purpose of this change?

It is a bug correction. You can test why it is needed by running the KernelMemory example (it will crash without it).

martindevans:
Thanks!

PR #487, opened 2024-02-05
CLBlast for linux

jasoncouture:
As part of #464 I had hoped to support linux as well as windows, and I was able to (after some digging through llamacpp source) figure out how to do it.

This PR adds clblast on linux, and in addition to that, also copies the shared library for clblast to the artifact. This _should_, *but may not always* allow clblast to function without any additional local installations. Users may still need to install other dependencies.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -92,7 +92,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [
-#          ubuntu-20.04, # This is not built in llalam.cpp, so for now we'll skip it
+          ubuntu-22.04,
           windows-latest
         ]
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/compile.yml
```diff
@@ -92,7 +95,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [
-#          ubuntu-20.04, # This is not built in llalam.cpp, so for now we'll skip it
+          ubuntu-22.04,
```
martindevans, line 98: I think this needs to remain 20.04. We moved down from 22.04 a while ago for compatibility reasons (https://github.com/SciSharp/LLamaSharp/commit/b52422a0280c638d1fd3f0f65b6eb90f261f4cc0).

martindevans: (Remember to change all of the `matrix.os == 'ubuntu-22.04` lines below)

martindevans (changes requested):

martindevans:
Just one small review change. Other than that this looks great, thanks!

.github/workflows/compile.yml
jasoncouture: Unfortunately, in this case, 22.04 is required. As libclblast-dev does not exist prior to 22.04, at least not in the standard package repos.

https://packages.ubuntu.com/search?keywords=libclblast-dev&searchon=names

martindevans: Ok, that's not a major problem. It only affects the new binaries.

martindevans (approved):

PR #496, opened 2024-02-07
Smaller Unit Test Model

martindevans:
Switched to `Q3_K_S` for unit test model, instead of `Q4`. This is almost 1gb smaller and _may_ make the macos tests less flakey.


martindevans:
MacOS tests passed: 1

PR #497, opened 2024-02-07
Rework how native libraries are loaded

jasoncouture:
The way we currently load libraries doesn't handle the case where we load a library, but it doesn't initialize due to a lazy loaded dependnecy, or some other issue.

This PR reworks that to use SetDllImportResolver instead. Which we then use to dynamically unload libraries that don't initialize properly. This makes startup much more resilient.

One breaking change:
`WithLibrary` was intentionally renamed to `WithLibraryPath`, since originally it took a file name, and must now take a path instead.

I also added some missing CLBlast stuff

### DIFF
--- a/LLama/LLamaSharp.Runtime.targets
+++ b/LLama/LLamaSharp.Runtime.targets
@@ -28,6 +28,14 @@
         <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         <Link>runtimes/win-x64/native/cuda12/llama.dll</Link>
       </None>
+      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/clblast/llama.dll">
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        <Link>runtimes/win-x64/native/clblast/llama.dll</Link>
+      </None>
+      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/clblast/clblast.dll">
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        <Link>runtimes/win-x64/native/clblast/clblast.dll</Link>
+      </None>
 
       <None Include="$(MSBuildThisFileDirectory)runtimes/deps/libllama.so">
         <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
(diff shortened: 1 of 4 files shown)


LLama/LLamaSharp.Runtime.targets
```diff
@@ -28,6 +28,14 @@
         <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         <Link>runtimes/win-x64/native/cuda12/llama.dll</Link>
       </None>
+      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/clblast/llama.dll">
```
martindevans, line 31: Can all the changes in this file be pulled out into a separate PR?

martindevans:
It seems like there are several changes wrapped up in one here, since this is one of the most delicate parts of the whole library is it possible to split up into several smaller PRs?

e.g.
1. Additional CLBLAST bits
2. General cleanup (some of the changes look like they're neatening up how NativeApi.Load.cs is written, which is great, and would make subsequent changes here easier to review).
3. Unloading and retrying if loading fails
4. Migrate to `SetDllImportResolver`

jasoncouture:
> It seems like there are several changes wrapped up in one here, since this is one of the most delicate parts of the whole library is it possible to split up into several smaller PRs?
> 
> e.g.
> 
> 1. Additional CLBLAST bits
> 2. General cleanup (some of the changes look like they're neatening up how NativeApi.Load.cs is written, which is great, and would make subsequent changes here easier to review).
> 3. Unloading and retrying if loading fails
> 4. Migrate to `SetDllImportResolver`

1 is absolutely doable, I debated doing it last night, but there was one part tied in with the rest. I'll break that off today and rebase this on top of it.

2, 3 and 4 are pretty tied together. 3 is especially challenging because if we load an additional library indirectly, we don't have it's handle to unload it. The general cleanup was pretty much just deleting what wasn't used after switching to import resolver.

martindevans:
Alright, I thought that would probably be the case. In that case please split out the CLBLAST bits for now, and we can come back to the new loading stuff afterwards.

As for the other bits, I'm not really sure if unloading after a failed load is actually what we want? For example, if you have installed the cuda backend package and it fails to load for some reason I would say it's _better_ to hard fail as quickly as possible (preferably with a helpful error message saying why) rather than silently falling back to CPU inference. I whether it fails fast or makes a "best-effort" could be something that's configurable in `NativeLibraryConfig` perhaps?

martindevans:
Another thought, it sounds like 2+4 could be done spearately from 3? if so that'd be good to break up into smaller pieces.

jasoncouture:
> As for the other bits, I'm not really sure if unloading after a failed load is actually what we want? For example, if you have installed the cuda backend package and it fails to load for some reason I would say it's _better_ to hard fail as quickly as possible (preferably with a helpful error message saying why) rather than silently falling back to CPU inference. I whether it fails fast or makes a "best-effort" could be something that's configurable in `NativeLibraryConfig` perhaps?

Isn't that what AllowFallback should control? I assumed that was it's purpose, and implemented accordingly. Although, if that's enabled then it should throw on the first failure.

martindevans:
Oh right, sorry, I didn't realise there was already an option for that

jasoncouture:
> Oh right, sorry, I didn't realise there was already an option for that

It's a good call-out though, in the catch block, if that is set, we shouldn't unload and retry. Because that's falling back.

jasoncouture:
> Another thought, it sounds like 2+4 could be done spearately from 3? if so that'd be good to break up into smaller pieces.

That should be doable. So we'll have 3 PRs then:
1. Wire up OpenCL
2. Hook up dll import resolver, without unload and retry logic.
3. Add unload and retry logic

martindevans:
Given the discussion above I'll close this PR now

PR #510, opened 2024-02-15
Add LLamaSharp.Backend.Vulkan

moozoo64:
.WithVulkan is by default false because I wasn't sure how to test for the presence of vulkan and hence it would fail the CI Tests.
One idea, without requiring the vulkan sdk, is to call and parse the result of "vulkaninfo --summary"

I had trouble with Jimver/cuda-toolkit@v0.2.14 not working so I changed it to match llama.cpp Jimver/cuda-toolkit@v0.2.11
I have only tested this LLamaSharp.Backend.Vulkan under windows with a Radeon VII.


martindevans:
This is great, thanks very much! I'll give it a full review soon.

There's one small issue I immediately spotted; I'm trying to keep all binaries committed traceable back to a build action run in this repo. Could you delete the committed binaries and squash the commits. They'll get added back next time a binary update is done.

PR #514, opened 2024-02-15
Add LLamaSharp.Backend.Vulkan #2

moozoo64:
.WithVulkan is by default false because I wasn't sure how to test for the presence of vulkan and hence it would fail the CI Tests.
One idea, without requiring the vulkan sdk, is to call and parse the result of "vulkaninfo --summary"

I had trouble with Jimver/cuda-toolkit@v0.2.14 not working so I changed it to match llama.cpp Jimver/cuda-toolkit@v0.2.11
I have only tested this LLamaSharp.Backend.Vulkan under windows with a Radeon VII.


martindevans:
That's fine, I'll add new binaries next I do the update of LLamaSharp to a new llama.cpp version. Usually about once a month.

.github/workflows/compile.yml
```diff
@@ -172,7 +172,85 @@ jobs:
...
+      matrix:
+        os: [
+          ubuntu-22.04,
+          windows-latest
+        ]
+    env:
+      OPENBLAS_VERSION: 0.3.23
+      OPENCL_VERSION: 2023.04.17
+      CLBLAST_VERSION: 1.6.0
+      VULKAN_VERSION: 1.3.261.1
```
martindevans, line 189: Are all of the defines required? It looks like only `VULKAN_VERSION` is used?

martindevans:
Apart from that one comment this all looks good, thanks again for putting in the work!

.github/workflows/compile.yml
moozoo64: > Are all of the defines required? It looks like only `VULKAN_VERSION` is used?

Well no... I copied that from "compile-clblast:"....
It just needs VULKAN_VERSION:
Your compile.yml is inverted compared to ggerganov/llama.cpp build.yml. Which builds a lot of the backends together per os.
I think this way will be a lot more verbose when adding all of the backends. eg kcompute, openBlas

martindevans:
Note: use this branch as the base of the next binary-update branch

moozoo64:
Sorry I didn't know committing to my fork would affect this pull request.
Working with GitHub is new to me. 
So, I withdraw it until I can work it out.

PR #517, opened 2024-02-18
Add LLamaSharp.Backend.Vulkan #3

moozoo64:

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -172,7 +172,85 @@ jobs:
             ./build/libllama.so
             # ./build/libclblast.so
           name: llama-bin-linux-clblast-x64.so
-          
+  
+  compile-vulkan:
+    name: Compile (vulkan) - ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [
+          ubuntu-22.04,
+          windows-latest
+        ]
+    env:
+      OPENBLAS_VERSION: 0.3.23
+      OPENCL_VERSION: 2023.04.17
+      CLBLAST_VERSION: 1.6.0
+      VULKAN_VERSION: 1.3.261.1
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Clone
+        id: checkout
+        uses: actions/checkout@v4
+        with:
+          repository: ggerganov/llama.cpp
+
+      - name: Download dependencies - Linux
+        if: ${{ matrix.os == 'ubuntu-22.04' }}
+        run: |
+          wget -qO- https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo tee /etc/apt/trusted.gpg.d/lunarg.asc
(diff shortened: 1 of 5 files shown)


martindevans:
Note to self: **Tests failing is expected** because the binaries are missing in this PR. Base the next binary update PR on this branch.

moozoo64:
Just some general comments
I have only tested on Windows with an AMD Radeon VII
I'd request that other test this PR in order to validate it.
I just know it works for me and my simple use case.
I've not tested under Linux but I believe it has a high chance of working.

On a system with CUDA the CUDA library should precede the vulkan library and hence be used.
In theory vulkan could be made to work on macos using the Vulkan->Metal emulation. But I don't see the point.

The llama.cpp vulkan backend fully loads into vram. I believe the CLBlast doesn't neither does the kcompute.
In my testing the Vulkan backend is also much faster than the CLBlast
Benchmarks on the llama.cpp site suggest the Vulkan backend is about 60% to 90% as fast as the HIPBlast backend However the HIPBlast backend supports a much limited set of graphic cards.

With
private bool _useVulkan = true;
The presence of vulkan will be checked for via GetVulkanVersion()
This uses the command "vulkaninfo --summary" which on windows with a AMD graphics card is installed along with the driver.
Ditto with Nvidia cards. I don't know about Intel but I assume so.

On linux you might need to use  "sudo apt install vulkan-tools". I needed to do this under wsl on my nvidia gpu system but not on my AMD gpu system.
I've not tested under wsl. I don't believe vulkan works properly this way.

The safest option is _useVulkan = false and to remove the check via GetVulkanVersion()
This however means that the vulkan backend won't automatically be used on systems with it.
I don't fully parse "vulkaninfo -summary", only that there is a vulkan device present.
Obviously this would break if the format of "vulkaninfo -summary" was ever changed.
It would be complicated to workout if there was a vulkan device present and that it could run llama.cpp
The only other options to using "vulkaninfo --summary"
1) have Llamasharp use the Vulkan SDK , which I don't think is desirable.
2) create a small stub program using the Vulkan SDK that simply returns VulkanOK and have that linked against the Vulkan SDK.

martindevans:
@moozoo64 Sorry for the delay with this PR. I'm just starting to look at the next round of binary updates now.

To make that easier, would you mind creating a new PR with _just_ the changes to the github build action. That way I can merge it in without breaking CI, we can merge the rest of the stuff in a separate PR later once the binaries are in place.

m0nsky:
Maybe it would be better to skip the Vulkan device check for now (but leave the code in place), and leave it to the user to enable Vulkan on the NativeLibraryConfig? This way (if I understand it correctly) Vulkan support can be introduced in LLamaSharp, it will just not automatically be selected.

If somebody actively working with Vulkan can maintain and test the parsing of vulkaninfo, the automatic selection could be re-introduced.

I wouldn't mind testing the current GetVulkanVersion() implementation WSL and maintaining the parser, but I only have a NVIDIA GPU and don't test/use Vulkan on a regular basis.

martindevans:
Superseded - m0nsky took this work to completion in #797.

PR #523, opened 2024-02-19
ChatSession: improve exception message

swharden:
This is a small fix, but I noticed that the original exception message contained the word "preceeded" which should be spelled as "preceded"

### DIFF
--- a/LLama/ChatSession.cs
+++ b/LLama/ChatSession.cs
@@ -196,7 +196,7 @@ public class ChatSession
             if (lastMessage is null
                 || lastMessage.AuthorRole != AuthorRole.User)
             {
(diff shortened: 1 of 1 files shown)


martindevans:
Thanks!

swharden:
Thanks for your quick review and for your help maintaining this fantastic project! 🚀 

I'm running across similar typos in messages and variable names. Do you want me to continue to fix these and create PRs? If so, would you prefer a larger PR or a few small ones like this? Also if it's not that important, that's okay too. Thanks for your input!

martindevans:
Absolutely, please do! It's best to keep the PRs small if possible, just to minimise the chance that the changes clash with anything :)

PR #525, opened 2024-02-20
Improve "embeddings" example

swharden:
Previously the embeddings example throws an exception because `EmbeddingMode` was not set to `true`.

This PR fixes that and adds extra documentation to the embeddings example.

image


martindevans:
Thanks again. Our docs and examples have really fallen behind recently, so these fixes are really valuable!

xuzeyu91:
What kind of model should be used for embedding? When I use nomic-embed-text-v1.5.f32.gguf, it will report protected memory, while when I use tinyllama-1.1b-chat.gguf, it can run normally. However, I feel that the returned float array is not correct. When I use the same text for vector matching, the similarity is only 0.42

PR #526, opened 2024-02-21
LLama.Examples: improve model path prompt

swharden:
This PR adds a user settings class which can store/retrieve the last used model path. When users are prompted for a model, they can now leave the field blank and press enter to use the last used model. If an invalid path is provided, the user can attempt to re-enter a valid path (previously an exception was thrown).

The coding assistant example previously offered to download the model for you, but I removed that functionality and added a check and warning message if the model used is not a code llama model. The new configuration class could be extended to add automatic model downloading functionality to all examples if that functionality is desired.

This PR also makes a few refinements to code styling and documentation. Since these example files serves as documentation I'd like to also use top-level namespaces, but I didn't implement that in this PR because it would appear as many lines of code changed and be hard to review. Let me know if you'd like me to do that (in this PR or another one).

image

image

image


swharden:
Note that I also deleted `ConsoleLogger` in this PR because it wasn't used by anything... If it's intended to be there for documentation purposes, let me know and I can add it back and demonstrate its use in one of the examples.

LLama.Examples/Program.cs
```diff
@@ -1,18 +1,24 @@
...
-
-Console.WriteLine("======================================================================================================");
+    """);
 
 NativeLibraryConfig
    .Instance
    .WithCuda()
    .WithLogs();
 
-NativeApi.llama_empty_call();
```
martindevans, line 15: It's probably best to keep there here. That way any exceptions due to DLL loading failures will happen as soon as possible, which will help people debugging issues with CUDA deps etc.

LLama.Examples/RepoUtils.cs
```diff
@@ -1,40 +0,0 @@
...
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-
-namespace LLama.Examples
-{
-    /// <summary>
-    /// Basic logger printing to console
-    /// </summary>
-    internal static class ConsoleLogger
```
martindevans, line 13: Not really a specific request, just thinking out loud: do you think something like this should be implemented in LLamaSharp itself?

LLama.Examples/UserSettings.cs
```diff
@@ -0,0 +1,74 @@
...
+        string path = File.ReadAllText(SettingsFilePath).Trim();
+        if (!File.Exists(path))
+            return null;
+
+        return path;
+    }
+
+    private static void WriteDefaultModelPath(string path)
+    {
+        Console.WriteLine($"!!!!WROTE!!!!!!: {SettingsFilePath}");
```
martindevans, line 21: Looks like a debug message you intended to remove before pushing?

martindevans:
Looks great, definitely a big improvement to usability. I'll certainly appreciated the saved model path when testing stuff!

Just a couple of small nits :)

martindevans:
> top-level namespaces

I've wanted to move to all top level namespaces for a while. Once this one is merged I think I'll put in a PR to at least move the entire example project over to top level namespaces.

I'm less confident about making such a "noisy" change on the main project. The best approach is probably to do it all in one big PR, but that would touch nearly every line of every file :/

LLama.Examples/UserSettings.cs
swharden: Good catch 😅😅😅

swharden:
Thanks for the feedback! I think this is ready to merge now.

> top level namespaces

Note that I goofed the language... the new C# features are _top level statements_ and _file-scoped namespaces_

I think this can be automated project wide by creating `.editorconfig` containing:

```
[*.cs]
csharp_style_namespace_declarations=file_scoped:suggestion
```

Then in visual studio hovering on the squiggly underline will let you fix it project-wide, or you could run

LLama.Examples/ExampleRunner.cs
```diff
@@ -1,8 +1,7 @@
 ﻿using Spectre.Console;
+using LLama.Examples.Examples;
 
-namespace LLama.Examples.Examples;
-
-public class Runner
+public class ExampleRunner
```
martindevans, line 4: The namespace appears to have disappeared from this file altogether

martindevans:
Thanks again for all your work on this!

PR #532, opened 2024-02-22
LLamaSharp.Examples: Document Q&A with local storage

swharden:
This PR adds an example that demonstrates how to use `Microsoft.KernelMemory` with local storage so documents can be analyzed once, saved to disk (as a plain-text vector database), then rapidly reloaded on future runs of the program.

Note that `Microsoft.KernelMemory` (https://microsoft.github.io/kernel-memory/) is a higher-level more feature-rich wrapper for `Microsoft.SemanticKernel.Memory` that was "built on the feedback received and lessons learned from developing Semantic Kernel", so it should probably be the favored strategy for document ingestion and interaction in the future.


martindevans:
Excellent work, thanks again!

PR #555, opened 2024-02-28
Multi modal

SignalRT:
This is a minimal example just to get feedback.

1. It builds the llava dynamic link libraries and upload the binaries. It contains Mac libraries updated just to my test.
2. It's a minimal dirty example. It only includes the minimum native API to work with multi modal models.
3. It includes an example just to try.

@martindevans , @AsakusaRinne. In my understanding the best way to integrate multi-modal models on the library is to have an extensible wrapper to add non-text types (images, video, audio, etc.) to the prompt. 

In the library there is a **Chat** abstraction with a **message** that could be extended, but in the ILLamaExecutor the abstraction manages a **text**:

`IAsyncEnumerable<string> InferAsync(string text, IInferenceParams? inferenceParams = null, CancellationToken token = default);`

I just make a json wrapper (Prompt.cs) to make this example.

**These are my questions:**

1. If there is any interest on integrate LLava on LLamaSharp.
2. If there is interest, to ask about your opinion on the best approach to begin to support not-only-text prompts.
3. The support of different kind of executors. In my initial development I implemented a Stateless executor for llava ( I didn´t upload to simplify the discussion). It nearly duplicates all the Stateless executor code. I feel that this should be simplified. In the example I just try to reuse the InteractExecutor (I know it's not correct, it's just a concept test) just to compare.


martindevans:
Fantastic work!

> If there is any interest on integrate LLava on LLamaSharp.

It's a pretty big expansion of the scope, but I would definitely be interested. Multimodal seems to be the direction all of the big SOTA models are moving in, it seems likely that eventually open models will all be multimodal too.

> If there is interest, to ask about your opinion on the best approach to begin to support not-only-text prompts.

I don't really know much how llava works. From the look of it it's basically converting an image into tokens that can be put into the prompt just like any other tokens?

If so, then I've been working on the new `BatchedExecutor` for the last few months which is currently oriented around tokens not text. If llava is working with tokens it should fit right into that pipeline. In fact it would be great to have llava in place as we're developing that wrapper around `BatchedExecutor`, since it would means we're properly handling multi-modal in the new abstraction right from the start!

HikiWaltz:
I'm very interested!
C# is way better do deal with NDI's and llava + NDI's is awesome!
Personally, I would prefer it worked similar to llama-cpp-python, as in, stateless executor where the prompt receives a string prompt + uri to file. But I guess it would limit the extensibility to later accept audio, video etc given the lack of a wrapper.

I'm not knowlegeable on how multimodals work, but I guess audio, video, images etc are fed into it in pretty different ways and not as a blob, but maintaining a way to use stateless would be apreciated.

Maybe methods to prepare the data in the proper way and then feed them directly along with the message prompt?

I'm just starting with real llm dev, so... Just opinion ^^

AshD:
The Azure OpenAI lib uses url OR base-64 data url string for the images
https://github.com/Azure/azure-sdk-for-net/blob/Azure.AI.OpenAI_1.0.0-beta.12/sdk/openai/Azure.AI.OpenAI/README.md#chat-with-images-using-gpt-4-vision-preview

Same with OpenAI
https://platform.openai.com/docs/guides/vision 

Following the same method, the API would be 
IAsyncEnumerable<string> InferAsync(string Url, IInferenceParams? inferenceParams = null, CancellationToken token = default);

where url is the url to a file or a data url.

HikiWaltz:
> The Azure OpenAI lib uses url OR base-64 data url string for the images https://github.com/Azure/azure-sdk-for-net/blob/Azure.AI.OpenAI_1.0.0-beta.12/sdk/openai/Azure.AI.OpenAI/README.md#chat-with-images-using-gpt-4-vision-preview
> 
> Same with OpenAI https://platform.openai.com/docs/guides/vision
> 
> Following the same method, the API would be IAsyncEnumerable InferAsync(string Url, IInferenceParams? inferenceParams = null, CancellationToken token = default);
> 
> where url is the url to a file or a data url.

What about the text prompt? Is it sent separately?

AshD:
Oops. I forgot about that. Thanks for catching it.
IAsyncEnumerable InferAsync(string prompt, string url, IInferenceParams? inferenceParams = null, CancellationToken token = default);

SignalRT:
In my opinion we need an abstraction to manage any kind of content where the code knows what supports and can understand the content passed to manage it in the proper way.

Something similar to semantic kernel:

https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/SemanticKernel.Abstractions/Contents/KernelContent.cs#L18

image

Or ollama:

```
import ollama

res = ollama.chat(
	model="llava",
	messages=[
		{
			'role': 'user',
			'content': 'Describe this image:',
			'images': ['./art.jpg']
		}
	]
)

print(res['message']['content'])

```

There is something similar included in LLamaSharp is ChatHistory / Message:


```
    public class ChatHistory
    {
        private static readonly JsonSerializerOptions _jsonOptions = new() { WriteIndented = true };

        /// <summary>
        /// Chat message representation
        /// </summary>
        public class Message
        {
            /// <summary>
            /// Role of the message author, e.g. user/assistant/system
            /// </summary>
            [JsonConverter(typeof(JsonStringEnumConverter))]
            [JsonPropertyName("author_role")]
            public AuthorRole AuthorRole { get; set; }

            /// <summary>
            /// Message content
            /// </summary>
            [JsonPropertyName("content")]

            /// <summary>
            /// Create a new instance
            /// </summary>
            /// <param name="authorRole">Role of message author</param>
            /// <param name="content">Message content</param>
            public Message(AuthorRole authorRole, string content)
            {
                this.AuthorRole = authorRole;
                this.Content = content;
            }
        }
```

Once we take this decision the implementation it's easy. There is no way something based on llama.cpp works in a very different way.

zsogitbe:
@SignalRT, I have tried your code, but could not make it work. First I had to replace the library loading code, then adjust the inference code, ... and finally it worked, but produced garbage output. I would suggest simply adding the lava library loading code to the existing llama.dll load and remove your loading completely.
There is something wrong with the processing of the prompt in your example, but I could not find the problem (garbage output). Maybe look at the example from @IntptrMax because that works well.

Concerning the concept, I would choose a simple approach and just use an extra binary content parameter (byte array) together with a type parameter as input without the overhead of serializing/deserializing large json content all the time (image, audio). 
Based on the type we can decide what to do with the binary content.

SignalRT:
@zsogitbe It's tested only on Mac and the binary libraries are only updated on Mac. It's tested with mistral llava 1.6 the prompts are different with llava 1.5

I find simpler to pass a well structured data. It doesn´t mean that you need to serialize the data, it could be URL or whatever, but it's a clear interface. It should be a solution to manage multi-content

zsogitbe:
No problem. It is better to move forward even if it is not perfect. In this way we can help each other. Thank you for investing time in this!
I think that first you should try to make a basic, but working example that we can help better. For this, you need to replace your loading code by just simply adding the loading of lava to the existing loading routine in the library. Lava will be needed anyway, so this is already a step in the right direction (this is how I made your example work for me). Let me know if this is ready and then I will test the code again (I am on Windows + GPU).

Are you sure that the image is serialized well and not destroyed while serializing/deserializing?  Maybe you should try to put it in base64 format if you really want this json format.

AsakusaRinne:
It's a great work! Though I need some time to get start with the code again after disappearing from community for two months for some reason... I'd like to join the discussion.

> If there is any interest on integrate LLava on LLamaSharp.

For me I think it's exciting to use LLaVA in LLamaSharp

> If there is interest, to ask about your opinion on the best approach to begin to support not-only-text prompts.

Users won't refuse one more choice of the input format. The most basic ones are file path and byte array. Besides Base64 string, Bitmap and other kinds of multi-media format may make sense.

Here's a [repo](https://github.com/IntptrMax/LLavaSharp) which implements similar functionality with this PR. I've got contacted with the @IntptrMax and maybe we can get some ideas from him.

SignalRT:
I will close this draft PR and begin to make partial PR to introduce the changes iteratively.

PR #556, opened 2024-02-29
Build Llava binaries

SignalRT:
First step to integrate LLava. Build llava binaries when updating binaries


martindevans:
Just found [this test run](https://github.com/SignalRT/LLamaSharp/actions/runs/8103736979), is that an up to date run?

If so, the output in deps.zip looks fine to me.

SignalRT:
That’s right, It’s the check of the PR

Issue #559, opened 2024-03-01
Switching ChatSessions without writing to file.


eublefar:
Hi

Could you please confirm that this is correct way to switch between chat sessions without writing them to files (for performance reasons)?
https://github.com/eublefar/LLAMASharpUnityDemo/blob/main/Assets/LLamaSharpTestScript.cs

Do you think it's worth to implement it into some interface at ChatSession level? (similar to how it's done at Executor level)

Also one thing I struggled to find is something like `AddMessage` but in non-lazy way with context.Decode happening at the same time, to precompute KV cache before generating the first message.

AsakusaRinne:
Hi, I think it's a correct way with a rough look at your code. 

> Do you think it's worth to implement it into some interface at ChatSession level? (similar to how it's done at Executor level)

From my perspective I'd like to see such an API in `ChatSession`. As you can see, some extra process of the output is performed when switching context, which could be wrapped in an API.

> Also one thing I struggled to find is something like AddMessage but in non-lazy way with context.Decode happening at the same time, to precompute KV cache before generating the first message.

It's a good idea for performance improvement. More generally The `generation` and `prefill` could be separate parts and exposed to users in a certain way. Currently I think there's also a tricky way to do so. When you want to prefill it without generation, you just need to set the last several words of your prompt as antiprompt (but be sure it's alone).

FYI @martindevans

eublefar:
Great, thanks for taking time to look at the code.

I think I could make a PR for these changes if you want.

AsakusaRinne:
It would be great if you'd like to make it😀

PR #560, opened 2024-03-02
Chat session state management

eublefar:
As discussed here https://github.com/SciSharp/LLamaSharp/issues/559

I've added few things:
- `StatefulExecutorBase.AddPromptAsync` - basically runs the text through decode without sampling new tokens.
  I am basically running InferInternal twice with WaitForInput = true, but it's very reliant on specific implementation of child classes and not the abstraction itself, so maybe you'll have some suggestions on how to improve it.
- SessionState record class reperesenting ChatSession state in memory.
- `void LoadSession(SessionState state)` and `SessionState GetSessionState()` for ChatSession class to save session state at arbitrary point.
- `static Task<ChatSession> ChatSession.InitializeSessionFromHistoryAsync(ILLamaExecutor executor, ChatHistory history, CancellationToken cancellationToken = default)` and `Task<ChatSession> ChatSession.ProcessSystemMessage(string content)` to pre-process KV cache.
- Functions to reset states of StatefulExecutors and LlamaContext.
- Example `ChatSessionWithRestart` to show off how it works

I didn't find any ChatSession unit tests so I did not write any.

BR,
Mykyta

### DIFF
--- a/LLama.Examples/Examples/ChatSessionWithHistory.cs
+++ b/LLama.Examples/Examples/ChatSessionWithHistory.cs
@@ -48,6 +48,10 @@ public class ChatSessionWithHistory
 
         Console.ForegroundColor = ConsoleColor.Yellow;
         Console.WriteLine("The chat session has started.");
+        Console.WriteLine("Type 'exit' to end the chat session.");
+        Console.WriteLine("Type 'save' to save the chat session to disk.");
+        Console.WriteLine("Type 'load' to load the chat session from disk.");
+        Console.WriteLine("Type 'regenerate' to regenerate the last response.");
 
         // show the prompt
         Console.ForegroundColor = ConsoleColor.Green;
@@ -55,12 +59,14 @@ public class ChatSessionWithHistory
 
         while (userInput != "exit")
         {
+            // Save the chat state to disk
             if (userInput == "save")
             {
                 session.SaveSession("Assets/chat-with-bob");
                 Console.ForegroundColor = ConsoleColor.Yellow;
                 Console.WriteLine("Session saved.");
             }
+            // Load the chat state from disk
             else if (userInput == "load")
             {
                 session.LoadSession("Assets/chat-with-bob");
--- a/LLama.Examples/Examples/ChatSessionWithHistory.cs
+++ b/LLama.Examples/Examples/ChatSessionWithHistory.cs
@@ -61,6 +61,12 @@ public class ChatSessionWithHistory
                 Console.ForegroundColor = ConsoleColor.Yellow;
                 Console.WriteLine("Session saved.");
             }
+            else if (userInput == "load")
+            {
+                session.LoadSession("Assets/chat-with-bob");
+                Console.ForegroundColor = ConsoleColor.Yellow;
+                Console.WriteLine("Session loaded.");
+            }
             else if (userInput == "regenerate")
             {
                 Console.ForegroundColor = ConsoleColor.Yellow;
--- a/LLama.Examples/Examples/ChatSessionWithRestart.cs
+++ b/LLama.Examples/Examples/ChatSessionWithRestart.cs
@@ -1,0 +1,94 @@
+using LLama.Common;
+
+namespace LLama.Examples.Examples;
+
+public class ChatSessionWithRestart
+{
+    public static async Task Run()
+    {
+        string modelPath = UserSettings.GetModelPath();
+
+        var parameters = new ModelParams(modelPath)
+        {
+            ContextSize = 1024,
+            Seed = 1337,
+            GpuLayerCount = 5
+        };
+        using var model = LLamaWeights.LoadFromFile(parameters);
+        using var context = model.CreateContext(parameters);
+        var executor = new InteractiveExecutor(context);
+
+        var chatHistoryJson = File.ReadAllText("Assets/chat-with-bob.json");
+        ChatHistory chatHistory = ChatHistory.FromJson(chatHistoryJson) ?? new ChatHistory();
+        ChatSession prototypeSession = 
+            await ChatSession.InitializeSessionFromHistoryAsync(executor, chatHistory);
+        prototypeSession.WithOutputTransform(new LLamaTransforms.KeywordTextOutputStreamTransform(
+            new string[] { "User:", "Assistant:" },
+            redundancyLength: 8));
+        var resetState = prototypeSession.GetSessionState();
+
+        ChatSession session = new ChatSession(executor);
+        session.LoadSession(resetState);
+
+        InferenceParams inferenceParams = new InferenceParams()
+        {
+            Temperature = 0.9f,
+            AntiPrompts = new List<string> { "User:" }
+        };
+
+        Console.ForegroundColor = ConsoleColor.Yellow;
+        Console.WriteLine("The chat session has started.");
+
+        // show the prompt
+        Console.ForegroundColor = ConsoleColor.Green;
+        string userInput = Console.ReadLine() ?? "";
+
+        while (userInput != "exit")
+        {
+            if(userInput == "reset")
+            {
+                session.LoadSession(resetState);
+                Console.WriteLine($"History: {session.HistoryTransform.HistoryToText(session.History)}");
+                Console.ForegroundColor = ConsoleColor.Yellow;
+                Console.WriteLine("Session reset.");
+            }
+            else if (userInput == "save")
+            {
+                session.SaveSession("Assets/chat-with-bob");
+                Console.ForegroundColor = ConsoleColor.Yellow;
+                Console.WriteLine("Session saved.");
+            }
+            else if (userInput == "regenerate")
+            {
+                Console.ForegroundColor = ConsoleColor.Yellow;
+                Console.WriteLine("Regenerating last response ...");
+
+                await foreach (
+                    var text
+                    in session.RegenerateAssistantMessageAsync(
+                        inferenceParams))
+                {
+                    Console.ForegroundColor = ConsoleColor.White;
+                    Console.Write(text);
+                }
+            }
+            else
+            {
+                await foreach (
+                    var text
+                    in session.ChatAsync(
+                        new ChatHistory.Message(AuthorRole.User, userInput),
+                        inferenceParams))
+                {
+                    Console.ForegroundColor = ConsoleColor.White;
+                    Console.Write(text);
+                }
+            }
+
+            Console.ForegroundColor = ConsoleColor.Green;
+            userInput = Console.ReadLine() ?? "";
+
+            Console.ForegroundColor = ConsoleColor.White;
+        }
+    }
+}
--- a/LLama.Examples/Examples/ChatSessionWithRestart.cs
+++ b/LLama.Examples/Examples/ChatSessionWithRestart.cs
@@ -37,7 +37,8 @@ public class ChatSessionWithRestart
         };
 
         Console.ForegroundColor = ConsoleColor.Yellow;
-        Console.WriteLine("The chat session has started.");
+        Console.WriteLine("The chat session has started. Write `save` to save session in memory."
+            + " Write `reset` to start from the last saved checkpoint");
 
         // show the prompt
         Console.ForegroundColor = ConsoleColor.Green;
@@ -48,13 +49,13 @@ public class ChatSessionWithRestart
             if(userInput == "reset")
             {
                 session.LoadSession(resetState);
-                Console.WriteLine($"History: {session.HistoryTransform.HistoryToText(session.History)}");
+                Console.WriteLine($"Reset to history:\n{session.HistoryTransform.HistoryToText(session.History)}");
                 Console.ForegroundColor = ConsoleColor.Yellow;
                 Console.WriteLine("Session reset.");
             }
             else if (userInput == "save")
             {
-                session.SaveSession("Assets/chat-with-bob");
+                resetState = session.GetSessionState();
                 Console.ForegroundColor = ConsoleColor.Yellow;
                 Console.WriteLine("Session saved.");
             }
--- a/LLama.Examples/Examples/ChatSessionWithRestart.cs
+++ b/LLama.Examples/Examples/ChatSessionWithRestart.cs
@@ -37,8 +37,11 @@ public class ChatSessionWithRestart
         };
 
         Console.ForegroundColor = ConsoleColor.Yellow;
-        Console.WriteLine("The chat session has started. Write `save` to save session in memory."
-            + " Write `reset` to start from the last saved checkpoint");
+        Console.WriteLine("The chat session has started. Starting point saved.");
+        Console.WriteLine("Type 'exit' to end the chat session.");
+        Console.WriteLine("Type 'save' to save chat session state in memory.");
+        Console.WriteLine("Type 'reset' to reset the chat session to its saved state.");
+        Console.WriteLine("Type 'answer for assistant' to add and process provided user and assistant messages.");
 
         // show the prompt
         Console.ForegroundColor = ConsoleColor.Green;
@@ -46,6 +49,7 @@ public class ChatSessionWithRestart
 
         while (userInput != "exit")
         {
+            // Load the session state from the reset state
             if(userInput == "reset")
             {
                 session.LoadSession(resetState);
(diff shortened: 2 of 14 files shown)


LLama/LLamaContext.cs
```diff
@@ -24,6 +24,7 @@ public sealed class LLamaContext
         : IDisposable
     {
         private readonly ILogger? _logger;
+        private readonly State _emptyState;
```
martindevans, line 27: This empty state doesn't seem to be used anywhere?

LLama/LLamaContext.cs
```diff
@@ -214,6 +216,15 @@ public void LoadState(State state)
             }
         }
 
+        /// <summary>
+        /// Reset the context to the empty state.
+        /// </summary>
+        /// <exception cref="RuntimeError"></exception>
+        public void ResetState()
+        {
+            NativeApi.llama_kv_cache_clear(NativeHandle);
```
martindevans, line 225: You can do `NativeHandle.KvCacheClear()` instead.

LLama/LLamaContext.cs
```diff
@@ -75,6 +76,7 @@ public LLamaContext(LLamaWeights model, IContextParams @params, ILogger? logger
 
             @params.ToLlamaContextParams(out var lparams);
             NativeHandle = SafeLLamaContextHandle.Create(model.NativeHandle, lparams);
+            _emptyState = GetState();
```
martindevans, line 79: Could you check the size of this state? Generally you don't want to be keeping states around in memory if you don't need to, they can be huge. The empty state might be a special case exception to that rule though.

LLama/LLamaExecutorBase.cs
```diff
@@ -250,6 +250,14 @@ protected virtual void TryReuseMathingPrefix()
         /// <returns></returns>
         public abstract ExecutorBaseState GetStateData();
 
+        
+        /// <summary>
+        /// Resets the executor to its initial state.
+        /// Note: Does not affect the context and KV cache.
+        /// </summary>
+        /// <returns></returns>
+        public abstract void ResetState();
```
martindevans, line 259: What's the purpose of resetting an executor to the initial state, rather than just constructing a new one?

LLama/LLamaInteractExecutor.cs
```diff
@@ -93,6 +93,22 @@ public override async Task LoadState(string filename)
...
+            _embed_inps = new List<LLamaToken>();
+            _is_prompt_run = true;
+            _consumedTokensCount = 0;
+            _embeds = new List<LLamaToken>();
+            _last_n_tokens = new FixedSizeQueue<LLamaToken>((int)Context.ContextSize);
+            _n_matching_session_tokens = 0;
+            _pastTokensCount = 0;
+            _pathSession = null;
+            _session_tokens = new List<LLamaToken>();
+            MirostatMu = 0;
```
martindevans, line 109: Mu should be initialised to `null`, not `0`.

LLama/LLamaInstructExecutor.cs
```diff
@@ -110,6 +110,22 @@ public override async Task LoadState(string filename)
...
+            _embed_inps = new List<LLamaToken>();
+            _is_prompt_run = true;
+            _consumedTokensCount = 0;
+            _embeds = new List<LLamaToken>();
+            _last_n_tokens = new FixedSizeQueue<LLamaToken>((int) Context.ContextSize);
+            _n_matching_session_tokens = 0;
+            _pastTokensCount = 0;
+            _pathSession = null;
+            _session_tokens = new List<LLamaToken>();
+            MirostatMu = 0;
```
martindevans, line 126: Mu should be initialised to `null` not `0`.

martindevans:
Thanks for this PR! I'm not very familiar with the higher level executors, but I've just left a few comments on things I spotted :)

As a general note. I'm personally working on bring a brand new executor to LLamaSharp based on the `BatchedExecutor` which I introduced in the least release. It's currently very "low level" compared to these executors, but you may be interested in taking a look at it since it can do a lot of the things you want to do here. I'd be interested in any feedback you have about it.

LLama/LLamaContext.cs
eublefar: Right, forgot to clean it up after I switched to llama_kv_cache_clear, instead of setting the empty state.

LLama/LLamaContext.cs
eublefar: Removed it altogether, as it's not needed

eublefar:
Sorry, I've over-thought it a bit with nullable ExecutorBaseState and context State. I've just made those non-nullable and removed all the `ResetState` methods (as now there is no need for them, there will always be some state to load so no need to reset).

eublefar:
@martindevans 
> As a general note. I'm personally working on bringing a brand new executor to LLamaSharp based on the BatchedExecutor which I introduced in the least release. It's currently very "low level" compared to these executors, but you may be interested in taking a look at it since it can do a lot of the things you want to do here. I'd be interested in any feedback you have about it.

Huh, it indeed does, I should've checked it out. 
As for feedback, my use case is a bit weird for this project I guess, but I shall share it nontheless :)
Basically the use case for me is that I am making a game in Unity with a lot of prompts.
I started this project of mine some time ago and decided to write my own high-level interface based on NativeAPI/handles directly.

Some notes on what would make BatchExecutor viable for my use-case:
- I found that it's impossible for me to keep everything in one KV cache on GPU because of the limited context size, so Save/Load state in-memory is very important. Any ways to `spill` the context into RAM in seamless and optimal way, while still being able to run inference in batch mode would be very appreciated. 
  On a side note: I found out that vanilla Executors actually do something like this while writing this PR and I am very interested in checking out exactly how they do it (I've cleaned KV cache, but if I did not remove Executor state, and then LLM actually continued conversation that should've been removed from the KV cache).
- If I process batches too big - frames per second suffer, so I need to be able to batch requests `implicitly` (e.g. split single prompt/message into multiple runs). If I am not doing it manually, llama.cpp just crashes whole application with assert that batch size is too small, so C# side needs to be smart in this regard.

martindevans (approved):
Thanks for fixing those previous comments, this looks good to me 👍 

I won't merge it myself immediately, since I'm not very familiar with the high level chat/executor stuff. If no one else has any comments I'll merge it before the next release happens :)

LLama/ChatSession.cs
```diff
@@ -494,3 +584,50 @@ in OutputTransform
...
+
+/// <summary>
+/// The state of a chat session in-memory.
+/// </summary>
+public record SessionState
+{
+    /// <summary>
+    /// Saved executor state for the session in JSON format.
+    /// </summary>
+    public string ExecutorState { get; init; }
```
AsakusaRinne, line 596: I don't think we should use type `string` for it here. Though it's convenient to serialize the state and then save as file with a json format executor state, it doesn't seem to make sense in the case of keeping states in memory. It's better to serialize the executor state to string only when saving the `SessionState` to file.

LLama/ChatSession.cs
```diff
@@ -494,3 +584,50 @@ in OutputTransform
...
+
+    /// <summary>
+    /// The input transform pipeline used in this session.
+    /// </summary>
+    public List<ITextTransform> InputTransformPipeline { get; init; } = new();
+
+    /// <summary>
+    /// The output transform used in this session.
+    /// </summary>
+    public ITextStreamTransform OutputTransform { get; init; } = new LLamaTransforms.EmptyTextOutputStreamTransform();
```
AsakusaRinne, line 611: I think we should re-consider the design here. Since the transforms are included here, it's likely telling users that the state of each transform will be well saved. However, in some cases it does not, for example if the user implement a transform with internal state, the state may be lost after re-load it from the saved state file.

Let's gather the concerns here:
1. Should the transforms be a part of the session state?
2. Is the session state expected to be able to save as file?
3. If transforms are included, how to ensure they can be correctly saved and loaded?

I'm open for these points and I'd like to provide my views here.
1. I prefer to include them but should only do so if they could be properly saved and loaded. If you finally decide to remove them, please add some warnings to let users know the potential inconsistent behavior after loading the session.
2. Yes, I think it's a common utilization of the state.
3. Maybe we could add a new interface `ISerializableTransform` with a pair of method `Serialize` and `Deserialize`, moving the responsibility of converting between (json) string and class to user. For the internally defined transforms in LLamaSharp, we could implement this interface for them.

LLama/ChatSession.cs
```diff
@@ -134,6 +158,45 @@ public void SaveSession(string path)
...
+                    state.ExecutorState, statefulExecutor.GetStateData().GetType()
+                ) as ExecutorBaseState ?? throw new ArgumentException("Executor state is invalid", nameof(state))
+            );
+        }
+        else
+        {
+            throw new ArgumentException("Executor must be a StatefulExecutorBase to support loading of session state", nameof(state));
+        }
+        Executor.Context.LoadState(state.ContextState);
+        History = ChatHistory.FromJson(state.History) ?? new();
```
AsakusaRinne, line 197: If the transforms are finally included in the session state, we should also load them here.

LLama/ChatSession.cs
```diff
@@ -494,3 +584,50 @@ in OutputTransform
...
+
+    /// <summary>
+    /// The history transform used in this session.
+    /// </summary>
+    public IHistoryTransform HistoryTransform { get; init; } = new LLamaTransforms.DefaultHistoryTransform();
+    
+    /// <summary>
+    /// The JSON representation of the chat history for this session.
+    /// </summary>
+    public string History { get; init; } = new ChatHistory().ToJson();
```
AsakusaRinne, line 621: Similar to the previous comment, I think we should only convert it to string when we need to save the state as file.

LLama/LLamaExecutorBase.cs
```diff
@@ -315,6 +315,34 @@ public virtual async IAsyncEnumerable<string> InferAsync(string text, IInference
             }
         }
 
+        /// <summary>
+        /// Asynchronously runs a prompt through the model to compute KV cache without generating any new tokens.
+        /// </summary>
+        /// <param name="prompt">Prompt to process</param>
+        /// <param name="cancellationToken">A cancellation token</param>
+        /// <returns></returns>
+        public virtual async Task AddPromptAsync(string prompt, CancellationToken cancellationToken = default)
```
AsakusaRinne, line 324:
```suggestion
        public virtual async Task PrefillPromptAsync(string prompt)
```

Please remove the `canlellationToken` since it's not used and `prefill` matches the functionality of the method better.

LLama/LLamaExecutorBase.cs
```diff
@@ -315,6 +315,34 @@ public virtual async IAsyncEnumerable<string> InferAsync(string text, IInference
             }
         }
 
+        /// <summary>
+        /// Asynchronously runs a prompt through the model to compute KV cache without generating any new tokens.
```
AsakusaRinne, line 319:
```suggestion
        /// Asynchronously runs a prompt through the model to compute KV cache without generating any new tokens.
        /// It could reduce the latent of the first time response if the first input from the user is not immediate.
```

LLama/LLamaExecutorBase.cs
```diff
@@ -315,6 +315,34 @@ public virtual async IAsyncEnumerable<string> InferAsync(string text, IInference
...
+                ReturnValue = false,
+                WaitForInput = true,
+                NeedToSaveSession = false
+            };
+
+            await PreprocessInputs(prompt, args);
+            // First run adds the prompt to the _embeds
+            await InferInternal(inferenceParams, args);
+            // Second run puts it through decode
+            await InferInternal(inferenceParams, args);
```
AsakusaRinne, line 343: I'm not sure if it leads to infinite generation if no correct anti-prompt is set. Could you please confirm that?

AsakusaRinne (changes requested):
Thank you for the contribution! The part of pre-filling prompt looks good to me while I have some concerns about the the part of session state. I'm open for any discussion and please feel free to ping me if there's something blocking you. :)

martindevans:
@eublefar I opened a discussion thread [here](https://github.com/SciSharp/LLamaSharp/discussions/564) about the your BatchedExecutor feedback, to keep this PR on topic :)

LLama/ChatSession.cs
eublefar: The reason I am using string here is that ExecutorState classes contain references to the internal state of the Executor (specifically the List variables I think).  So when Executor state is created and later executor is used again - ExecutorState get's updated unintentionally (by referencing internal state of the Executor).
Maybe it would make more sense to fix executor state creation and deep-copy those Lists/variables though?

LLama/ChatSession.cs
eublefar: I generally was aiming at in-memory only, non-serializable SessionState, as there is already a way to save/load ChatSession into file.
If it's non-serializable, then just storing transform objects suffices I think.

LLama/ChatSession.cs
eublefar: I think the case here is the same as with the ExecutorState (although I am not 100% positive on that)

LLama/LLamaExecutorBase.cs
eublefar: I set `WaitForInput = true` in inference args from the start which supposedly stops it from generating any tokens at all, from what I understand. (It's set to true when antiprompts are encountered in the generation implementation).

LLama/ChatSession.cs
eublefar: Oh, forgot to do that, I'll correct it.

LLama/ChatSession.cs
AsakusaRinne: I prefer to making deep-copy here, because if we keep the data structure of `ExecutorState` here, it's convenient for users to modify the internal state. :) Currently there's no deep-copy method for `ExecutorState` so it may take you some extra time to make it.

LLama/ChatSession.cs
AsakusaRinne: It's okay to keep the current design in this PR. The file saving and loading of `ChatSession` still has some problems. I'll head to it after having this PR merged.

Xsanf:
[эвблефар](https://github.com/eublefar)  > If I process batches too big - frames per second suffer....


I'm not sure if this will help, but it helped me in a similar situation.

By default, we transfer control to await UniTask.NextFrame(); and we believe that this is sufficient from the point of view of control logic.

```
               await foreach (var text in executor.InferAsync(query, inferenceParams))
                {
                    output += text;
                    Output.text = output;
                    await UniTask.NextFrame();
                }
```

At a fixed frequency of 35 FPS
```
        Application.targetFrameRate = 35;
        QualitySettings.vSyncCount = 1;
```
During execution, FPS may drop to 7-6 frames for the entire time of inference.
There arises the difficult question of synchronizing two sequences with different periods, the Moire effect, at what exact moment in the frame the control transfer occurs.

Quite a crude technique, but you can simply add additional await UniTask.NextFrame();

```
                await foreach (var text in executor.InferAsync(query, inferenceParams))
                {
                    output += text;
                    Output.text = output;
                    await UniTask.NextFrame();
                    await UniTask.NextFrame();
                    await UniTask.NextFrame();
                }

```
When adding one additional UniTask.NextFrame(); FPS reaches up to 25-35.
When adding two UniTask.NextFrame(); FPS does not drop.

This has almost no effect on the speed of the withdrawal itself, and eliminates the Moire effect (frieze), since regardless of the generation length, only the generation time of one token matters.

Additional question. Why is the "seed" set in ModelParams. When it is more logical in inferenceParams, where it is needed.
This is not very convenient, because if you need to regenerate the request to obtain a new option, you cannot change the "seed" without additional manipulations.

martindevans:
> Why is the "seed" set in ModelParams.

LLamaSharp has [IModelParams](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IModelParams.cs) which is everything required to load a model into memory and `IContextParams` which is everything required to create an inference context. Seed is set on the **context** params. The `ModelParams` class is a convenience that implements both interfaces in one place, so you can set all config options in one go. But if you want more control you can implement those two params interfaces yourself.

(If you want to ask any followup questions please open an issue and ping me, to keep this PR on topic)

AsakusaRinne:
@eublefar Hi, how's it going? Many thanks for your contribution. We'll be happy if you could complete this PR but it shouldn't be blamed if you don't have time to continue. Please let us know if you're not available in the future two weeks and I'll finish this work myself. :)

eublefar:
@AsakusaRinne Hey, sorry, just got back from the vacation. I'll get on it today :)

eublefar:
@AsakusaRinne I've suffered a bit with serializing/deserializing transforms, but It's working now I think. Also should be backward compatible with previous versions of serialization (although I am not sure about context state)

eublefar:
Tests pass on my machine and I can't figure out which one is creashing here, any suggestions?

martindevans:
Tests are a little flakey at the moment (language models are huge, so I think we're just trying to do too much work in some tests). Since you passed on Windows, and this PR isn't really platform specific, you're probably ok. I've restarted the failed CI runs.

AsakusaRinne (changes requested):
Hi, sorry for the delay of the review. It's impressing and the overall looks good to me, with a few comments left. :)

LLama/ChatSession.cs
```diff
@@ -204,6 +262,33 @@ public ChatSession AddMessage(ChatHistory.Message message)
         return this;
     }
 
+    
+    /// <summary>
+    /// Compute KV cache for the system message and add it to the chat history.
+    /// </summary>
+    /// <param name="content"></param>
+    /// <returns></returns>
+    public async Task<ChatSession> ProcessSystemMessage(string content)
```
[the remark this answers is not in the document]
eublefar: There is InitializeSessionFromHistoryAsync(ILLamaExecutor executor, ChatHistory history)
I've mimicked the AddMessage API you have to provide messages in different roles though.

LLama/ChatSession.cs
```diff
@@ -494,3 +579,185 @@ in OutputTransform
...
+        File.WriteAllText(historyTransformFilepath, JsonSerializer.Serialize(HistoryTransform));
+    }
+
+    /// <summary>
+    /// Load the session state from folder.
+    /// </summary>
+    /// <param name="path"></param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentException">Throws when session state is incorrect</exception>
+    public static SessionState Load(string path)
```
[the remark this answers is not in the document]
eublefar: Added

LLama.Examples/Examples/ChatSessionWithRestart.cs
```diff
@@ -0,0 +1,95 @@
...
+        Console.WriteLine("The chat session has started. Write `save` to save session in memory."
+            + " Write `reset` to start from the last saved checkpoint");
+
+        // show the prompt
+        Console.ForegroundColor = ConsoleColor.Green;
+        string userInput = Console.ReadLine() ?? "";
+
+        while (userInput != "exit")
+        {
+            if(userInput == "reset")
```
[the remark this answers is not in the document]
eublefar: Added

LLama.Examples/Examples/ChatSessionWithHistory.cs
```diff
@@ -61,6 +61,12 @@ public static async Task Run()
                 Console.ForegroundColor = ConsoleColor.Yellow;
                 Console.WriteLine("Session saved.");
             }
+            else if (userInput == "load")
```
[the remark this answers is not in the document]
eublefar: Added

AsakusaRinne (approved):
LGTM, many thanks for this contribution! :)

PR #563, opened 2024-03-03
Llava api

SignalRT:
This is a Draft pull request yet. This is a summary of the issues that I need to solve:

- [x] The changes in load seems to work. It's nearly a hack to do not change all the way to load the llama library. I have concerns with CUDA. 
- [x] To make all the test I updated all the binaries from the **update binaries** job. With these binaries the embedding test fails. I don´t review the reason. I just skip this test to check llava API. The binaries used are previous to https://github.com/ggerganov/llama.cpp/pull/5849
- [x] I introduced a minimum test to check llava API that pass the test in Mac and Windows. It seems to fail on linux, but I'm not sure that this is related with llava. I nee to check this on a linux computer.
- [x] ModelParameters.Thread seems to assign an amount of threads to llama parameters when the parameter is null, but do not preserve the setting when you get the property from the context. For my point of view it would be cleaner to ensure that if we set a default value in the assignment, the same value is the one that we get from the property.


SignalRT:
If someone can review this it will be perfect.

I added the reminded binaries in the binary build (I cannot test this).

AsakusaRinne:
@IntptrMax Maybe you would be interested in this PR, which does similar things with yours.

LLama/Native/LLamaRopeType.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace LLama.Native;
+
+public enum LLamaRopeType
```
AsakusaRinne, line 3: Please add comment for it.

AsakusaRinne (commented):
Great work! The overall looks good to me and it'll be better if you'd like to add an example to `LLama.Examples` for using it. 

How is `llava_shared` library expected to be published, is there already a preferred way?

LLama/Native/LLamaRopeType.cs
martindevans: I think that's my responsibility, it's showing up here because this PR is based on #565

SignalRT:
> Great work! The overall looks good to me and it'll be better if you'd like to add an example to `LLama.Examples` for using it.
> 
> How is `llava_shared` library expected to be published, is there already a preferred way?

At this moment the binaries are in the runtime directory of LLamaSharp. Llava is really not completely integrated in llama.cpp and it doesn´t support GPU. I suppose that the easiest way is to add this libraries to each corresponding backend. But this is my opinion. I would expect changes in llama.cpp project in the future, so may be is better not overthink this now.

This PR depends on @martindevans #565,. Answering you question related with the Examples and executors as commented with Martin I will make a later PR. I just introduce in this PR the binaries, the API and the test of the API.

martindevans:
@SignalRT I've just merged #565. Could you rebase this one onto master? That'll get rid of all my changes from the diff in this PR and make it a lot cleaner to review.

LLama/LLavaWeights.cs
```diff
@@ -0,0 +1,51 @@
+
+using System;
+using LLama.Native;
+
+namespace LLama;
+
+public sealed class LLavaWeights : IDisposable
+{
+    public SafeLlavaModelHandle NativeClipHandle { get; }
```
martindevans, line 9: Rename to `NativeHandle` to be inline with other similar classes

LLama/Native/NativeApi.LLava.cs
```diff
@@ -0,0 +1,103 @@
...
+        public Int32 n_head;
+        public Int32 n_layer;
+        public float eps;
+    };*/
+
+    /// <summary>
+    /// LLaVa Image embeddings 
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]
+    public struct llava_image_embed
```
martindevans, line 30: I think it's best to to move these out so they're not nested inside `NativeApi`. That class is already very overloaded!

LLama/Native/SafeLlavaModelHandle.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace LLama.Native
+{
+    /// <summary>
+    /// A reference to a set of llava model weights
+    /// </summary>
+    public sealed class SafeLlavaModelHandle
+        : SafeLLamaHandleBase
+    {
+
+        internal protected SafeLlavaModelHandle(IntPtr handle)
```
martindevans, line 18: No point making something `protected` inside a `sealed` class

LLama/Native/NativeApi.LLava.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// <summary>
+    /// Frees MULTI MODAL PROJECTIONS model / Clip Model
+    /// </summary>
+    /// <param name="ctx"></param>
+    [DllImport(llavaLibraryName, EntryPoint = "clip_free", CallingConvention = CallingConvention.Cdecl)]
+    public static extern void clip_free(clip_ctx ctx);
+
+  
+    /// <summary>
+    /// Sanity check for clip <-> llava embed size match
```
martindevans, line 54: Lots of XML errors here, I tink it's because you've used `<->` instead of `&lt;-&gt;`

LLama/Native/SafeLlavaModelHandle.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        /// <summary>
+        /// Load a model from the given file path into memory
+        /// </summary>
+        /// <param name="modelPath"></param>
+        /// <param name="lparams"></param>
+        /// <returns></returns>
+        /// <exception cref="RuntimeError"></exception>
+        public static SafeLlavaModelHandle LoadFromFile(string modelPath, int verbosity )
+        {
```
martindevans, line 40: Can you include a copy of [this code](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Native/SafeLlamaModelHandle.cs#L93-L99) at the start of this method. It provides **much** better error messages if an invalid file path is used.

LLama/Native/SafeLlavaModelHandle.cs
```diff
@@ -0,0 +1,92 @@
...
+            var ctxContext =  NativeApi.clip_model_load(modelPath, verbosity );            
+            if (ctxContext == IntPtr.Zero)
+                throw new RuntimeError($"Failed to load LLaVa model {modelPath}.");
+
+            return new SafeLlavaModelHandle(ctxContext);
```
martindevans, line 45: You can modify `clip_model_load` to directly return `SafeLlavaModelHandle`. That way you never have to directly handle the poniter. See [here](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Native/SafeLlamaModelHandle.cs#L119) for example.

LLama/Native/SafeLlavaModelHandle.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        public void LoadImage( string imagePath, int threads )
```
martindevans, line 48: Are there any requirements on image format etc that should be documented here?

LLama/runtimes/deps/avx/libllama.dll
martindevans, line 1: It looks like you've accidentally included some libllama.dll files (avx, avx2, avx512 and noavx), probably leftover from before we changed the name to `llama.dll`. Can you delete those.

LLama/runtimes/deps/avx2/libllama.dll
martindevans, line 1: delete this file (see comment above).

LLama/runtimes/deps/avx512/libllama.dll
martindevans, line 1: delete this file (see comment above).

martindevans:
Various nitpick comments, overall looks like a good foundation though 👍

SignalRT:
@martindevans, reviewed the suggested changes.

LLama/Native/NativeApi.LLava.cs
```diff
@@ -88,7 +62,7 @@ public struct llava_image_embed
     /// <param name="embed"></param>
     /// <returns></returns>
     [DllImport(llavaLibraryName, EntryPoint = "llava_image_embed_free", CallingConvention = CallingConvention.Cdecl)]
-    public static extern llava_image_embed* llava_image_embed_free(llava_image_embed* embed);
+    public static extern LLavaImageEmbed* llava_image_embed_free(LLavaImageEmbed* embed);
```
martindevans, line 65: If a `LLavaImageEmbed` is being allocated in some methods and free in another it should be handled with a `SafeHandle` to absolutely ensure it is disposed properly.

Sorry I didn't spot this in my last review!

martindevans: Unless it's extremely short lived, in which case it can be handled with try/finally everywhere it's used. But a `SafeHandle` is probably easier and safer.

zsogitbe:
If I may have a suggestion to this API and also for the Llama part: 

Save the handle of the DLL when loading the library and provide the possibility to Unload (free) the library completely. This will terminate all processes and free all memory (system or CUDA) definitively regardless of any current and future bugs upstream.

AsakusaRinne:
@zsogitbe Could you please tell how to unload the native library? That's something I once searched for. If it's convincing to unload it successfully, I'd like to work on it in another issue (it's out of range of this PR).

martindevans:
[`NativeLibrary.Free`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary.free?view=net-8.0#system-runtime-interopservices-nativelibrary-free(system-intptr)).

I'm not actually sure if doing that would free up memory though, or if it would just be a huge memory leak. e.g. does unloading the DLL (**without** tearing down the whole process) actually free memory that was allocated but never freed, or does it just leak until the process is ended?

zsogitbe:
No leak is expected if you do it well! The aim is to force free all memory (cuda, sys) allocated by C++. If the C++ process stops all memory is released. We of course need to do it in a clever way by freeing everything we can first.I find the force unloading the library a valuable help if bugs stay in C++ allocation,  what is expected...(Sent from Smartphone)
-------- Oorspronkelijk bericht --------Van: Martin Evans ***@***.***> Datum: 08-03-2024  20:28  (GMT+01:00) Aan: SciSharp/LLamaSharp ***@***.***> Cc: Zoli Somogyi ***@***.***>, Mention ***@***.***> Onderwerp: Re: [SciSharp/LLamaSharp] Llava api (PR #563) 
NativeLibrary.Free.
I'm not actually sure if doing that would free up memory though, or if it would just be a huge memory leak. e.g. does unloading the DLL (without tearing down the whole process) actually free memory that was allocated but never freed, or does it just leak until the process is ended?

martindevans:
Windows CI seems to be broken due to #2 on this list: https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes. I suggest either:
 - Set `overwrite: true` on upload artifacts
 - Revert to v3, upgrade to v4 in a separate PR

IntptrMax:
> @IntptrMax Maybe you would be interested in this PR, which does similar things with yours.

It looks good！The coding style is much more consistent with LLamaSharp.

martindevans (approved):

AsakusaRinne (approved):

SignalRT:
@AsakusaRinne , @martindevans,  I would like to clarify the approach in relation with #594 

My original approach was to integrate the LLavaAPI first and after that make another PR with at least an executor and an example. But right now I don´t know if this is the best approach.

martindevans:
I think that's still a good plan, if we need to changed around how binaries are distributed/loaded that can always be done in other followup PRs.

SignalRT:
OK, I will be working on the next PR

martindevans:
Shall we merge this one now?

PR #575, opened 2024-03-04
GPU Memory Leak test

zsogitbe:
There is a GPU memory leak in LLamaSharp. This test extension shows the amount of memory which is not freed. This will help to trace the problem.


LLama.Unittest/MemoryDisposalTests.cs
```diff
@@ -1,12 +1,39 @@
...
+    public static extern void ggml_backend_cuda_get_device_memory(int device, out ulong free, out ulong total);
+
+    private static ulong GetGPUFreeMemory()
+    {
+        ulong freemem = 0;
+        int dc = ggml_backend_cuda_get_device_count();
+        if (dc > 0)
+        {
+            ulong totalmem = 0;
+            StringBuilder sb = new StringBuilder(1024 * 1024);
```
martindevans, line 25: Is this `StringBuilder` bit necessary? Looks like it might be a leftover from debugging?

LLama/Native/SafeLlamaModelHandle.cs
```diff
@@ -74,6 +74,13 @@ public string Description
         /// <returns></returns>
         public int MetadataCount => llama_model_meta_count(this);
 
+        /// <inheritdoc />
+        public new void Dispose()
```
martindevans: Does including this fix the problem? If not, can you remove it from this PR.

zsogitbe:
Yes, it is needed because there is a problem with GPU memory. Some things are not freed.

For some reason my old PR which we removed is again included (Disposing/Releasing GPU memory), we need to remove it...

martindevans:
CI is failing with:
> System.EntryPointNotFoundException : Unable to find an entry point named 'ggml_backend_cuda_get_device_count' in shared library 'llama'.

This is because the CI isn't running with CUDA, not a huge issue. Can you add `[Fact(Skip = "CI does not use CUDA")]` onto these tests as a workaround for now.

zsogitbe:
Unable to find an entry point named 'ggml_backend_cuda_get_device_count' in shared library 'llama'.

LLama.Unittest/MemoryDisposalTests.cs
martindevans: To be clear, talking about these specific lines:
```csharp
StringBuilder sb = new StringBuilder(1024 * 1024);
ggml_backend_cuda_get_device_description(0, sb, 1024 * 1024);
```
which appear to get a string, and then do nothing with it!

zsogitbe:
> CI is failing with:
> 
> > System.EntryPointNotFoundException : Unable to find an entry point named 'ggml_backend_cuda_get_device_count' in shared library 'llama'.
> 
> This is because the CI isn't running with CUDA, not a huge issue. Can you add `[Fact(Skip = "CI does not use CUDA")]` onto these tests as a workaround for now.

Added!

LLama.Unittest/MemoryDisposalTests.cs
zsogitbe: yes, you are right, I have removed it

martindevans:
From some investigation on the llama.cpp side I'm not sure if this is a good way to debug memory usage.

[This code](https://github.com/ggerganov/llama.cpp/blob/master/ggml-cuda.cu#L8593) appears to be a memory pool. which keeps a pool of `MAX_CUDA_BUFFERS` and doesn't free them (even when free is called).

However `ggml_backend_cuda_get_device_memory` simply asks the device how much memory is free (see here: https://github.com/ggerganov/llama.cpp/blob/master/ggml-cuda.cu#L12470) and doesn't take into account that pool!

Can I suggest opening an issue on llama.cpp, describing the issue and linking to this PR so they can see how you've been debugging the issue? I'll chip in with any extra details they may need.

zsogitbe:
> From some investigation on the llama.cpp side I'm not sure if this is a good way to debug memory usage.
> 
> [This code](https://github.com/ggerganov/llama.cpp/blob/master/ggml-cuda.cu#L8593) appears to be a memory pool. which keeps a pool of `MAX_CUDA_BUFFERS` and doesn't free them (even when free is called).
> 
> However `ggml_backend_cuda_get_device_memory` simply asks the device how much memory is free (see here: https://github.com/ggerganov/llama.cpp/blob/master/ggml-cuda.cu#L12470) and doesn't take into account that pool!
> 
> Can I suggest opening an issue on llama.cpp, describing the issue and linking to this PR so they can see how you've been debugging the issue? I'll chip in with any extra details they may need.

I will do so, but if this would be the case, then the code would not crash the second time with a message that there is not enough GPU memory (the first time it runs well), it would use the memory pool. Unless, it allocates a new memory pool each time it is loaded what would be even worse...

zsogitbe:
https://github.com/ggerganov/llama.cpp/issues/5873

martindevans:
> but if this would be the case, then the code would not crash the second time with a message that there is not enough GPU memory

That's definitely true!

If you have a way to reliably reproduce a crash could you add a test for it? e.g. have a loop loading/disposing a model 5 times.

If we can isolate crashing behaviour in a minimal test that helps a _lot_ debugging it.

zsogitbe:
It is approximately 1GB what always stays in GPU memory. For reproducing the crash we need a model which just fits into GPU memory and then the second time it runs there will be approximately 1GB missing.
The problem is that the model you need depends on how much GPU memory you have... and then it is just a simple inference.

zsogitbe:
The problem is related to model type and model size also. I could not reproduce the crash with the small model in the test suite.
It is possible that some supported model types are not freed properly.

zsogitbe:
image

This is the GPU memory allocation after one of the models (size 10GB) is completely Disposed. There is nearly 1GB  of GPU memory still present.

martindevans:
@zsogitbe it looks like you've found the issue upstream in llama.cpp (issue https://github.com/ggerganov/llama.cpp/issues/5873, PR and https://github.com/ggerganov/llama.cpp/pull/5576).

In that case shall we close this PR and open an issue the track the problem instead?

zsogitbe:
> @zsogitbe it looks like you've found the issue upstream in llama.cpp (issue [ggerganov/llama.cpp#5873](https://github.com/ggerganov/llama.cpp/issues/5873), PR and [ggerganov/llama.cpp#5576](https://github.com/ggerganov/llama.cpp/pull/5576)).
> 
> In that case shall we close this PR and open an issue the track the problem instead?

We should keep this here still open Martin because unfortunately the PR https://github.com/ggerganov/llama.cpp/pull/5576 is not good (causes crash) and even when corrected it frees only 15% of the still occupied GPU memory. There is thus still 85% of the originally not freed GPU memory somewhere which needs to be freed.
I will add a correct PR to the cpp library for cublas to correct pull/5576 there.

martindevans:
Definitely not saying there's not an issue! Just it seems like since the bug is upstream we'd be better off with an issue tracking the problem until someone fixes it, rather than keeping a PR open long term (I try not keep "long term" PRs hanging around, or else it becomes very hard to manage).

zsogitbe:
Yes, I see what you mean. You can of course merge this PR into main branch already. The test will be needed when the fix is ready.

martindevans:
@zsogitbe if you're closing this one don't forget to open an issue tracking the problem, linking out the relevant upstream issues/PRs. It's definitely something we want to keep track of!

zsogitbe:
There are overall issues and PRs about the memory leak, but nobody seems to be interested in to solve the problem. I have tried, I have solved 15% of it in C++, but I do not know the code made in the past well enough to help further. I do not have time to play around in this way.

martindevans:
Yeah unfortunately I'm not expecting it to be solved soon. I've seen your comments over on llama.cpp, not getting a lot of traction :(

That said, it's still good to keep an issue around on our side. It can act as a central place to track progress when it does eventually happen, and if other people discover the same problem we've got a place to point them that has all the relevant information and links.

AsakusaRinne:
On my side I'd like to see such an unit test added to LLamaSharp. It provides a good practice to monitor the native memory usage.

PR #577, opened 2024-03-05
Feature： Add llava support

IntptrMax:
1. This is a simple demo for llava, it can work but still have a long time to work well. 
2. It need llava_shared.dll, you can replace the version for your own PC environment. 
3. It can work on llava 1.5 and llava 1.6, but should set ContextSize at least 3392.
4. It's better to free the resource afer work, but in this demo it's not good enough now.
5. Thanks to Rinne, zsogitbe and SignalRT, the demo cann't work without there help.


AsakusaRinne (changes requested):
Thank you for the contribution! At the same time there's also similar work on LLaVA support #563. Maybe @SignalRT and @martindevans would like to look into this PR. Hope that we could gather all the efforts to make llava be well integrated in LLamaSharp.

For me the overall of this PR looks good but there're many details needs to be resolved. If long time is required to finish this PR, you may consider converting it to `draft PR`.

martindevans:
How does this PR relate to the llava work that @SignalRT has been doing (e.g. #563)?

IntptrMax:
> Thank you for your effort [IntptrMax](https://github.com/IntptrMax)! What [martindevans](https://github.com/martindevans) means is that SignalRT is working on a llava implementation which fits well into the current library. Your solution works ok, but it has a different strategy which differs from the strategy used in LlamaSharp. Your code was however useful for learning how llava works! Maybe it is better to close this PR and keep your codebase in your GitHub as reference.

OK，I will close this pr.

Issue #589, opened 2024-03-11
Nuget references both CPU and CUDA simultaneously


xuzeyu91:
<PackageReference Include="LLamaSharp" Version="0.10.0" />
    <PackageReference Include="LLamaSharp.Backend.Cpu" Version="0.10.0" />
    <PackageReference Include="LLamaSharp.Backend.Cuda12" Version="0.10.0" />
When nuget references both CPU and CUDA simultaneously
```
NativeLibraryConfig
.Instance
.WithCuda(true)
.WithLogs(true);

```
The setting is not effective, GPU inference can only be used when deleting the nuget of the CPU
I hope to use CPU or GPU inference through configuration, rather than letting users reference different nuget packages themselves or load different nuget packages during packaging.
This may be more convenient for users

vvdb-architecture:
Probably a duplicate of https://github.com/SciSharp/LLamaSharp/issues/456

AsakusaRinne:
Close this issue as a duplicate. please move to #456 and I'll dig at this problem there.

Issue #456, opened 2024-01-24
using CUDA when both CPU and Cuda12 back-ends are present.


vvdb-architecture:
I'm using Kernel-memory with LLamaSharp. Despite having a RTX 3080 and the latest CUDA drivers installed, CUDA is not used.

Not sure if this is a bug or I'm missing something, so here's a question instead:

The [LlamaSharp.csproj](https://github.com/microsoft/kernel-memory/blob/main/extensions/LlamaSharp/LlamaSharp/LlamaSharp.csproj) contains

     <PackageReference Include="LLamaSharp.Backend.Cpu"/>
     <PackageReference Include="LLamaSharp.Backend.Cuda12"/>

I found out that if both Cpu and Cuda12 back-ends are referenced, only the CPU is being used even if the CUDA DLL is loaded.
Interestingly, the logs do say that the CUDA back-end is loaded, but no Cuda is used.
~~~
[LLamaSharp Native] [Info] Detected OS Platform: WINDOWS
[LLamaSharp Native] [Info] Detected cuda major version 12.
[LLamaSharp Native] [Info] ./runtimes/win-x64/native/cuda12/libllama.dll is selected and loaded successfully.
llama_model_loader: loaded meta data with 19 key-value pairs and 291 tensors from D:\Source\km\Models\llama-2-7b-chat.Q5_K_M.gguf (version GGUF V2)
...
~~~

If I remove the reference to LLamaSharp.Backend.Cpu, then the CUDA back-end will start to be used. The logs show:

~~~
[LLamaSharp Native] [Info] Detected OS Platform: WINDOWS
[LLamaSharp Native] [Info] Detected cuda major version 12.
[LLamaSharp Native] [Info] ./runtimes/win-x64/native/cuda12/libllama.dll is selected and loaded successfully.
ggml_init_cublas: GGML_CUDA_FORCE_MMQ:   no
ggml_init_cublas: CUDA_USE_TENSOR_CORES: yes
ggml_init_cublas: found 1 CUDA devices:
  Device 0: NVIDIA GeForce RTX 3080, compute capability 8.6
llama_model_loader: loaded meta data with 19 key-value pairs and 291 tensors from D:\Source\km\Models\llama-2-7b-chat.Q5_K_M.gguf (version GGUF V2)
...
~~~

I've reported this to the [kernel memory project](https://github.com/microsoft/kernel-memory/issues/266), but was advised to report  this here.

vvdb-architecture:
[...]

PR #594, opened 2024-03-12
fix: incorrect loading of native library with multiple backends

AsakusaRinne:
## What this PR does

It fixes the incorrect loading of native library when having multiple backends installed. This issue has been mentioned in #456 and #589.

## Description about the issue

The behavior in master branch now is quite wierd.
1. When using LLama.Example project directly, all things go as exepected.
2. When installing both cpu and cuda backend packages and using LLamaSharp in another project, it cannot load correct library **even if the log does show it loads the desired library**.

Let's consider the case2. I had many tries and it's concluded as below.
- When I used `WithCuda` or `WithLibrary` to load the cuda version library, the log shown it had selected that file to load but actually no gpu is used. Besides I could confirm it's not because of the model loading problem.
- There were 5 folders/files in my output path, which are `llama.dll`, `avx`, `avx2`, `avx512`, `cuda11`. When I was trying to load the cuda library, I could delete neither `avx` or `cuda11` folder. That means, these two files were both occupied by a certain process, which is obviously the one I started.
- If I deleted `avx` folder, the case became that `avx2` and `cuda11` were both occupied. Still, cuda library could not be used.
- I confirmed the order of searching was correct and I didn't see a double-loading in logs.

After deleting all the folders except `cuda11`, I found it works. The only thing I could come up with is that `avx`, `avx2`, etc. hit the keywords of dotnet runtime and it does not obey the rules we write in our code.

Finally I changed all the paths and insert one more level directory `llama-sharp` into the paths of native libraries. All things worked as expected again on my PC.

**To be honest I don't think my solution is good enough because I still don't know the reason behind it. Besides, through this debug I found some problems in current native loading code, so I'd like to take this PR as a discussion before releasing next version with LLaVA.** 

@martindevans @SignalRT Hope that this mention doesn't bother you. Your participation will help a lot. :)

---

### NativeLibraryConfig with LLaVA
I think we just need to use the same way with llama.dll to deal with LLaVA here. If I'm not mistaking it, there was already a discussion about whether to put llava libraries in each backend packages. I'm not sure if it still needs discussion. 

### OpenCL support
It seems that OpenCL related logic has not been added to `NativeLibraryConfig`. I'll add it later if it's not ignored intentionally. 

### More flexible loading strategy
I think maybe we should make a better abstraction and allow user to override some APIs to add more flexible control of library loading. Currently, as far as I know, users who want to publish an app with LLamaSharp often maintain the location of native libraries themselves, while their user could have various environments. Finally it always fallbacks to using `WithLibrary`, limiting the usage of our control strategy.

However the time cost of it seems to be a problem. I think it should at least be no hurry.

### DIFF
--- a/LLama/LLamaSharp.Runtime.targets
+++ b/LLama/LLamaSharp.Runtime.targets
@@ -4,68 +4,68 @@
     </PropertyGroup>
     <ItemGroup Condition="'$(IncludeBuiltInRuntimes)' == 'true'">
 
-      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/llama.dll">
+      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/llama-sharp/llama.dll">
         <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-        <Link>runtimes/win-x64/native/noavx/llama.dll</Link>
+        <Link>runtimes/win-x64/native/llama-sharp/noavx/llama.dll</Link>
       </None>
-      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/avx/llama.dll">
+      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/llama-sharp/avx/llama.dll">
         <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-        <Link>runtimes/win-x64/native/avx/llama.dll</Link>
+        <Link>runtimes/win-x64/native/llama-sharp/avx/llama.dll</Link>
       </None>
-      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/avx2/llama.dll">
+      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/llama-sharp/avx2/llama.dll">
         <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-        <Link>runtimes/win-x64/native/avx2/llama.dll</Link>
+        <Link>runtimes/win-x64/native/llama-sharp/avx2/llama.dll</Link>
       </None>
-      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/avx512/llama.dll">
+      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/llama-sharp/avx512/llama.dll">
         <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-        <Link>runtimes/win-x64/native/avx512/llama.dll</Link>
+        <Link>runtimes/win-x64/native/llama-sharp/avx512/llama.dll</Link>
       </None>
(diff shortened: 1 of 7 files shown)


AsakusaRinne:
cc @vvdb-architecture @xuzeyu91 Would you like to help to test on this PR? Here's some steps for reference:

1. clone the code and checkout to this PR.
2. build the `LLamaSharp` project.
3. Go to `LLama/packages` and find the nuget package generated in step 2.
4. In Visual Studio, add the path in step3 as a nuget package folder and select it as the source.
5. (important) **delete C:\Users\xxx\.nuget\packages\llamasharp\0.10.0** to clear the cache.
6. Install your self-compiled LLamaSharp package.
7. a) Follow `.github/prepare_release.sh` to compile backends and install them. or b) Install published backend packages but modify the content of your output path **every time you rebuild your project.**

martindevans:
If you're unable to delete those two folders my best guess would be that we're explicitly loading one library (e.g. CUDA) but then it's _also_ implicitly loading the AVX library when one of the native methods is called.

Maybe we should experiment with using `SetDllImportResolver` ([link](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary.setdllimportresolver)? There's been discussion about doing that before, and it's really the "proper" way to do this I think.

AsakusaRinne:
Yes, there must be two libraries loaded at the same time.

I've tried `SetDllImportResolver` but it seemed to only show the library name (llama.dll) instead of library file path when calling a native api.

martindevans:
> I've tried SetDllImportResolver but it seemed to only show the library symbol name (llama) instead of library file path when calling a native api.

What I meant about using `SetDllImportResolver` is rather than using our current system (which loads the DLL ahead of time, and then just hopes that dotnet uses it) we could instead do all of our loading logic inside `SetDllImportResolver`. It's the proper API for it, and if we do that we'll know exactly what the loading system is doing because we're in control of it!

SignalRT:
> NativeLibraryConfig with LLaVA
> 
> I think we just need to use the same way with llama.dll to deal with LLaVA here. If I'm not mistaking it, there was already a discussion about whether to put llava libraries in each backend packages. I'm not sure if it still needs discussion.

I agree that we need the same way to load both libraries, llama.dll and llava_share.dll. My approach was to use the same directory layout, so I would suggest keeping this path and adapt the directory structure to the one that works better to the project.  

The only thing that we didn´t discuss yet was if llava libraries will be in the same nuget packages (that's my preference, it will be the easiest way to the user).

AsakusaRinne:
> @AsakusaRinne did you still want me to try this change out or should I wait while you all explore SetDllImportResolver as the solution?

It's okay to try it later, no hurry please

martindevans:
I've started experimenting with a `SetDllImportResolver` based approach, just as a prototype for testing. I'll push it up as a separate PR.

AsakusaRinne:
After some test, I believe #603 is a better approach to resolve this problem and it'll be completed in another PR. Close this now, thank you all!

Issue #599, opened 2024-03-13
Examples don't run with CUDA12


EtienneT:
I have CUDA 12, I know it works because I ran custom pytorch model for other projects.

But when I try to run LlamaSharp.Examples with LLamaSharp.Backend.Cpu, it works fine.  But when I try to use it with LLamaSharp.Backend.Cuda12, it crash right away with the following error:

```
System.TypeInitializationException
  HResult=0x80131534
  Message=The type initializer for 'LLama.Native.NativeApi' threw an exception.
  Source=LLamaSharp
  StackTrace:
   at LLama.Native.NativeApi.llama_empty_call() in C:\work\Projects\LLamaSharp\LLama\Native\NativeApi.cs:line 27
   at Program.<<Main>$>d__0.MoveNext() in C:\work\Projects\LLamaSharp\LLama.Examples\Program.cs:line 24

  This exception was originally thrown at this call stack:
    LLama.Native.NativeApi.NativeApi() in NativeApi.Load.cs

Inner Exception 1:
RuntimeError: The native library cannot be correctly loaded. It could be one of the following reasons: 
1. No LLamaSharp backend was installed. Please search LLamaSharp.Backend and install one of them. 
2. You are using a device with only CPU but installed cuda backend. Please install cpu backend instead. 
3. One of the dependency of the native library is missed. Please use `ldd` on linux, `dumpbin` on windows and `otool`to check if all the dependency of the native library is satisfied. Generally you could find the libraries under your output folder.
4. Try to compile llama.cpp yourself to generate a libllama library, then use `LLama.Native.NativeLibraryConfig.WithLibrary` to specify it at the very beginning of your code. For more informations about compilation, please refer to LLamaSharp repo on github.
```

I tried running the project in Debug, also with GPU in the configuration manager, tried running it in .net 8, .net 6, and all combinations but always the same error.  I am running the latest version for the nuget packages, 0.10.0.

zsogitbe:
You have probably added both backends CPU and CUDA. That is the reason for the crash. You need to remove one and thus keep only one backend.

EtienneT:
Pretty sure I was testing it one at a time.  I just to confirm I tested it again this morning, making sure GPU was alone:
![image](https://github.com/SciSharp/LLamaSharp/assets/265924/41b216fb-d7ea-42b2-9935-7a9653037970)

Also to make sure, I deleted the bin and obj folders of the example project before testing again.

Same problem unfortunately.

[...]

PR #602, opened 2024-03-14
build: add llama.cpp as submodule.

AsakusaRinne:
Add llama.cpp as a submodule here for some reason:

1. It's easy for users to find which commit they should head to if they want to compile native library themselves.
2. It makes it more convenient for development based on llama.cpp in LLamaSharp.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "llama.cpp"]
+	path = llama.cpp
+	url = https://github.com/ggerganov/llama.cpp.git


zsogitbe:
This will be great! We can then try to make 1 VS solution which compiles the C++ and C# code together with the chosen configuration and with 1 click.

AsakusaRinne:
@SignalRT The ci of windows has failed after 3 times of retry. Is that the same issue with #599, I guess?

martindevans:
Can this new VS Solution use the make build scripts included in llama.cpp? I don't really know much about cpp in VS so maybe that integrates easily.

Asking because at the moment our build is relatively simple because we just clone the llama.cpp repo and call cmake - no maintenance work is required to keep our build up to date. Don't even need cmake or any other build stuff setup and installed locally, because it's all setup to run in GitHub actions. I'd rather not change either of these properties if possible.

SignalRT:
> @SignalRT The ci of windows has failed after 3 times of retry. Is that the same issue with #599, I guess? If that's confirmed a known issue, I'll merge this PR without verification.

The issue #599 , if understand is related to that I missed llava_shared DLL on the cuda build, but this should not break anything related with the test because no cuda test are running in CI.

The test that fail seems to not be related with llava: `The active test run was aborted. Reason: Test host process crashed : llama_model_loader: loaded meta data with 19 key-value pairs and 291 tensors from Models/llama-2-7b-chat.Q3_K_S.gguf (version GGUF V2)`

martindevans:
The same error was happening in https://github.com/SciSharp/LLamaSharp/pull/603, not sure why.

SignalRT:
> The same error was happening in #603, not sure why.

I running the test on the master to try to understand the reason

AsakusaRinne:
If cuda is excluded, is it possible that the newly added model weight leads to a lack of disk space or memory?

zsogitbe:
It is simply because of the loaded llava_shared.dll is wrong. If you replace it with the correct dll, then it will run OK.
I have tested it with replacing it and the test passes.

AsakusaRinne:
Merge this PR without check now since it has no reason for ci to fail because of the submodule.

SignalRT:
> It is simply because of the loaded llava_shared.dll is wrong. If you replace it with the correct dll, then it will run OK. I have tested it with replacing it and the test passes.

I tried in my fork and CI test pass. Some times at the first try sometimes at second try. As an example:
 https://github.com/SignalRT/LLamaSharp/actions/runs/8306390273 

I'm aware that we didn´t introduce llava_shared in the CUDA directories and this is something that I will try to add on my next PR, but as far as I'm aware there is no CUDA test on CI. If you are aware of some wrong llava version I would like to know  specifically the dll that you are talking about because the DLL theoretically comes from update_binaries pipeline.

AsakusaRinne:
@zsogitbe I've created a new branch [experimental_cpp](https://github.com/SciSharp/LLamaSharp/tree/experimental_cpp) for you. On this branch our review won't be stringent. Good luck for you!

zsogitbe:
Thank you Rinne!

PR #603, opened 2024-03-14
`SetDllImportResolver` based loading

martindevans:
Using `SetDllImportResolver` to load llama DLLs as a prototype.

This is just a prototype to check if using this fixes the issue with some DLLs seemingly being loaded twice.

@AsakusaRinne @zsogitbe I'd appreciate if you would test this (particularly with CUDA) and tell me if it seems to help.

### DIFF
--- a/LLama.Examples/Program.cs
+++ b/LLama.Examples/Program.cs
@@ -16,11 +16,14 @@ AnsiConsole.MarkupLineInterpolated(
 
     """);
 
+// Configure native library to use
 NativeLibraryConfig
    .Instance
    .WithCuda()
-   .WithLogs(LLamaLogLevel.Warning);
+   .WithLogs(LLamaLogLevel.Info);
 
+// Calling this method forces loading to occur now.
 NativeApi.llama_empty_call();
 
 await ExampleRunner.Run();
+
--- a/LLama/Native/NativeApi.Load.cs
+++ b/LLama/Native/NativeApi.Load.cs
@@ -1,11 +1,9 @@
 ﻿using LLama.Exceptions;
-using Microsoft.Extensions.Logging;
 using System;
-using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Text.Json;
+using System.Collections.Generic;
 
 namespace LLama.Native
 {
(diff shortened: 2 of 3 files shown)


zsogitbe:
I do not see much difference when using this code. It just uses TryLoadLibrary. But the dll is registered with path=null what probably means 'legacy behavior' in SetDllImportResolver.
You know that the parameter you used as 'path' is actually not a path, but an enum where to search for the path? 
Maybe set it to `path = DllImportSearchPath.UseDllDirectoryForDependencies;` before exiting the delegate!?

martindevans:
With our normal loading system we lead the DLL and then just hope that the one we loaded is the one that's used. However, since Rinne discovered there were _two_ DLLs loaded that's clearly not working!

The idea here is to replaced the built in loading system with our own. That way it can't possibly load two! It's still using TryLoadLibrary because that does seem to correctly load the correct library, so it's simply using the same logic in a different context.

I'm not sure what you mean about the path stuff, it didn't seem relevant to our use case. Could to explain more?

zsogitbe:
I was just wondering if you left the path parameter deliberately null or forgot it. If you are sure that this is what you want, then it is OK. I cannot see any difference on my system between this code and the former one.

zsogitbe:
About something else: there was recently an important update to llama.cpp. The embeddings generation was wrong and it was corrected 2 weeks ago, just 2 days after our binary update. This is one of the reasons why KernelMemory does not work in LLamaSharp when you have more than one document. The system searches for relevant documents based on the embeddings, but they are wrong... Would it be possible to push an early update of the binaries? I know that this is a big ask because some things need to be changed (I have tried the new C++ code, but it does not work with the current LLamaSharp version).

AsakusaRinne:
@martindevans I've tested it and it works well!

martindevans:
@AsakusaRinne thanks for testing! So just to check with this running can you delete all but one of the DLLs, and CUDA works as expected?

If so I'll do another PR with a proper implementation of this and get that merged ASAP.

AsakusaRinne:
> @AsakusaRinne thanks for testing! So just to check with this running can you delete all but one of the DLLs, and CUDA works as expected?
> 
> If so I'll do another PR with a proper implementation of this and get that merged ASAP.

Yes, it also works well in this case! It's a better way than #594

martindevans:
@AsakusaRinne @SignalRT this has been totally rewritten and is now ready for review. I'd appreciate it if you could both test it.

The main change from the prototype to use `SetDllImportResolver` is still in there of course.

Another change is that I've completely split apart llava and llama loading so they can both be done separately on demand. All of the resolution logic now runs twice, once to select the `llama.dll` to use, and again (if necessary) to select the `llava.dll` to use.

AsakusaRinne (approved):
All things looks good to me and it works as expected on Windows with cuda! I have one concern but it shouldn't block the merging of this PR:
- Since the OpenCL (cutlass) backend has been added, shall we implement the logic of it in this PR together?

martindevans:
I've fixed the MacOS loading.

> Since the OpenCL (cutlass) backend has been added, shall we implement the logic of it in this PR together?

I'd rather get that done in another PR if that's ok? This one is already getting a bit larger than I wanted.

AsakusaRinne:
> I'd rather get that done in another PR if that's ok? This one is already getting a bit larger than I wanted.

Sure, that shouldn't be a problem. :)

SignalRT:
@martindevans it's also working for me.

PR #607, opened 2024-03-17
LLava Test Slowdown

martindevans:
Investigated why LLava embedding tests are slow. It looks like they are simply extremely CPU heavy (even on my powerful PC, taking ~10-20 seconds to finish with CPU load the entire time). I tried shrinking the image down 50% (25% as many pixels) but this wasn't much of an improvement.

For now I have **disabled** the image embedding tests to keep CI time under control (another PR took 25 minutes to complete the tests).

Other things fixed while investigating this:
 - Added doc comments on various llava methods/classes.
 - Fixed `llava_image_embed_free` returning `SafeLlavaImageEmbedHandle` when it should be `void`

### DIFF
--- a/LLama.Unittest/LLavaWeightsTests.cs
+++ b/LLama.Unittest/LLavaWeightsTests.cs
@@ -1,5 +1,4 @@
 ﻿using LLama.Common;
-using LLama.Native;
 
 namespace LLama.Unittest
 {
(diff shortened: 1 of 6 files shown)


SignalRT:
@martindevans, The image embeddings are very slow (and will be very slow). I will review the Issues that you report

SignalRT:
@martindevans, Some of the changes are on code already changed (you can see  #609  with a prototype to clarify the ILlamaExecutor interface) .

martindevans:
Ok, in that case do you want to close this PR and you can add whichever changes you want to your PR?

The only really important one is the return type of `llava_image_embed_free`.

PR #619, opened 2024-03-19
Thread-safe queued batched executor

eublefar:
Some changes to illustrate my proposal [here](https://github.com/SciSharp/LLamaSharp/discussions/564).
Changes:
- Added a way to handle batch overflow in BatchedExecutor (executor real batch size > params batch size).  
  Basically now BatchedExecutor has internal Queue of batches to process and it adds there a new batch in case of overflow.
- Added some thread safety for BatchedExecutor.

Use case for this is to just run BatchedExecutor on a separate worker thread while using Conversation instances as front-end objects.

### DIFF
--- a/LLama/Batched/BatchedExecutor.cs
+++ b/LLama/Batched/BatchedExecutor.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using LLama.Abstractions;
@@ -56,6 +58,7 @@ public sealed class BatchedExecutor
     private readonly AsyncMutex _mutex = new AsyncMutex();
 
     private ConcurrentQueue<LLamaBatch> _batchQueue;
+    private readonly ConcurrentDictionary<(ulong Epoch, int BatchIndex), float[]> _logitCache;
 
     /// <summary>
     /// Epoch is incremented every time Infer is called. Conversations can use this to keep track of
@@ -103,6 +106,7 @@ public sealed class BatchedExecutor
         _batchQueue = new ConcurrentQueue<LLamaBatch>();
         Context = model.CreateContext(contextParams);
         Epoch = 1;
+        _logitCache = new ConcurrentDictionary<(ulong Epoch, int BatchIndex), float[]>();
     }
 
     /// <summary>
@@ -154,6 +158,7 @@ public sealed class BatchedExecutor
                     {
                         Epoch++;
                         _batchQueue.TryDequeue(out _);
+                        AddToLogitCache(batch);
                         batch.Clear();
                     }
                 }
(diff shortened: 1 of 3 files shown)


martindevans:
I like this approach of having a per-executor queue instead of a per-conversation queue is much better. I tried prototyping the per-cnoversation thing last night and it was looking very ugly!

I think there are a few thread safety issues in this implementation, but before getting into the details on that I wanted to ask some more questions about how you're using multiple threads in your setup.

There's a sequence of events you have to follow with the `BatchedExecutor`:
1. Prompt some conversations
2. Call infer
3. Sample the same conversations as prompted in step 1. This must be done before infer is called again.

How do you coordinate this with multiple threads?

eublefar:
> Sample the same conversations as prompted in step 1. This must be done before infer is called again.  

Oh, I've missed that condition and I've tested only with single conversation running at the same time, so probably it would'nt work with multiple conversations.

Right now I am just asynchronously waiting for RequiresInference/RequiresSampling flags in the Conversation wrapper and while single conversation is ok, multiple will cause issues as batch may be cleared before I sample.

eublefar:
So I've added a logit cache that would keep all logits based on epoch and batch index. Admittedly it creates some copying and memory overhead, so maybe it's not the best solution.

eublefar:
Now for some reason Guidance example stopped working :/

martindevans:
I think we're paying quite a high cost for thread safety here, both in terms of additional complexity and also actual runtime cost with the copying of the logits.

Do you mind if I take the queueing idea and implement that in a separate PR, and we discuss thread safety some more to work out the best way to do it?

eublefar:
Sure, that's fine by me.

eublefar:
Maybe it would be okay to just create another executor with this functionality?

martindevans:
It could be done as a threadsafe wrapper, or by adding thread safety into the batch/conversation objects themselves.

My main concern really is just working out how multithreading is expected to be used here - as far as I can see the only part of the whole process that can really benefit from parallelism is sampling (which isn't currently threadsafe, so there's definitely some benefit to be had there).

eublefar:
> My main concern really is just working out how multithreading is expected to be used here - as far as I can see the only part of the whole process that can really benefit from parallelism is sampling (which isn't currently threadsafe, so there's definitely some benefit to be had there).

For unity case it's that:
a) Infer should not block main application thread
b) Hard to synchronize components all over the place that may use Conversation. E.g. two dialogues happening around same time, both requesting for next line to be generated.

Hence worker (Executor) with an inference job queue (coming from Conversations) is a natural answer.
Probably true for web services as well (although I never did those in c#, so no idea).

martindevans:
> Infer should not block main application thread

For this issue, would using something like [this](https://github.com/zsaladin/Asyncoroutine) work? The actual inference work in infer is already shifted off to another thread, so if you call infer and then yield instead of blocking on it that should free up the main thread (I think).

eublefar:
Dayum, I've missed this Task.Run in LLamaContext.DecodeAsync and I thought it was calling Decode directly. That simplifies a lot of things, got to test this.

martindevans:
I'll close this PR now, since I think we've moved in a bit of a different direction with BatchedExecutor.

I did experiment quite a lot with different systems based on this idea, but I couldn't quite find anything that works yet. Allowing over-size batches which are then split up into many smaller batches introduces a huge amount of incidental complexity into the executor. It's still something I'd like to address though (some kind of solution will be required for image embeddings, which are almost guaranteed to overflow the batch).

PR #629, opened 2024-03-28
Update LLamaSharpCpp

zsogitbe:
Simple process to create any type of shared LLama Cpp DLL together with the C# code. Only CMake is needed which creates the VS solution automatically.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,114 @@
+#----------------------------------------------------------------------------------------
+# INFO: 
+# - How to use: change the flags in the 'Set one of these ON and all others OFF' section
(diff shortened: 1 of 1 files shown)


zsogitbe:
Finalized the solution. It still needs extension with all types of setups which are needed. I have only added CUDA+AVX2, AVX2 and CUDA. Easy to extend.
TODOs:
- check all default options that the right ones are ON
- add all needed setups
- test on all platforms

AsakusaRinne:
Please ping me if you want this to be merged. :)

zsogitbe:
Yes, please merge. If anyone has some time, then please try to finalize it with the todo's mentioned above. Thank you Rinne.

PR #630, opened 2024-03-28
KernelMemory update with adding the use of already loaded model

zsogitbe:
When using KernelMemory one may have already loaded a model which can then be used with this extension instead of loading the model again.


zsogitbe:
Please check that there is no conflict or Dispose problems with this solution before merging. Thank you.

martindevans:
Can the other overload (which loads the weights) be made to call this new overload? That way we're not duplicating some of the other stuff.

martindevans:
Lifetime management seems ok to me 👍 

Both the executor and the embedder will create new contexts, which will keep the model in memory until done.

zsogitbe:
> Can the other overload (which loads the weights) be made to call this new overload? That way we're not duplicating some of the other stuff.

Removed code repetition.

martindevans (approved):

Issue #180, opened 2023-09-23
'The type initializer for 'LLama.Native.NativeApi' threw an exception' in MAUI App


nappa0326:
I thought the problem was similar to [#58](https://github.com/SciSharp/LLamaSharp/issues/58), but I could not solve it with that issue's solution.

First, I added the following source to the constructor of the MAUI App's MainPage.xaml.cs created in the wizard.

```
string modelPath = "{model_path}"

// Load model
var parameters = new ModelParams(modelPath)
{
    ContextSize = 2048,
    Seed = 1337,
    GpuLayerCount = 5,
};

using var model = LLamaWeights.LoadFromFile(parameters);
using var context = model.CreateContext(parameters);
var ex = new InteractiveExecutor(context);
ChatSession session = new ChatSession(ex);
```

Then, execute this application, the following exception is occurred.

> System.TypeInitializationException
>   HResult=0x80131534
>   Message=The type initializer for 'LLama.Native.NativeApi' threw an exception.
>   Source=LLamaSharp
> 
> RuntimeError: The native library cannot be found. It could be one of the following reasons: 
> 1. No LLamaSharp backend was installed. Please search LLamaSharp.Backend and install one of them. 
> 2. You are using a device with only CPU but installed cuda backend. Please install cpu backend instead. 
> 3. The backend is not compatible with your system cuda environment. Please check and fix it. If the environment is expected not to be changed, then consider build llama.cpp from source or submit an issue to LLamaSharp.
> 4. One of the dependency of the native library is missed.

The environment is as follows.

**Windows11 [Version 10.0.22621.2283]
maui-windows 7.0.92/7.0.100 VS 17.7.34031.279
LLamaSharp 0.5.1
LLamaSharp.Backend.Cuda12 0.5.1**

![スクリーンショット 2023-09-23 140214](https://github.com/SciSharp/LLamaSharp/assets/4612828/cecf7d18-3a26-4ed2-82ff-5319c53b7ca1)

Note that the above source works fine when implement into a console application created by VS2022 wizard in the same environment. The following is the output of that application.

_ggml_init_cublas: found 1 CUDA devices:
  Device 0: NVIDIA GeForce RTX 3060 Laptop GPU, compute capability 8.6
[...]

PR #631, opened 2024-03-30
Add path to find llama.dll for MAUI

evolcano:
This commit is originally made by [lcarrere](https://github.com/lcarrere) in https://github.com/SciSharp/LLamaSharp/issues/180 .

I have confirmed this modification is OK in my windows 11 laptop, MAUI could load GGUF model and infer successfully.

I make this commit according require of [AsakusaRinne](https://github.com/AsakusaRinne).


martindevans:
This needs to be skipped on non-Windows platforms, or else it'll fail because the imported method does not exist. e.g. check `System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(OSPlatform.Windows);` and skipping the call should probably work.

evolcano:
RuntimeInformation.IsOSPlatform(OSPlatform.Windows) is already exist, the imported method AddDllDirectory will only run in windows platform:

```
        private static int GetCudaMajorVersion()
        {
            string? cudaPath;
            string version = "";
            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
            {
                cudaPath = Environment.GetEnvironmentVariable("CUDA_PATH");
                if (cudaPath is null)
                {
                    return -1;
                }

                //Ensuring cuda bin path is reachable. Especially for MAUI environment.
                string cudaBinPath = Path.Combine(cudaPath, "bin");

                if (Directory.Exists(cudaBinPath))
                {
                    AddDllDirectory(cudaBinPath);
                }

                version = GetCudaVersionFromPath(cudaPath);
            }
            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
            {
                // Try the default first
                cudaPath = "/usr/local/bin/cuda";
                version = GetCudaVersionFromPath(cudaPath);
                if (string.IsNullOrEmpty(version))
                {
                    cudaPath = Environment.GetEnvironmentVariable("LD_LIBRARY_PATH");
                    if (cudaPath is null)
                    {
                        return -1;
                    }
                    foreach (var path in cudaPath.Split(':'))
                    {
                        version = GetCudaVersionFromPath(Path.Combine(path, ".."));
                        if (string.IsNullOrEmpty(version))
                        {
                            break;
                        }
                    }
                }
            }

            if (string.IsNullOrEmpty(version))
                return -1;

            version = version.Split('.')[0];
            if (int.TryParse(version, out var majorVersion))
                return majorVersion;

            return -1;
        }
```

@martindevans, did I missed something?

martindevans:
Ah sorry, I didn't zoom out far enough in the diff to see that! In that case it all looks fine to me.

martindevans (approved):

AsakusaRinne (approved):
LGTM, thank you a lot for this contribution!

PR #633, opened 2024-03-31
fix: the missing of llava_shared library.

AsakusaRinne:


SignalRT:
The osx llava libraries should be included on the cpu package.

SignalRT (approved):
It seems ok to me

PR #649, opened 2024-04-03
Logging interceptor

### LINKED ISSUE #645: Can not intercept all the log messages with llama_log_set
`NativeApi.llama_log_set` allows to intercept log messages from llama.cpp. But even if you set log callback as early as possible in your code, you will loose a few first messages that are produced by a call to the `llama_backend_init` from here:

https://github.com/SciSharp/LLamaSharp/blob/fa73c8f07c220c42ea26e2d75267564b7fae7250/LLama/Native/NativeApi.Load.cs#L38

For example, this messages aren't handled by the provided callback:

```text
ggml_init_cublas: GGML_CUDA_FORCE_MMQ:   no
ggml_init_cublas: CUDA_USE_TENSOR_CORES: yes
ggml_init_cublas: found 1 CUDA devices:
  Device 0: NVIDIA GeForce <...>
```

In order for `NativeApi.llama_log_set` to be executed, static constructor of `NativeApi` must run. This static constructor will load native library, and also call `llama_backend_init`. So there is no way to set a callback with `NativeApi.llama_log_set` before that.

There is also a slightly related problem with logging, but this time with LLamaSharp messages: logs, activated by `NativeLibraryConfig.Instance.WithLogs()` will always be written to the console and currently there is not way to redirect them elsewhere.

One possible solution is to make an overload for `NativeLibraryConfig.WithLogs()` that accepts an optional `LLamaLogCallback`, and pass it to the `llama_log_set` during the loading before any other function of llama.cpp is called. Also the same callback instance may be used to intercept LLamaSharp initialization log messages, or maybe it should be another dedicated instance, passed along with the one for llama.cpp.

martindevans:
Rewritten logging callbacks so it can be configured without touching native methods. This allows it to be used before llama.cpp dll is loaded.

Resolves #645

### DIFF
--- a/LLama/Native/NativeApi.Load.cs
+++ b/LLama/Native/NativeApi.Load.cs
@@ -36,6 +36,15 @@ namespace LLama.Native
 
             // Init llama.cpp backend
             llama_backend_init();
+
+            // Set flag to indicate that this has been done. No native library config can be done after this point.
+            NativeLibraryConfig.LibraryHasLoaded = true;
+
+            // Now that the "loaded" flag is set, configure logging in llama.cpp
+            if (NativeLibraryConfig.Instance.LogCallback != null)
+                NativeLogConfig.llama_log_set(NativeLibraryConfig.Instance.LogCallback);
+            if (NativeLibraryConfig.Instance.LoggerCallback != null)
+                NativeLogConfig.llama_log_set(NativeLibraryConfig.Instance.LoggerCallback);
         }
 
 #if NET5_0_OR_GREATER
--- a/LLama/Native/NativeApi.Load.cs
+++ b/LLama/Native/NativeApi.Load.cs
@@ -87,6 +87,9 @@ namespace LLama.Native
 
         private static void Log(string message, LLamaLogLevel level)
         {
+            if (!message.EndsWith("\n"))
+                message += "\n";
+
             NativeLibraryConfig.Instance.LogCallback?.Invoke(level, message);
         }
 
--- a/LLama/Native/NativeApi.Load.cs
+++ b/LLama/Native/NativeApi.Load.cs
@@ -4,7 +4,6 @@ using System.IO;
 using System.Runtime.InteropServices;
 using System.Text.Json;
 using System.Collections.Generic;
-using Microsoft.Extensions.Logging;
 
 namespace LLama.Native
 {
@@ -41,8 +40,6 @@ namespace LLama.Native
             // Now that the "loaded" flag is set configure logging in llama.cpp
             if (NativeLibraryConfig.Instance.LogCallback != null)
                 NativeLogConfig.llama_log_set(NativeLibraryConfig.Instance.LogCallback);
-            if (NativeLibraryConfig.Instance.LoggerCallback != null)
-                NativeLogConfig.llama_log_set(NativeLibraryConfig.Instance.LoggerCallback);
 
             // Init llama.cpp backend
             llama_backend_init();
@@ -90,57 +87,7 @@ namespace LLama.Native
 
         private static void Log(string message, LLamaLogLevel level)
         {
-            // Log to ILogger
-            if (NativeLibraryConfig.Instance.LoggerCallback != null)
-            {
-                NativeLibraryConfig.Instance.LoggerCallback.Log(level.ToLogLevel(), "{message}", message);
-                return;
-            }
-
-            // Skip other log methods unless enabled
-            if (!enableLogging)
-                return;
-            if ((int)level > (int)logLevel)
-                return;
-
-            // Log to callback
-            if (NativeLibraryConfig.Instance.LogCallback != null)
-            {
-                NativeLibraryConfig.Instance.LogCallback(level, message);
-                return;
-            }
-
-            // Log directly to console
-            var fg = Console.ForegroundColor;
-            var bg = Console.BackgroundColor;
-            try
-            {
-                ConsoleColor color;
-                if (level == LLamaLogLevel.Debug)
-                {
-                    color = ConsoleColor.Cyan;
-                }
-                else if (level == LLamaLogLevel.Info)
-                {
-                    color = ConsoleColor.Green;
-                }
-                else if (level == LLamaLogLevel.Error)
-                {
-                    color = ConsoleColor.Red;
-                }
-                else
-                {
-                    color = ConsoleColor.Yellow;
-                }
-
-                Console.ForegroundColor = color;
-                Console.WriteLine($"[LLamaSharp Native] {level} {message}");
-            }
-            finally
-            {
-                Console.ForegroundColor = fg;
-                Console.BackgroundColor = bg;
-            }
+            NativeLibraryConfig.Instance.LogCallback?.Invoke(level, message);
         }
 
         #region CUDA version
@@ -382,8 +329,6 @@ namespace LLama.Native
         {
 #if NET6_0_OR_GREATER
             var configuration = NativeLibraryConfig.CheckAndGatherDescription(lib);
-            enableLogging = configuration.Logging;
-            logLevel = configuration.LogLevel;
 
             // Set the flag to ensure the NativeLibraryConfig can no longer be modified
             NativeLibraryConfig.LibraryHasLoaded = true;
@@ -475,7 +420,5 @@ namespace LLama.Native
         internal const string libraryName = "llama";
         internal const string llavaLibraryName = "llava_shared";        
         private const string cudaVersionFile = "version.json";
-        private static bool enableLogging = false;
-        private static LLamaLogLevel logLevel = LLamaLogLevel.Info;
     }
 }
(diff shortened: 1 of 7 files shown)


AsakusaRinne:
Will this work also suppress native logs from llava?

martindevans:
Unsure. It's just using the llama.cpp logging callback (`llama_log_set`), so if llava messages are _not_ captured by it I think that would be a bug that would be easy to fix upstream.

LLama/Native/NativeApi.Load.cs
```diff
@@ -36,6 +37,15 @@ static NativeApi()
...
+
+            // Set flag to indicate that this has been done. No native library config can be done after this point.
+            NativeLibraryConfig.LibraryHasLoaded = true;
+
+            // Now that the "loaded" flag is set, configure logging in llama.cpp
+            if (NativeLibraryConfig.Instance.LogCallback != null)
+                NativeLogConfig.llama_log_set(NativeLibraryConfig.Instance.LogCallback);
+            if (NativeLibraryConfig.Instance.LoggerCallback != null)
+                NativeLogConfig.llama_log_set(NativeLibraryConfig.Instance.LoggerCallback);
         }
```
dlyz, line 49: There is still a problem here with logs from `llama_backend_init` call: they won't be intercepted. So i believe `llama_backend_init` call should appear after logging initialization. Also hope that `llama_empty_call` won't log anything.

LLama/Native/NativeLogConfig.cs
```diff
@@ -0,0 +1,67 @@
...
+            llama_log_set((LLamaLogCallback?)null);
+            return;
+        }
+
+        // Bind a function that converts into the correct log level
+        llama_log_set((level, message) =>
+        {
+            logger.Log(level.ToLogLevel(), "(llama.cpp): {message}", message);
+        });
+    }
```
dlyz, line 66: Looks like this is slightly inconsistent with NativeLibraryConfig.WithLogCallback(ILogger) when the method called before the initialization: it saves a callback while WithLogCallback saves a logger and NativeApi.Log has different logic for this two cases.

LLama/Native/NativeApi.Load.cs
martindevans: That's a good point. I think moving `llama_backend_init();` below the rest of the code here should catch that.

LLama/Native/NativeLogConfig.cs
martindevans: You mean because the `ILogger` is used at the top of the `Log` method, but the callback in only used after the filtering? For now I can fix that by moving the callback up to the top as well.

I think this probably needs a better fix though. Perhaps just removing the default console logging altogether, so you _must_ supply some kind of logger if you want log messages.

LLama/Native/NativeLogConfig.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            llama_log_set((LLamaLogCallback?)null);
+            return;
+        }
+
+        // Bind a function that converts into the correct log level
+        llama_log_set((level, message) =>
+        {
+            logger.Log(level.ToLogLevel(), "(llama.cpp): {message}", message);
+        });
```
dlyz, line 65: Not 100% sure, not an expert in P/Invoke, but probably the callback may be collected by the garbage collector (after the native call). I caught such exception when used an instance method as a callback (solved it by saving method as a delegate).
There should be no such problem with static methods/lambdas, but this callback captures logger, so it is not static.

LLama/Native/NativeLogConfig.cs
dlyz: > You mean because the ILogger is used at the top of the Log method, but the callback in only used after the filtering?

Yes, not only the filtering, but also the `(llama.cpp): ` prefix here that will be absent in direct `ILogger` call.

I guess this consistency problem could be solved by "duplicating" llama_log_set(LLamaLogCallback? logCallback) logic, but for ILogger.

Or maybe we should not support `ILogger` here and in `WithLogCallback` at all for now. This make sense for low level messages from llama.cpp, because more sophisticated ways of handling them should be applied. For example, llama.cpp calls a log callback for each `.` symbol in `...........` line that is printed during the loading, and a hundred `.` log messages is not ideal for ILogger.

LLama/Native/NativeApi.Load.cs
martindevans: The only call above logging is now `llama_empty_call`. That shouldn't ever log anything (it's specifically selected as a method that does nothing, just to trigger DLL loading).

LLama/Native/NativeLogConfig.cs
martindevans: I've changed this to redirect `WithLogCallback(ILogger? logger)` to `NativeLogConfig.llama_log_set` which will wrap it in a delegate and call back into `WithLogCallback(NativeLogConfig.LLamaLogCallback? callback)`. That way the formatting is consistent and there's no duplicated logic.

I also removed all of the default printing to console, that allows all the associated things (enable logging, log level) to be removed.

> For example, llama.cpp calls a log callback for each `.` symbol in `...........` line....

I have noticed that, any thoughts on how it could be better handled?

LLama/Native/NativeLogConfig.cs
martindevans: Added a GCHandle rooting the log callback when it's set.

LLama/Native/NativeLogConfig.cs
dlyz: > I also removed all of the default printing to console, that allows all the associated things (enable logging, log level) to be removed.

That's fine by me, but not sure this is friendly to users who use console. Maybe it is better to embrace llama.cpp strategy: write to the console always unless there is a callback provided? Then LLamaSharp behavior will be fully consistent with llama.cpp.

LLama/Native/NativeLogConfig.cs
```diff
@@ -0,0 +1,78 @@
...
+        if (logger == null)
+        {
+            llama_log_set((LLamaLogCallback?)null);
+            return;
+        }
+
+        // Bind a function that converts into the correct log level
+        llama_log_set((level, message) =>
+        {
+            logger.Log(level.ToLogLevel(), "(llama.cpp): {message}", message);
```
dlyz, line 75: Prefix `(llama.cpp): ` isn't necessary since ILogger usually have a category.

martindevans: Removed.

LLama/Native/NativeLogConfig.cs
martindevans: I've added in a system to buffer together messages until a newline is encountered, as you suggested. At the moment I've only done this for the `ILogger` method, not the direct callback. I don't want to add too much "magic" over the base function here.

LLama/Native/NativeApi.Load.cs
```diff
@@ -80,47 +87,7 @@ private static void SetDllImportResolver()
...
-
-                Console.ForegroundColor = color;
-                Console.WriteLine($"{loggingPrefix} {levelPrefix} {message}");
-            }
-            finally
-            {
-                Console.ForegroundColor = fg;
-                Console.BackgroundColor = bg;
-            }
+            NativeLibraryConfig.Instance.LogCallback?.Invoke(level, message);
```
dlyz, line 90: Since log callback now used by both llama.cpp and LLamaSharp, and logger-base callback implementation relies on trailing `\n`, we should probably add trailing `\n` here to the message.

LLama/Native/NativeLogConfig.cs
```diff
@@ -0,0 +1,93 @@
...
+            llama_log_set((LLamaLogCallback?)null);
+            return;
+        }
+
+        var builder = new StringBuilder();
+
+        // Bind a function that combines messages until a newline is encountered, then logs it all as one message
+        llama_log_set((level, message) =>
+        {
+            lock (builder)
```
dlyz, line 78: There might be a threading problem for example when two models are loading at the same time from different threads. Messages from different models may get mixed up in the builder. That's why I use thread-local approach for builders.

LLama/Native/NativeApi.Load.cs
martindevans: Done.

LLama/Native/NativeLogConfig.cs
martindevans: I have added this.

Although wouldn't it be a bug in llama.cpp anyway? Since normally these messages go straight to console multiple threads printing at the same time would get interleaved.

dlyz: Yes, you can easily reproduce a mess in console with two threads.
I think logging it is not a priority for llama.cpp, if it is important for the user, they may use llama_log_set and implement something like what we did.

dlyz:
Thanks @martindevans for your work! I have no further notes.

martindevans:
Thanks very much for your detailed review!

PR #656, opened 2024-04-07
Added Whisper.net x LLamaSharp examples for Speech Chat

Lyrcaxis:
Added whisper.net usage example & usage in combination with LlamaSharp.

Note that while the example is named 'Speech Chat', it actually uses non-instruct prompting. Let me know any thoughts!

Also, I added a note on where to download the models from, even when Whisper.net allows auto-download functionality.
This could be embedded by downloading the 150mb `base` model and copying it to build, but it's up to you to decide.

### Videos:

https://github.com/SciSharp/LLamaSharp/assets/32474602/3f5e9f22-8e88-4821-a0cb-40c7c1912aac

https://github.com/SciSharp/LLamaSharp/assets/32474602/f3e23a14-b641-496d-81a1-c185c52a1a56


LLama.Examples/ExampleRunner.cs
```diff
@@ -29,6 +29,8 @@ public class ExampleRunner
         { "Batched Executor: Fork", BatchedExecutorFork.Run },
         { "Batched Executor: Rewind", BatchedExecutorRewind.Run },
         { "Batched Executor: Guidance", BatchedExecutorGuidance.Run },
+        { "Whisper.net: Speech Recognition", SpeechTranscription.Run },
```
AsakusaRinne, line 32: It seems that this example uses nothing of LLamaSharp, though it's used in the other example as an util. Could you please move this part of code into the other example as a method?

LLama.Examples/Examples/SpeechTranscription.cs
```diff
@@ -0,0 +1,201 @@
...
+                var builder = whisperFactory.CreateBuilder().WithThreads(16).WithPrompt(whisperPrompt).WithSingleSegment().WithLanguage("en");
+                (builder.WithBeamSearchSamplingStrategy() as BeamSearchSamplingStrategyBuilder)!.WithPatience(0.2f).WithBeamSize(5);
+                processor = builder.Build();
+
+                waveIn = new WaveInEvent() { BufferMilliseconds = clipLength, WaveFormat = waveFormat };
+                waveIn.DataAvailable += WaveIn_DataAvailable;
+                waveIn.StartRecording();
+            }
+
+            void WaveIn_DataAvailable(object? sender, WaveInEventArgs e)
```
AsakusaRinne, line 93: Please keep the naming convention the same with others here (WaveInDataAvailable).

LLama.Examples/Examples/SpeechTranscription.cs
```diff
@@ -0,0 +1,201 @@
...
+            int nonIdleTime;        // ..clip-based quant-length (1 = clipLength).
+            public (int minBlanksPerSeperation, int minNonBlanksForValidMessages) detectionSettings = (2, 3);
+            // Default detection settings: A speech of 750ms, followed by pause of 500ms. (2x250ms)
+
+            public HashSet<ISpeechRecognitionServiceUser> ServiceUsers = [];
+
+            public SpeechRecognitionServer(string modelPath)
+            {
+                // Adjust the path based on your GPU's type. On your build you ideally want just the correct runtime build for your project, but here we're having all references, so it's getting confused.
+                var libPath = @$"{Environment.GetFolderPath(Environment.SpecialFolder.UserProfile)}\.nuget\packages\whisper.net.runtime.cublas\1.5.0\build\win-x64\whisper.dll"; // Defaulting to cuBlas.
```
AsakusaRinne, line 80: If the user is using other operation systems such as Linux, is CUDA still available for whisper.net?

AsakusaRinne (changes requested):
The overall looks great except of some minor issues. Thanks a lot for your contribution!

In addition to the comments, could you please print a brief description of this example when the user runs it? For example, `This example shows how to recognize the voice and then get response from LLM. It uses whisper.net and models could be found in....... To use it, please open your microphone and speak to it, then......`

LLama.Examples/Examples/SpeechTranscription.cs
AsakusaRinne: It's okay to ignore the auto library selection here, however could you please print this note in the console rather than only add it to the code comment? When printing it in the console, please use yellow or red color since it's an important information. :)

LLama.Examples/ExampleRunner.cs
AsakusaRinne: If I'm not misunderstanding it, the `ApeechChat` example already supports speech recognition internally. If that's right, it's totally okay to remove this example. But if you would like to, you could add a link of your own repo to tell other users to find the example of pure speech recognition to let others know your work. :)

Lyrcaxis: Great! I think the `SpeechChat` example is enough yup!

AsakusaRinne (approved):

AsakusaRinne:
Merging this PR now, thank you for the contribution!

Mrgaton:
I love you, but where do i donwload the llama models a auto model downloader would be extremely usefoul and nice

Lyrcaxis:
> I love you, but where do i donwload the llama models a auto model downloader would be extremely usefoul and nice

@Mrgaton Thanks :) You mean the quantized language models? Those could be obtained from huggingface.

You can either download a prequantized model (GGUF) or provide a non-quantized version (FP32, FP16, BF16) and convert it to GGUF via the llama.cpp library: https://github.com/ggerganov/llama.cpp/blob/master/convert.py

Here are some GGUF quants for LLaMA3-8B-Instruct: https://huggingface.co/bartowski/Meta-Llama-3-8B-Instruct-GGUF

Mrgaton:
I managed to found a quantized llama3 8b model but is horrible is worse than gpt 3.5

Mrgaton:
Also extremely slow cause I don't know why it's not using my gpu

Lyrcaxis:
@Mrgaton Well, gpt-3.5 is better than any released open source models atm :p You've set the bar too high.
As for the GPU thing, you can join LLamaSharp's discord and figure it out with other people! https://discord.gg/F5wRCNAQ

AsakusaRinne:
> I love you, but where do i donwload the llama models a auto model downloader would be extremely usefoul and nice

The model downloading feature is on the way now. :)

Issue #382, opened 2023-12-24
dotnet publish issue - multiple publish output files with same relative path


JohnGalt1717:
dotnet publish/src/Tools/Chat/Chat.csproj -c Release -o /out -p:SelfContained=true -p:PublishReadyToRun=true

```
C:\Program Files\dotnet\sdk\8.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.ConflictResolution.targets(112,5): error NETSDK1152: Fo
und multiple publish output files with the same relative path: C:\Users\kk\.nuget\packages\llamasharp.backend.cpu\0.8.1\runtimes\linux 
-x64\native\avx\libllama.so, C:\Users\jrlh\.nuget\packages\llamasharp.backend.cpu\0.8.1\runtimes\linux-x64\native\avx2\libllama.so, C:\U 
sers\kk\.nuget\packages\llamasharp.backend.cpu\0.8.1\runtimes\linux-x64\native\avx512\libllama.so, C:\Users\kk\.nuget\packages\llama 
sharp.backend.cpu\0.8.1\runtimes\linux-x64\native\libllama.so.
```

philippjbauer:
Same here for Windows x64 publishing

martindevans:
It appears that the issue here is publish tries to place all of the various binaries (e.g. `avx\libllama.so`, `avx2\libllama.so`, `avx512\libllama.so` etc) into the root folder. Since they all have the same name and are being placed in the same directory this causes a conflict. They should be placed into separate folders, I'm not sure why publish isn't respecting that.

martindevans:
Publishing `LLama.Examples` from the main project _does_ work, so this must be an issue with the way the nuget packages are set up. Unfortunately I don't know much about that at all. Is this something you'd be interested in investigating further @JohnGalt1717 ?

martindevans:
Perhaps related to this? https://github.com/SciSharp/LLamaSharp/blob/master/LLama/runtimes/build/LLamaSharpBackend.props

Looks like it tells it to copy all the files to the same directory.

JohnGalt1717:
> Perhaps related to this? https://github.com/SciSharp/LLamaSharp/blob/master/LLama/runtimes/build/LLamaSharpBackend.props
> 
> Looks like it tells it to copy all the files to the same directory.

I would agree, however I don't have experience with props so I wouldn't feel comfortable changing this.

[...]

PR #658, opened 2024-04-08
Nuget Universal Props File

martindevans:
Moved the .props file to build directory, instead of just netstandard2.0 directory. This should mean it executes for all platforms.

I haven't tested this, I'm not sure how to best test a nuget change. Any assistance there would be appreciated!

### DIFF
--- a/LLama/Native/NativeApi.Load.cs
+++ b/LLama/Native/NativeApi.Load.cs
@@ -254,7 +254,7 @@ namespace LLama.Native
         {
             GetPlatformPathParts(out _, out var os, out var fileExtension, out var libPrefix);
 
-            return $"runtimes/{os}/native/{libPrefix}{libraryName}{fileExtension}";
+            return $"llama/{os}/native/{libPrefix}{libraryName}{fileExtension}";
         }
 
         /// <summary>
@@ -267,7 +267,7 @@ namespace LLama.Native
         {
             GetPlatformPathParts(out _, out var os, out var fileExtension, out var libPrefix);
 
-            return $"runtimes/{os}/native/{cuda}/{libPrefix}{libraryName}{fileExtension}";
+            return $"llama/{os}/native/{cuda}/{libPrefix}{libraryName}{fileExtension}";
         }
 
         /// <summary>
@@ -284,7 +284,7 @@ namespace LLama.Native
             if (!string.IsNullOrEmpty(avxStr))
                 avxStr += "/";
 
-            return $"runtimes/{os}/native/{avxStr}{libPrefix}{libraryName}{fileExtension}";
+            return $"llama/{os}/native/{avxStr}{libPrefix}{libraryName}{fileExtension}";
         }
 
         private static void GetPlatformPathParts(out OSPlatform platform, out string os, out string fileExtension, out string libPrefix)
--- a/LLama/runtimes/build/LLamaSharp.Backend.Cpu.nuspec
+++ b/LLama/runtimes/build/LLamaSharp.Backend.Cpu.nuspec
@@ -2,7 +2,7 @@
 <package >
   <metadata>
     <id>LLamaSharp.Backend.Cpu</id>
-    <version>$version$</version>
+    <version>0.11.2</version>
     <title>LLamaSharp.Backend.Cpu, the backend for LLamaSharp</title>
     <authors>llama.cpp Authors</authors>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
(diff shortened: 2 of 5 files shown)


AsakusaRinne:
On windows I take the following steps to test nuget package.

1. Get the packages with `nuget pack xxx.nuspec -version xxx`. Please rename the version in `LLamaSharp.csproj` to a unique one such as `v0.11.2-test-0408` so that it won't conflict with the nuget cache.
2. Gather the backend packages to a folder.
3. Open Visual Studio nuget package manager and add the local path as package source (on the top right).

I haven't tried this way in Linux or MAC because VS only supports Windows. Maybe dotnet cli has some instructions to use local path.

SignalRT:
I used the same approach with MacOS and Rider.

martindevans:
Looks like this _still_ doesn't fix it :(

The issue only seems to happen when publishing as "self contained".

martindevans:
@AsakusaRinne @SignalRT thanks for the tips. I've been testing with a local folder as suggested and I _think_ this now works. I'm not very confident though, I've never done much with nuget before!

I'd really appreciate it if you guys could build packages and test them yourselves.

The issue I'm trying to fix is:
1. Create a project
2. Add LLamaSharp + LLamaSharp.Backend.whatever
3. Create a new publish profile (to folder, self contained)
4. Publish it. You should get an error about duplicate files (it's trying to placed every llama.dll in the same folder).

This change _should_ fix that last issue.

martindevans:
I guess the tests failing tell me this did not, in fact, fix the problem :(

LLama/runtimes/build/LLamaSharp.Backend.Cpu.nuspec
```diff
@@ -13,38 +13,42 @@
     <releaseNotes></releaseNotes>
     <copyright>Copyright 2023 The llama.cpp Authors. All rights reserved.</copyright>
     <tags>LLamaSharp LLama LLM GPT AI ChatBot SciSharp</tags>
+    <contentFiles>
+      <files include="any\any\llama\**\*.*" buildAction="None" copyToOutput="true" />
```
AsakusaRinne, line 17: May I ask why `any/any/llama` is used here, instead of using `runtime`?

martindevans: When publishing a "self contained" app The runtime folder hierarchy seems to be flattened. That means that all of the files are put into the same folder, which means there are duplicates and therefore the build fails. By putting the files in `content/any/any` there's no special handling and so the folder hierarchy is preserved.

At least... that's the theory!

LLama/runtimes/build/LLamaSharp.Backend.Cpu.nuspec
```diff
@@ -2,7 +2,7 @@
 <package >
   <metadata>
     <id>LLamaSharp.Backend.Cpu</id>
-    <version>$version$</version>
+    <version>0.11.2</version>
```
AsakusaRinne, line 5: Please revert this change, seems that it's modified when you were testing it.

LLama/Native/NativeApi.Load.cs
```diff
@@ -254,7 +254,7 @@ private static string GetMacLibraryPath(string libraryName)
         {
             GetPlatformPathParts(out _, out var os, out var fileExtension, out var libPrefix);
 
-            return $"runtimes/{os}/native/{libPrefix}{libraryName}{fileExtension}";
+            return $"llama/{os}/native/{libPrefix}{libraryName}{fileExtension}";
```
AsakusaRinne, line 257: I'm a bit curious about this. The binaries are copied to `any/any/llama/**/`, while `llama/**/` is used here. Will the runtime find this path without adding a search directory?

AsakusaRinne (commented):
There's no rush for this PR but it will be conflicted with #688 . So just a reminder here, you could complete this PR before #688 and I'll resolve the conflicts. :)

LLama/runtimes/build/LLamaSharp.Backend.Cpu.nuspec
martindevans: Oops, good catch!

LLama/Native/NativeApi.Load.cs
martindevans: As I understand it `any/any` isn't _really_ a folder, instead it just tells the system this content should be included with **any** language on **any** runtime. e.g. these are all valid declarations:
 - `any/net45/`
 - `cs/any`
 - `cs/netstandard2.1`
 - `any/any`

So the idea here is simply to tell the system that whatever language/runtime is using LLamaSharp, we want these files copied to these exact paths.

martindevans:
> So just a reminder here, you could complete this PR before https://github.com/SciSharp/LLamaSharp/pull/688 and I'll resolve the conflicts.

I haven't had any more time to look at it since last time, but it's been stalled because I didn't understand what caused CI to fail last time. It's obviously some kind of path resolution issue, but it's very painful to debug!

I'll close this one for now since it doesn't work at the moment and we can go ahead with your more developed PRs.

This is a really important issue though - I think it's probably _the_ most important bug to fix in LLamaSharp!

AsakusaRinne:
What about opening an issue for it so that we won't forget it?

martindevans:
I think https://github.com/SciSharp/LLamaSharp/issues/382 covers it.

AsakusaRinne:
@martindevans IIRC you have once tried flatten the libraries with different names but it somehow failed. May I ask about the reason of the failure? I couldn't find the related history. :)

I'm looking into this issue and am wondering if it's a possible solution:
1. rename all the libraries but we could choose to keep the current directory structure, like below.
2. After dynamic selection, copy the selected file to a certain folder (`runtimes/**/llama.dll` is available because we have renamed the library file). Note that this change exactly matches the new interfaces in #688, that a library path should be returned from the `INativeLibrary.Prepare`. So it sounds reasonable to do the copying in this method, no matter for us or for users who want to implement it.

```py
-- runtimes
    -- native
        -- win-x64
            -- avx2
               * win-x64-avx2-llama.dll
            -- cuda11
               * win-x64-cuda11-llama.dll
        -- osx-arm64
            ...
        -- linux-x64
            ...

# If we select windows cuda 11 and avx2 backend in proper order on Windows, the file structure will become what's shown below

-- runtimes
    -- native
        -- win-x64
            -- avx2
               * win-x64-avx2-llama.dll
               * llama.dll  # file is copied.
            -- cuda11
               * win-x64-cuda11-llama.dll
               * llama.dll  # file is copied.
        -- osx-arm64
            ...
        -- linux-x64
            ...
```

TBH I suspected path could not be found correctly after publishing if we take the way below. Therefore I'm also wondering the following alternative.

```py
-- llama
    * win-x64-avx2-llama.dll
    * win-x64-cuda11-llama.dll
    * linux-x64-cuda12-llama.so
    ...

# If we select windows cuda 11 and avx2 backend in proper order on Windows, the file structure will become what's shown below. In this way, the file structure will be resumed to what it is now! 

-- llama
    * win-x64-avx2-llama.dll
    * win-x64-cuda11-llama.dll
    * linux-x64-cuda12-llama.so
    ...
-- runtimes
    -- win-x64
        -- native
            -- avx2
               * llama.dll  # file is copied.
            -- cuda11
               * llama.dll  # file is copied.
```

Thus, things partially went back to your previous work. So your answer will really help a lot. :)

martindevans:
> IIRC you have once tried flatten the libraries with different names but it somehow failed. May I ask about the reason of the failure? I couldn't find the related history. :)

The issue was that I could rename the DLLs that we load directly (e.g. llama.dll). However, some DLLs we do _not_ load ourselves (e.g. clblast.dll) and they cannot be renamed, so we could still need to handle them somehow. The folder hierarchy solves it by having the dependencies next to the ones we load directly.

> Copying

Ideally I'd like a solution which doesn't need dynamically writing DLLs to folders. Most software should be installed in a readonly directory, and copying at runtime would be incompatible with that.

If we can work out how to get the folder hierarchy distributed properly it's all fine. I think this PR was 90% of the way there, and actually was just breaking in CI.

AsakusaRinne:
Thank you for the clarification! Glad to hear that it is going to be resolved. I'll head to another issues then

PR #662, opened 2024-04-12
April 2024 Binary Update

martindevans:
Updated binaries, using [this build](https://github.com/SciSharp/LLamaSharp/actions/runs/8654672719/job/23733195669) for llama.cpp commit `f7001ccc5aa359fcf41bba19d1c99c3d25c9bcc7`.

 - Added all new functions.
 - Moved some functions (e.g. `SafeLlamaModelHandle` specific functions) into `SafeLlamaModelHandle.cs`.
 - Exposed tokens on `SafeLlamaModelHandle` and `LLamaWeights` through a `Tokens` property. As new special tokens are added in the future they can be added here.
 - Changed all token properties to return nullable tokens, to handle some models not having some tokens.
 - Fixed `DefaultSamplingPipeline` to handle no newline token in some models.
 - Switched embeddings tests to use an embedding model (`all-MiniLM-L12-v2.Q8_0`). This model is tiny (<100MB) so it should speed up tests slightly.
 - Added higher level methods for saving/loading sequence state

Testing:
 - [x] Windows (CPU)
 - [x] Windows (CUDA 11)
 - [x] Windows (CUDA 12)
 - [x] Windows (OpenCL)
 - [x] Linux (CPU)
 - [ ] Linux (CUDA 11)
 - [x] Linux (CUDA 12)
 - [ ] Linux (OpenCL)
 - [x] MacOS


zsogitbe:
It is not clear which llama.cpp version this is. Please update the llama.cpp submodule.

martindevans:
@zsogitbe I've updated the submodule to `f7001ccc5aa359fcf41bba19d1c99c3d25c9bcc7` which is the version this is based on.

martindevans:
Not sure what's wrong with OSX CI, I _think_ it's just the normal OSX flakiness.

SignalRT:
@martindevans, basic test works on MacOS.

zsogitbe:
Thank you Martin! It is a bit confusing because it is impossible to find a version called f7001ccc5aa359fcf41bba19d1c99c3d25c9bcc7 on llama.cpp. Without your link I cannot find it. Not sure what is happening here but I trust that it is a recent version that we have the last important updates of llama.cpp.

martindevans:
`f7001ccc...` is the commit id, so it's [this](https://github.com/ggerganov/llama.cpp/tree/f7001ccc5aa359fcf41bba19d1c99c3d25c9bcc7).

zsogitbe:
I think that it would be better if you would use an official release from here: https://github.com/ggerganov/llama.cpp/releases.
Your f7001ccc5aa359fcf41bba19d1c99c3d25c9bcc7 cannot be found in these releases, but I understand that your link points to it. I do not feel comfortable with this GitHub behavior.

martindevans:
That's not specific GitHub behaviour, it's just the commit ID you'd use if you wanted to check out llama.cpp at the right version (i.e. `git checkout f7001ccc5aa359fcf41bba19d1c99c3d25c9bcc7`).

Normally almost every commit in llama.cpp is associated with a "release" since the entire process is automated, we got unlucky with this one because their CI failed (looks like they have issues with flakey MacOS CI too) so the final release step got cancelled.

Next time I'll have a look at the releases as well as the commits. I can always pick a slightly older commit, to line up with a valid release, if there isn't one for the latest commit at the time I started the work (normally I just take whatever is the very latest commit).

m0nsky:
Unit tests for CPU AVX2 and CUDA 12 both passed on my Windows 10 x64 system.

Lyrcaxis:
I didn't run the unit tests, but inference examples work fine in Linux (CPU) + Linux (CUDA)
image

AsakusaRinne:
Windows CUDA11 works fine for me.

zsogitbe:
https://github.com/SciSharp/LLamaSharp/commit/c325ac91279428e8d88719357b93e43cd9c5ed52#commitcomment-141108660

Issue #622, opened 2024-03-22
Exception error when sending any prompts.


Webslug:
I have not been able to get LlamaSharp to work since version 0.2.0 - I am about to give up trying.  I'm trying to output to a rich text box and I previously had this working in older versions.

                ```
        
        public async static void ProcInput()
        {
            string userPrompt = Program.MainForm.txtInput.Text; // Get the user's input from the TextBox

            Program.MainForm.txtOutput.AppendText("\nUser: " + userPrompt + "\n");
            MainForm.txtInput.Text = "";

            // Call ChatAsync with the correct arguments
            await foreach (var text in Program._session.ChatAsync(Program.chatHistory, true, new InferenceParams() { Temperature = 0.9f, AntiPrompts = new List<string> { "User:" } }))
            {
                Program.MainForm.txtOutput.AppendText(text);
            }

            if (Program.MainForm.txtOutput.Lines.Length > 0)
            {
                // Remove the last line
                int lastLineIndex = Program.MainForm.txtOutput.Lines.Length - 1;
                Program.MainForm.txtOutput.Text = string.Join(Environment.NewLine, Program.MainForm.txtOutput.Lines.Take(lastLineIndex));
            }
        }

```

```
        public static void InitAI()
        {
            // Load a model
            var parameters = new ModelParams(Program.modelPath)
            {
                ContextSize = 1024,
                Seed = 1337,
                GpuLayerCount = 80
            };

            // Initialize chat history
            Program.chatHistory = new ChatHistory(); // Reset chat history to start fresh

            // Concatenate the user prompt to the existing prompt
            Program.chatHistory.AddMessage(AuthorRole.System, "Transcript of a dialog, where the User interacts with an Assistant named Bob. Bob is helpful, kind, honest, good at writing, and never fails to answer the User's requests immediately and with precision.");
            Program.chatHistory.AddMessage(AuthorRole.User, "Hello, Bob.");
            Program.chatHistory.AddMessage(AuthorRole.Assistant, "Hello. How may I help you today?");

            InferenceParams inferenceParams = new InferenceParams()
            {
                MaxTokens = 256, // No more than 256 tokens should appear in answer. Remove it if antiprompt is enough for control.
                AntiPrompts = new List<string> { "User:" } // Stop generation once antiprompts appear.
            };

            var model = LLamaWeights.LoadFromFile(parameters);
            var context = model.CreateContext(parameters);
            var executor = new InteractiveExecutor(context);
[...]

Issue #606, opened 2024-03-17
Fatal error. System.Runtime.InteropServices.SEHException


K1tiK4:
When trying to run a model on CUDA 12, a fatal error occurs
![image](https://github.com/SciSharp/LLamaSharp/assets/111649965/d37f4fb7-dc00-4cb6-9045-0d0124209d14)

AsakusaRinne:
Hi, which version of LLamaSharp and dotnet runtime are you using?

K1tiK4:
![image](https://github.com/SciSharp/LLamaSharp/assets/111649965/25f9c751-f3e0-42ba-b927-d3428ec1debf)

AsakusaRinne:
Could you please check your model path again to confirm it's the right path?

K1tiK4:
The model path is definitely correct since everything is running on the processor back end.

AsakusaRinne:
OK, then could you please take a try with the master branch code? There've been many commits ahead of v0.10.0 now. If the error still appears, please give us a minimal example to reproduce it and we'll dig on this problem.

K1tiK4:
Okay, I'll try now

K1tiK4:
I tried updating net core to 8.0, but nothing changed. The code was taken from the quick start example. Here are the specs of the PC it runs on. Maybe this will help find the problem
Processor: AMD FX 8320
GPU: GTX 1660 super
RAM: 16 GB
OS: Windows 10 pro
![image](https://github.com/SciSharp/LLamaSharp/assets/111649965/6db3a709-2aa8-4108-8c81-c1f61c268f7c)

K1tiK4:
The same error in the example from master
![image](https://github.com/SciSharp/LLamaSharp/assets/111649965/0ac34aed-cc0c-4050-bca9-06c32b99e0b9)

AsakusaRinne:
Hi, I think I've found the problem. Your cpu is a bit old (released in 2012) so I think it has no avx support, while our cuda backend is compiled with avx2 by default.

Firstly, you could confirm this point by installing cpu backend and add this line to the beginning of your code. (with v0.10, not master)

```cs
using LLama.Native;
NativeLibraryConfig.WithLogs();
```

Then, you'll see which level avx of the native library is selected by default.

If it's confirmed that your PC has no avx support or does not support avx, please follow the steps to fix it:
1. clone llama.cpp repo and enter its directory.
2. `mkdir build && cd build`
[...]

PR #663, opened 2024-04-12
Removed `ContextSize` from most examples

martindevans:
Removed `ContextSize` from most examples.

If it's not set it's retrieved from the model, which is usually what you want. Leaving it in the examples is just unnecessary complexity.

### DIFF
--- a/LLama.Examples/Examples/ChatChineseGB2312.cs
+++ b/LLama.Examples/Examples/ChatChineseGB2312.cs
@@ -27,7 +27,6 @@ public class ChatChineseGB2312
 
         var parameters = new ModelParams(modelPath)
         {
-            ContextSize = 1024,
             Seed = 1337,
             GpuLayerCount = 5,
             Encoding = Encoding.UTF8
--- a/LLama.Examples/Examples/ChatSessionStripRoleName.cs
+++ b/LLama.Examples/Examples/ChatSessionStripRoleName.cs
@@ -12,7 +12,6 @@ public class ChatSessionStripRoleName
 
         var parameters = new ModelParams(modelPath)
         {
-            ContextSize = 1024,
             Seed = 1337,
             GpuLayerCount = 5
         };
--- a/LLama.Examples/Examples/ChatSessionWithHistory.cs
+++ b/LLama.Examples/Examples/ChatSessionWithHistory.cs
@@ -10,7 +10,6 @@ public class ChatSessionWithHistory
 
         var parameters = new ModelParams(modelPath)
         {
-            ContextSize = 1024,
             Seed = 1337,
             GpuLayerCount = 5
         };
--- a/LLama.Examples/Examples/ChatSessionWithRestart.cs
+++ b/LLama.Examples/Examples/ChatSessionWithRestart.cs
@@ -10,7 +10,6 @@ public class ChatSessionWithRestart
 
         var parameters = new ModelParams(modelPath)
         {
-            ContextSize = 1024,
             Seed = 1337,
             GpuLayerCount = 5
         };
--- a/LLama.Examples/Examples/ChatSessionWithRoleName.cs
+++ b/LLama.Examples/Examples/ChatSessionWithRoleName.cs
@@ -10,7 +10,6 @@ public class ChatSessionWithRoleName
 
         var parameters = new ModelParams(modelPath)
         {
-            ContextSize = 1024,
             Seed = 1337,
             GpuLayerCount = 5
         };
--- a/LLama.Examples/Examples/GrammarJsonResponse.cs
+++ b/LLama.Examples/Examples/GrammarJsonResponse.cs
@@ -14,7 +14,6 @@ namespace LLama.Examples.Examples
 
             var parameters = new ModelParams(modelPath)
             {
-                ContextSize = 1024,
                 Seed = 1337,
                 GpuLayerCount = 5
             };
--- a/LLama.Examples/Examples/InstructModeExecute.cs
+++ b/LLama.Examples/Examples/InstructModeExecute.cs
@@ -13,7 +13,6 @@ namespace LLama.Examples.Examples
 
             var parameters = new ModelParams(modelPath)
             {
-                ContextSize = 1024,
                 Seed = 1337,
                 GpuLayerCount = 5
             };
--- a/LLama.Examples/Examples/InteractiveModeExecute.cs
+++ b/LLama.Examples/Examples/InteractiveModeExecute.cs
@@ -13,7 +13,6 @@ namespace LLama.Examples.Examples
 
             var parameters = new ModelParams(modelPath)
             {
-                ContextSize = 1024,
                 Seed = 1337,
                 GpuLayerCount = 5
             };
--- a/LLama.Examples/Examples/LoadAndSaveSession.cs
+++ b/LLama.Examples/Examples/LoadAndSaveSession.cs
@@ -12,7 +12,6 @@ namespace LLama.Examples.Examples
 
             var parameters = new ModelParams(modelPath)
             {
-                ContextSize = 1024,
                 Seed = 1337,
                 GpuLayerCount = 5
             };
--- a/LLama.Examples/Examples/LoadAndSaveState.cs
+++ b/LLama.Examples/Examples/LoadAndSaveState.cs
@@ -13,7 +13,6 @@ namespace LLama.Examples.Examples
 
             var parameters = new ModelParams(modelPath)
             {
-                ContextSize = 1024,
                 Seed = 1337,
                 GpuLayerCount = 5
             };
--- a/LLama.Examples/Examples/SpeechChat.cs
+++ b/LLama.Examples/Examples/SpeechChat.cs
@@ -50,7 +50,7 @@ NOTE: You may need to poke around with the voice detection threshold, based on y
 
             public LlamaSession_SpeechListener(SpeechRecognitionServer server)
             {
-                var parameters = new ModelParams(UserSettings.GetModelPath()) { ContextSize = 1024, Seed = 1337, GpuLayerCount = 99 };
+                var parameters = new ModelParams(UserSettings.GetModelPath()) { Seed = 1337, GpuLayerCount = 99 };
                 model = LLamaWeights.LoadFromFile(parameters);
                 context = model.CreateContext(parameters);
                 executor = new InteractiveExecutor(context);
--- a/LLama.Examples/Examples/StatelessModeExecute.cs
+++ b/LLama.Examples/Examples/StatelessModeExecute.cs
@@ -11,7 +11,6 @@ namespace LLama.Examples.Examples
 
             var parameters = new ModelParams(modelPath)
             {
-                ContextSize = 1024,
                 Seed = 1337,
                 GpuLayerCount = 5
             };
--- a/LLama.Examples/Examples/TalkToYourself.cs
+++ b/LLama.Examples/Examples/TalkToYourself.cs
@@ -21,7 +21,7 @@ namespace LLama.Examples.Examples
             var bob = new InteractiveExecutor(bobCtx);
 
             // Initial alice prompt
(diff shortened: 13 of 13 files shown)


zsogitbe:
I think that the context size is an important parameter for GPU memory usage. If it is too high, then you may not be able to upload all layers to the GPU or you may not be able to optimize for speed. You can also have some models with huge context size nowadays which may cause trouble. Maybe better to make some tests and see what happens with memory usage based on different context size.

martindevans:
To be clear I'm not removing the ability to set the context size if you want to!

This is just removing the redundant setting of the context size in all the examples. Since you can load any model you like into the examples that context size was almost always wrong and an extra bit of unnecessary complexity in sample code which should be as simple as possible.

SignalRT:
I think this is a good idea.

Lyrcaxis:
You can remove it from `SpeechChat.cs` as well:
image

zsogitbe:
As I have expected when running an example with a recent model the example crashes with:
```
ggml_backend_cuda_buffer_type_alloc_buffer: allocating 49152.00 MiB on device 0: cudaMalloc failed: out of memory
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
```
This is a very small 3GB model, but the huge context size the model can handle crashes the model.

It is not a good idea to remove the context size from the examples because when people are trying out the library they will just leave because they will think that it is garbage. The context size should be set to the value the example needs and a comment may be placed to explain it.

martindevans:
50GB of context space! What's the model?

I think you're right though, we'll probably need to add context size back in but with a comment on it saying that it's not necessary and that leaving it unset will use the default value from the model.

We could also do with handling that error better, but I don't think that's possible without changes to llama.cpp (making the APIs fallible).

zsogitbe:
Phi-3 128K context size (but small model), recently came out. This will be the trend, increasing context size...

martindevans:
I had a feeling it might be Phi-3 😆 

> Saying that it is not necessary to set is not a good idea either

The reason I removed this originally was because in most cases you don't want to set it, and yet almost every snippet of code anyone ever pastes does include it. Unfortunately that's the power of example code, whatever you put there will be what everyone uses even if it's not quite right! In general you should only be setting ContextSize when you have a specific reason to (e.g. default model context is huge, as in your case).

e.g. a few recent examples:
 - [this question on Discord](https://discord.com/channels/1106946823282761851/1106947264938790972/1232992867299889214)
 - https://github.com/SciSharp/LLamaSharp/issues/686
 - https://github.com/SciSharp/LLamaSharp/issues/622
 - https://github.com/SciSharp/LLamaSharp/issues/606#issuecomment-2002558470

AsakusaRinne:
Models with long context-length will be more and more popular, but I don't think there're many users using  it with only CPU. I can't imagine how slow it will be! The key point is still the GPU memory instead of CPU memory. To solve the problems from the root, there are two ideal approaches.
- Automatically set the `ContextSize` and `GpuLayerCount` based on the information of model and machine.
- Catch the error when the memory is not enough.

Unfortunately, there're problems for both of them.

- Though we could calculate the approximation size of the memory needed the model will use based on model info, `ContextSize` and `GpuLayerCount`, it's hard to get the free GPU memory at multiple platforms. AFAIK there's no such .NET library for it.
- llama.cpp just doesn't support it, though their core team had already be aware of it. https://github.com/ggerganov/llama.cpp/issues/4315

The setting is actually important for both examples and applying to production environment. For examples, I think it's okay to take the first approach above. We could let user input a number which indicates the GPU memory in GiB. Then we could try to automatically set the `ContextSize` and `GpuLayerCount`, using a conservative strategy to avoid OOM. Though users are still required to input a value themselves in this way, it's much more easy for users to understand. For production application, however, I haven't had any good idea about it yet. :(

martindevans:
Yeah it's a bit of a usability pain-point for llama.cpp that it doesn't have some way to determine the best split of memory and context size etc. I've seen a lot of discussions over on the main repo about automatically trying to do it, and every time they decide it's too complex to tackle.

> Instead, we should return an error when an allocation fails. Let's fix that first, and then we can think about a memory estimation API.

I really hope this happens! llama.cpp has way too many APIs that just crash the entire application with an assertion failure. At least if you could detect an allocation failure you could just have a loop shrinking context size until it fits!

---

Back to the issue at hand - how about adding context size back into all of the examples something like this:

```csharp
var parameters = new ModelParams(Program.modelPath)
{
    // Set the maximum amount of tokens the model can hold in memory. If this is set too high it will cause
    // llama.cpp to crash with an out-of-memory error. If you do not set this parameter the default context
    // size of the model will be used.
    ContextSize = 1024,

    // Set how many layers are moved to GPU memory. Setting this parameter too high will cause llama.cpp
    // to crash with an out-of-memory error.
    GpuLayerCount = 80
};
```

That way these are no longer just magical numbers hardcoded into the example, but they teach the user what they mean and how to use them to solve problems (too few tokens, too much memory).

AsakusaRinne:
> I've seen a lot of discussions over on the main repo about automatically trying to do it, and every time they decide it's too complex to tackle.

If we don't pursue maximizing the GPU memory utilization but somewhat lower number such as 80%, things will be easier. However, unless we find a way to get the free GPU memory size, it could be only a toy in our examples. 😞 

>  adding context size back into all of the examples something like this

I agree. We should also tell the users what will happen if their conversation with LLM exceeds the context length.

```cs
var parameters = new ModelParams(Program.modelPath)
{
    // Set the maximum amount of tokens the model can hold in memory. If this is set too high it will cause
    // llama.cpp to crash with an out-of-memory error. If your conversation with the LLM exceeds the context length, 
    // some early memory in the model will be dropped. If you do not set this parameter the default context
    // size of the model will be used.
    ContextSize = 1024,

    // Set how many layers are moved to GPU memory. Setting this parameter too high will cause llama.cpp
    // to crash with an out-of-memory error.
    GpuLayerCount = 80
};
```

zsogitbe:
1. First of all Context size must be set for all applications based on the needs of the application! Some applications need only 100 tokens (e.g., instruction based short answer), but some other need 8000 tokens (e.g., chat), etc. This, in order to optimize memory use and in order to optimize speed. The context size must thus be set by the user case by case.

2. GPU will always have priority because it is just several times faster than CPU. There are use cases for CPU only application, for example, on small devices (smartphone).

3. It is possible to have an automatic load balancing of the (GPU) memory use. I am doing it. llama.cpp could do it, but we can also do it. You need to take into account several variables, free GPU memory at the moment you need to do inference, the available GPU's with their available memory (total memory of all devices), the speed of the GPU's (for optimizing the loading - first use the best GPU), etc... I am doing already all of this, thus it is possible.

4. Estimating the needed GPU memory per model is possible (we just need to overestimate it). The size of the model is already a possible estimate with an extra buffer added, but this can still be more precise. Since we have the option to distribute the layers to GPU's and CPU's it does not matter if the estimate is not exact. Knowing the free GPU memory at all times is possible on Windows, I do not know about other platforms, but with the cuda toolkit it should be possible everywhere.

5. Until we do not have automatic load balancing in the library (as an option) we need to explain the users that setting the context size and all GPU parameters is very important (for the above reasons). We should also add to the example all other parameters needed for this and not only GpuLayerCount (split mode...)! An example must be comprehensive to help new users to understand how to use the library.

AsakusaRinne:
> GPU will always have priority because it is just several times faster than CPU. There are use cases for CPU only application, for example, on small devices (smartphone).

Yes, but for long context (> 16K) I'm afraid few users will use it with only CPU. Smartphone has been much more powerful on  computation than 10 years ago. With OpenCL and Vulkan it could benefit from GPU computation, so smartphone != cpu-only.

> It is possible to have an automatic load balancing of the (GPU) memory use. I am doing it.

> Estimating the needed GPU memory per model is possible (we just need to overestimate it). 

It will be great if we could have these features! I'm looking forward to your work!

zsogitbe:
What I meant Rinne is that I am doing load balancing of the GPU in my project. I will not start a new thread/work about that in LLamaSharp because you and Martin are working on the scheduling and batching already which at the end when merged should include GPU load balancing also. Since GPU load balancing is relatively simple compared to scheduling and batching I am sure that you will be able to include it. If you need any info, then please ask.
Short summary of GPU load balancing:
- check the free GPU memory (all GPU's)
- check total available GPU memory (all GPU's)
- estimate the needed GPU memory for the model inference (model+context). Estimation can be simply done by using the model size on disk + an estimate for context use (just test a model with a few context sizes and you will see how much, for example, each 1K context size uses) + small buffer.
- decide if GPU can be used
- decide if full offload is possible without splitting layers between GPU's and CPU
- decide how many layers can be offloaded to all available GPU's (this can be simply calculated from the above information)
- set LLamaSharp GPU parameters according to the above

AsakusaRinne:
@zsogitbe One of the biggest problems for me is the lack of a library to do the following things you mentioned.

- check the free GPU memory (all GPU's)
- check total available GPU memory (all GPU's)

Is there any cross-platform library to achieve them?

martindevans:
I don't think it's something that would be covered by scheduling batched work - instead I think what you're describing would happen first (model loading). Batching works within the constraints of a pre-established KV cache laid down by the loading process. That means you can work on it indepenently of any other work going on.

I think you'd want to be automatically determining values for these parameters in the `IModelParams`:
 - [`MainGpu`](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IModelParams.cs#L35)
 - [`SplitMode`](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IModelParams.cs#L40)
 - [`GpuLayerCount`](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IModelParams.cs#L45)
 - [`TensorSplits`](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IModelParams.cs#L65)

And _perhaps_ also these context params:
 - [`ContextSize`](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IContextParams.cs#L14)
 - [`TypeK`](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IContextParams.cs#L99)
 - [`TypeV`](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IContextParams.cs#L104)
 - [`NoKqvOffload`](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IContextParams.cs#L109)

If you do manage to come up with a way to do this it would be great. The llama.cpp project have consistently written it off as too complex a problem to solve, but it would be a huge usability enhancement if it could be done!

Personally I'd approach this by not even automating the process yet. Just design an algorithm to decide on those 4 model values and plug the numbers in by hand to see how well it works. That's the quickest way to validate the idea.

zsogitbe:
> @zsogitbe One of the biggest problems for me is the lack of a library to do the following things you mentioned.
> 
> * check the free GPU memory (all GPU's)
> * check total available GPU memory (all GPU's)
> 
> Is there any cross-platform library to achieve them?

@AsakusaRinne, I am sorry, but I do not work multiplatform. My solution is Windows only, and I told Martin the way how to do this on Windows.
If you want to do it multiplatform, then you will need to have a different solution. I advice you to look at the https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator project. You will need the C++ library and then create a C# bridge to get the info. Since you are probably also a game developer as Martin I guess that you know Vulkan. I don't know Vulkan.

>I think you'd want to be automatically determining values for these parameters in the IModelParams:
>
>[MainGpu](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IModelParams.cs#L35)
>[SplitMode](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IModelParams.cs#L40)
>[GpuLayerCount](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IModelParams.cs#L45)
>[TensorSplits](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Abstractions/IModelParams.cs#L65)

@martindevans, yes this is exactly how I do it on Windows and how I tried to explain it above!

PR #665, opened 2024-04-13
Fixed Beam Search Test

martindevans:
- Fixed beam search test to use decode
 - Enabled it in CI


martindevans:
Should we disable the osx CI for now? It's causing a lot of noise, failing 2 or 3 times for every PR before finally working.

SignalRT:
@martindevans I think it will be the best option until GitHub solve the infrastructure issue.

martindevans:
I'll make that change in another PR

PR #669, opened 2024-04-15
Introduce ChatHistory interface

StevenCyb:
This PR changed the way `ChatHistory` is used by abstracting it with an interface `IChatHistory`. 
This PR has **breaking changes**.

Examples teste with the changes.
Unit tests:
```bash
Passed!  - Failed:     0, Passed:    85, Skipped:     3, Total:    88, Duration: 8 s - LLama.Unittest.dll (net8.0)
```

CHANGES:
* Chat `Message` is not an inner class of `ChatHistory` anymore
* `IChatHistory` is now introduces to allow own chat history implementations
* Chat history serialization is now handled by `ChatHistorySerializer`

### DIFF
--- a/LLama.Examples/Examples/ChatChineseGB2312.cs
+++ b/LLama.Examples/Examples/ChatChineseGB2312.cs
@@ -105,7 +105,7 @@ public class ChatChineseGB2312
                 await foreach (
                     var text
                     in session.ChatAsync(
-                        new ChatHistory.Message(AuthorRole.User, userInput),
+                        new Message(AuthorRole.User, userInput),
                         inferenceParams))
                 {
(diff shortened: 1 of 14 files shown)


AsakusaRinne:
Hi, the overall of this PR looks good. Could you please tell about why you want to make this change?

AsakusaRinne:
Hi Steven, thank you very much for this PR! However I prefer to blocking its merging for now because it introduces break changes, while I think we'll soon add some changes to the high-level APIs. Thus your modifications are very likely to be changed again in the near future, making users a bit confused. You don't need to close this PR -- that will remind me about your requested APIs and take your proposal into account when I work on these APIs in the future. Your proposal this PR is reasonable to me. :)

PR #672, opened 2024-04-17
Reintroduce MacOS on CI

SignalRT:
- [x] Reintroduce MacOS on CI. This time running on ARM
- [x] Create a new Fact attribute to ignore test on Release (In this way we can execute the test locally on debug without changing the code).  
- [x] The code disable metal on MacOS (GPU Layers) but should be done only if running on MacOS. Pending to solve.


.github/workflows/main.yml
```diff
@@ -15,19 +15,23 @@ jobs:
...
           - build: windows-release
             os: windows-2019
             config: release
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-dotnet@v4
       with:
-        dotnet-version: | 
+        dotnet-version: |
+          6.0.x
```
martindevans, line 34: Do we want to re-introduce 6.0? Not against it, since it's still supported until november of this year, just checking.

.github/workflows/main.yml
```diff
@@ -15,19 +15,23 @@ jobs:
...
           7.0.x
```
martindevans, line 35: Should we drop 7.0? The official end-of-support is in about a month.

LLama.Unittest/IgnoreOnCIFact.cs
```diff
@@ -0,0 +1,22 @@
+namespace LLama.Unittest;
+
+/// <summary>
```
martindevans, line 3: We could probably add a symbol into the build in CI, so you can detect it directly?

.github/workflows/main.yml
SignalRT: This was a test before I see the debug logs. This could be removed.

.github/workflows/main.yml
SignalRT: My opinion would be to use only NET8. So I will make this.

AsakusaRinne: BTW, should we drop 7.0 in our main package if the official support ends in the future?

LLama.Unittest/Constants.cs
```diff
@@ -8,5 +10,25 @@ internal static class Constants
         public static readonly string LLavaModelPath = "Models/llava-v1.6-mistral-7b.Q3_K_XS.gguf";
         public static readonly string LLavaMmpPath = "Models/mmproj-model-f16.gguf";
         public static readonly string LLavaImage = "Models/extreme-ironing-taxi-610x427.jpg";
+
+        /// <summary>
+        /// Calculate GpuLayer Count to use in UnitTest
+        /// </summary>
+        /// <returns> Defaults to 20 in all the cases, except IOS release (to disable METAL on github CI)</returns>
```
martindevans, line 17: > IOS release

did you mean MacOS here?

SignalRT: Yes, sorry...

PR #674, opened 2024-04-19
Embeddings correction

zsogitbe:


zsogitbe:
This supports all types of models. Until we do not need more than one sequence or pooling we should use this code.

CMakeLists.txt
```diff
@@ -0,0 +1,126 @@
+#----------------------------------------------------------------------------------------
```
martindevans, line 1: There's a lot of stuff in this PR not relevant to embeddings. Please remove the `LLama.GenerateSolution` folder.

LLama/LLamaEmbedder.cs
```diff
@@ -97,9 +97,13 @@ public async Task<float[]> GetEmbeddings(string text, bool addBos, CancellationT
 
         private float[] GetEmbeddingsArray()
         {
-            var embeddings = NativeApi.llama_get_embeddings_seq(Context.NativeHandle, LLamaSeqId.Zero);
-            if (embeddings.Length == 0)
-                return Array.Empty<float>();
+            var embeddings = NativeApi.llama_get_embeddings(Context.NativeHandle);
```
martindevans, line 100: In my own branch for investigating this issue I have rewritten the various get_embeddings methods to return a raw `float*`, instead of a `Span<float>` (just exposing the llama.cpp symbol directly, instead of wrapping it). Can I suggest making the same change in your PR?

Since `null` is a valid value for these functions to return a Span isn't really a suitable wrapper.

CMakeLists.txt
zsogitbe: > There's a lot of stuff in this PR not relevant to embeddings. Please remove the `LLama.GenerateSolution` folder.

Would you be able to make a separate PR for https://github.com/SciSharp/LLamaSharp/pull/674/commits/9c91fac20f3ebde5d1f1bc6a9feacaaa61c4d087 please? Maybe you can do it as admin.
This is GitHub merging all of my PRs. I do not like it very much.

LLama/LLamaEmbedder.cs
zsogitbe: Changing to float* is a lot of work because you need to change several functions... maybe it is better if you merge this PR and then change it as you like. Pointers are 'unsafe', so I am not sure if that is a good idea, but I trust that you know what you are doing.

CMakeLists.txt
martindevans: I can't do that, it's coming from your fork which I have no admin rights over.

LLama/LLamaEmbedder.cs
martindevans: Sure, I'll have a look at that once this one is merged :)

CMakeLists.txt
zsogitbe: I could split it. It is removed from here.

martindevans:
Looks good, I'll merge it once the tests pass. Thanks for helping me investigate the embeddings issue and working on this PR.

martindevans:
Oops, I pushed my followup changes to straight to master into of a PR branch 🤦‍♂️ 

@zsogitbe here's my followup - hopefully this looks ok to you? https://github.com/SciSharp/LLamaSharp/commit/3c764409579a417a85048984fb492b9530ef97d3

zsogitbe:
> Oops, I pushed my followup changes to straight to master into of a PR branch 🤦‍♂️
> 
> @zsogitbe here's my followup - hopefully this looks ok to you? [3c76440](https://github.com/SciSharp/LLamaSharp/commit/3c764409579a417a85048984fb492b9530ef97d3)

Maybe double check that `if (embeddings == null)` is enough. C++ may return an empty list and then it will not be NULL...

martindevans:
The C++ doesn't return a size, just a pointer to the first element. So it's either null or a valid embedding vector.

PR #676, opened 2024-04-19
Automatic Solution Generator - Work in progress

zsogitbe:
The console app asks the user for some input and then generates the VS solution automatically.

TODOs:

- [ ] apply the chosen parameters by the user - it is hardcoded now AVX2 + CUDA 
- [ ] create a template from CMakeList.txt.in by replacing the params we want to replace automatically with, for example {{param1}}
- [ ] replace the params {{paramn}} in CMakeList.txt.in automatically and copy the file to CMakeLists.txt
- [ ] select only the necessary projects in VS - at the moment all C++ projects are loaded what is too much

The app works already but with the hardcoded settings (see TODOs here above) and loads all C++ projects instead of only the necessary ones.

Maybe someone else can finalize this solution?


SignalRT:
If this PR intent to allow the developer to build llama.cpp It seems simpler to me dranger003 alternative. Just include the build in the .csproj. It works on Windows and Linux.

https://github.com/dranger003/llama.cpp-dotnet/blob/master/LlamaCppLib/LlamaCppLib.csproj

This readme describe how to use it:
https://github.com/dranger003/llama.cpp-dotnet/blob/master/README.md

zsogitbe:
> If this PR intent to allow the developer to build llama.cpp It seems simpler to me dranger003 alternative. Just include the build in the .csproj. It works on Windows and Linux.
> 
> https://github.com/dranger003/llama.cpp-dotnet/blob/master/LlamaCppLib/LlamaCppLib.csproj
> 
> This readme describe how to use it: https://github.com/dranger003/llama.cpp-dotnet/blob/master/README.md

[SignalRT](https://github.com/SignalRT), I have checked the project you mentioned here above. It works well for one specific setup (with a selection of hardcoded options), but our goal is to support several setups (combination of options). Also, it links the compilation of the C# part with the compilation of the C++ part what I think is not a good approach (we should have these two separated). What I find good is that the selection of C++ projects are limited to a few we need, we should check how this is done and do the same in our solution generator.
So, as a summary, our solution generator would be more versatile, with more options and easier to configure. The hardcoded cmake script in the dranger003 project file is a kind of hack and only text editor modifications are possible (it is not even in the project settings).

SignalRT:
@zsogitbe, I don't get the point.

1. We can execute the action to build ALL the binaries on GitHUB. Everybody can see the script and the output integrates directly in action execution. That covers the use case of the people that wants to be sure that the upstream binaries are "secure".
2. Any other use case is a developer that if wants to build all the binaries should use 1 or something similar. And if the developer want's to generate binaries for a specific configuration dranger003 solution seems to be enough for me.

Any way, for sure I'm missing the use case that you want to cover.

zsogitbe:
> @zsogitbe, I don't get the point.
> 
> 1. We can execute the action to build ALL the binaries on GitHUB. Everybody can see the script and the output integrates directly in action execution. That covers the use case of the people that wants to be sure that the upstream binaries are "secure".
> 2. Any other use case is a developer that if wants to build all the binaries should use 1 or something similar. And if the developer want's to generate binaries for a specific configuration dranger003 solution seems to be enough for me.
> 
> Any way, for sure I'm missing the use case that you want to cover.

[SignalRT](https://github.com/SignalRT), some of the current problems with GitHub builds are the following: cannot build with CUDA, we do not have debug versions, the DLLs do not include any vendor info (see below), downloading Dll's is still a security risk even from GitHub, etc.

image, 

I will not finalize this solution, I made it only to show you that it is possible (it works now with the hardcoded values = the same as dranger003 solution CUDA + AVX2). We could have a simple console app which configures several builds with a few user options. It is up to the team to decide if this is useful. If you find it useful, then you could merge it and someone else could finalize it.

PR #677, opened 2024-04-19
Made special tokens included in prompts tokenize as intended

Lyrcaxis:
There was an issue where executors wouldn't tokenize the special tokens at all, leaving them as text
(e.g.: `<|im_start|>` would be tokenized as [`<` + `|` + `im` + `_` + `start` + `|` + `>`] instead of a single token.)

I also saw no way of doing it via inference parameters. Not sure if that was intended but this PR fixes it.


martindevans:
Can you rebase this onto master? The CI is failing due to a mistake I made earlier (merging a bad commit to master)

Issue #670, opened 2024-04-16
[Proposal] Backend-free support


AsakusaRinne:
# Introduction

LLamaSharp uses llama.cpp as a backend, and have introduced [dynamic native library loading](https://scisharp.github.io/LLamaSharp/0.11.2/Tutorials/NativeLibraryConfig/), which allows us to choose which DLL to load at runtime. However, users still need to install the backend packages unless they have exactly one DLL to use. The problem is, at most of the time, an user only needs one DLL, for example, the CUDA11 one. However, many DLLs have to be included, especially if we support CUDA with AVX in the future.

Dividing into backend packages with a single file, as previously discussed in other issues, appears to be a solution. However, if the user has chosen a specific backend, what is the purpose of our backend selection strategy? Furthermore, this approach may lead to an excessive number of backends, causing potential difficulties.

Is it possible to select the native library by the configuration and system information, and only download the selected one, and without having too many backend packages? This is the point of this proposal.

# Brief Description

My idea is to put all the native library files on [HuggingFace](huggingface.co), then download the selected one according to the configuration and system information at runtime. That's all!

# APIs

The following APIs will be exposed for users to get this feature.

```cs
// Use along with other strategies such as `WithCuda`.
NativeLibraryConfig NativeLibraryConfig::WithAutoDownload(bool enable = true, string? cacheDir = null, string? endPoint = null, string? proxy = null, int timeout = -1);

// Explicitly download the library with filename and version.
void NativeLibraryConfig::DownloadNativeLibrary(string filename, string? version = null, string? cacheDir = null, string? endPoint = null, string? proxy = null, int timeout = -1);

// Explicitly download the library with specified configurations.
void NativeLibraryConfig::DownloadNativeLibrary(bool useCuda, AvxLevel avxLevel, string os = "auto", string? cacheDir = null, string? endPoint = null, string? proxy = null, int timeout = -1);

// Explicitly download the best library (for efficiency) selected by LLamaSharp according to detected system info.
void NativeLibraryConfig::DownloadBestNativeLibrary(string? cacheDir = null, string? endPoint = null, string? proxy = null, int timeout = -1);
```

p.s. To be honest, I don't think it's good to put the methods for downloading in `NativeLibraryConfig`, but I haven't come up  with a better idea yet.

# Behaviors

## Priorities

The most important thing is that what the behavior is when this feature is used with backend installed?

My answer would be that we'll follow the priorities below.

[...]

PR #688, opened 2024-04-23
feat: optimize the native library loading.

AsakusaRinne:
Implement the proposal in #670 

### Add `NativeLibraryConfig.WithAutoDownload`

By default it's disabled and the user could enable it with some settings.

### Add `NativeLibraryConfig.WithSelectingPolicy`

To allow users to customize the auto-selection process. Users could return a list of paths from `INativeLibrarySelectingPolicy.Select` , ordered by the priority.

### Make `NativeLibraryConfig.Description` public

I made this change to simplify some implementations. It seems not to be risky to expose this class.

### Add `NativeLibraryConfig.All`, `NativeLibraryConfig.LLama`, `NativeLibraryConfig.LLavaShared` and make `NativeLibraryConfig.Instance` obsolete.

There has been two native libraries now, so the name `Instance` seems no longer reasonable. With this modification, users could specify configurations for one of them. Meanwhile, using `NativeLibraryConfig.All` will apply same configurations to both of them.

### Add `NativeLibraryConfig.VersionMap`

Please tell me if there's a better design. I added it to provide a mapping from version tag (such as v0.11.2) to the llama.cpp commit hash, which is also the revision of huggingface remote branch.

### Add `NativeLibraryConfig.DryRun`

This API executes the whole loading process except assigning the loaded pointer to `NativeApi`. I added it for the following two reasons.

1. It will allow user to check the configuration at the runtime. If it returns `false`, the user could have chances to revise the configuration before loading the model.
2. The downloading is actually a asynchronized process. Though I synchronized it by calling `Task.Result`, I'm still not sure whether there's potential problem using async method in a static constructor. So, currently, users are supposed to call `DryRun` when auto-download is enabled.

## Other changes

- Add the package [HuggingfaceHub](https://github.com/AsakusaRinne/HuggingfaceHub) as a dependency.
- Adjust the file structures to make it more clear.

## TODO

- [x] Windows test
- [x] Linux test
- [x] MAC test
- [x] .NET standard2.0 test

## How to test it

**Comment all the content in LLama/LLamaSharp.Runtime.targets** and run the examples. It's better to enable the full logs (though in a bit mess now).


martindevans:
Can this be split into two parts (in two PRs):
1) Most of the changes you've made here to _enable_ downloading. This would include splitting the config, the dry run etc.
2) The actual downloader implementation.

With that split we've exposed all of the parts required for _custom_ downloaders to be implemented outside of the project, which then means people can use this auto download infrastructure but without having to trust our binaries at all.

If that separation is made we can also have the debate about if LLamaSharp should include a default implementation, I'm strongly of the opinion it _shouldn't_. We can provide a downloader as a separate package if we want (e.g. LLamaSharp.backend.AutoDownload backend). This would also mean we avoid adding another external dependency to LLamaSharp.

AsakusaRinne:
I'll modify this PR to leave only the parts not related with downloading and add the other as a separate PR.

AsakusaRinne:
@martindevans I've removed all things related with auto-download.

LLama/Native/Load/NativeLibraryConfig.cs
```diff
@@ -0,0 +1,605 @@
...
+    }
+#endif
+
+    public sealed partial class NativeLibraryConfig
+    {
+        /// <summary>
+        /// Set configurations for all the native libraries, including LLama and LLava
+        /// </summary>
+        [Obsolete("Please use NativeLibraryConfig.All instead, or set configurations for NativeLibraryConfig.LLama and NativeLibraryConfig.LLavaShared respectively.")]
+        public static NativeLibraryConfigContainer Instance { get; }
```
martindevans, line 268:
```suggestion
        public static NativeLibraryConfigContainer Instance => All;
```

If you change to this you can remove the code on line 306 which is generating a warning.

LLama/Native/Load/NativeLibraryConfig.cs
```diff
@@ -0,0 +1,605 @@
...
+
+        /// <summary>
+        /// Configuration for LLama native library
+        /// </summary>
+        public static NativeLibraryConfig LLama { get; }
+
+        /// <summary>
+        /// Configuration for LLava native library
+        /// </summary>
+        public static NativeLibraryConfig LLavaShared { get; }
```
martindevans, line 283: Should we just call it `LLava` instead?

LLama/Native/Load/NativeLibraryConfig.cs
```diff
@@ -0,0 +1,605 @@
...
+
+        /// <summary>
+        /// A dictionary mapping from version to corresponding llama.cpp commit hash.
+        /// The version should be formatted int `[major].[minor].[patch]`. But there's an exceptance that you can 
+        /// use `master` as a version to get the llama.cpp commit hash from the master branch.
+        /// </summary>
+        public static Dictionary<string, string> VersionMap { get; } = new Dictionary<string, string>()
```
martindevans, line 290: Does this need to be a dictionary? Could it just be a constant, with the commit hash (either of llama.cpp, or llamasharp, or both) hardcoded?

LLama/Native/Load/NativeLibraryConfig.cs
```diff
@@ -0,0 +1,605 @@
...
+    public sealed class NativeLibraryConfigContainer
+    {
+        private NativeLibraryConfig[] _configs;
+
+        /// <summary>
+        /// All the configurations in this container.
+        /// Please avoid calling this property explicitly, use <see cref="NativeLibraryConfig.LLama"/>
+        /// and <see cref="NativeLibraryConfig.LLavaShared"/> instead.
+        /// </summary>
+        public NativeLibraryConfig[] Configs => _configs;
```
martindevans, line 388: Can this be made `internal` if people aren't meant to access it directly?

LLama/Native/Load/NativeLibraryConfig.cs
```diff
@@ -0,0 +1,605 @@
...
+        #endregion
+
+        /// <summary>
+        /// Try to load the native library with the current configurations, 
+        /// but do not actually set it to <see cref="NativeApi"/>.
+        /// 
+        /// You can still modify the configuration after this calling but only before any call from <see cref="NativeApi"/>.
+        /// </summary>
+        /// <returns>Whether the running is successful.</returns>
+        public bool DryRun()
```
martindevans, line 569: This returns if it was succesful, but doesn't return _which_ native library was selected. That seems a bit odd for a dry run?

LLama/Native/Load/NativeLibraryFromPath.cs
```diff
@@ -0,0 +1,31 @@
+﻿using LLama.Abstractions;
+using System.Collections.Generic;
+
+namespace LLama.Native
+{
+    /// <summary>
+    /// A native library specified with a lcoal file path.
```
martindevans, line 7:
```suggestion
    /// A native library specified with a local file path.
```

LLama/Native/Load/NativeLibraryConfig.cs
AsakusaRinne: It's okay to me. @SignalRT Any idea about this since you added the support for LLaVA?

LLama/Native/NativeLogConfig.cs
martindevans, line 1: Should this be changed to `All` instead of `Instance` to fix those warnings?

martindevans:
I haven't gone through this fully in depth, but there's a few minor nitpicks I noticed above.

Overall this looks like a good direction. I like the idea to split the various type of native libs up into separate files altogether, that'll get some of the complexity of the system udner control!

LLama/Native/Load/NativeLibraryConfig.cs
AsakusaRinne: Using a dictionary here makes it possible for users to specify another version of native library when auto-download is enabled. This will make the behavior the same with our backend packages -- users could install a backend with version different with the main package.

I used to consider two alternatives for this:
- Hardcode the default version but allows to load a version map from a file outside.
- Just move the version map to #692

martindevans: That's not something that would ever work though? The version of llamasharp is always tied to exactly one version of llama.cpp.

LLama/Native/Load/NativeLibraryConfig.cs
AsakusaRinne: I exposed it because it's needed to keep auto-download in another package. I could remove it and hardcode it in #692

LLama/Native/Load/NativeLibraryConfig.cs
AsakusaRinne: You're right! Maybe the following API will be good

```cs
public bool DryRun(out var ILibrary?)
```

LLama/Native/NativeLogConfig.cs
```diff
@@ -37,7 +37,7 @@ public static class NativeLogConfig
     public static void llama_log_set(LLamaLogCallback? logCallback)
 #pragma warning restore IDE1006 // Naming Styles
     {
-        if (NativeLibraryConfig.LibraryHasLoaded)
+        if (NativeLibraryConfig.LLama.LibraryHasLoaded)
```
AsakusaRinne, line 40: @martindevans Somehow Github failed to show the code location for your previous comment https://github.com/SciSharp/LLamaSharp/pull/688#discussion_r1580318447 so I tag you here. It seems that `NativeLogConfig` only sets log callback for `llama` instead of `llava_shared`. Will it be more accurate to set it only depending on the settings in `NativeLibraryConfig.LLama`? Please correct me if I'm wrong. :)

LLama/Native/Load/NativeLibraryConfig.cs
AsakusaRinne: What about adding a method to get the llama.cpp commit hash but hide its implementation?

```cs
public string NativeLibraryCommitHash {get;}

private const string version;
private const string commitHash;
```

Thus it will not impact on the public API if we really want to support what I said above in the future.

LLama/Native/Load/NativeLibraryConfig.cs
SignalRT: It's ok to me. The reason was the name of the library used in llama.cpp.

LLama/Native/NativeLogConfig.cs
martindevans: As far as I know there's no log callback for llava, so I guess this should be done entirely based on llama only.

LLama/Native/Load/UnknownNativeLibrary.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public class UnknownNativeLibrary: INativeLibrary
+    {
+        /// <inheritdoc/>
+        public NativeLibraryMetadata? Metadata => null;
+
+        /// <inheritdoc/>
+        public IEnumerable<string> Prepare(SystemInfo systemInfo, NativeLogConfig.LLamaLogCallback? logCallback = null)
+        {
+            throw new NotImplementedException("This class is only a placeholder and should not be used to load native library.");
```
martindevans, line 22:
```suggestion
            throw new NotSupportedException("This class is only a placeholder and should not be used to load native library.");
```

AsakusaRinne:
Perhaps I'm going to merge this PR in several days. #692 and #738 will be both blocked by this PR. Test in windows, linux, MAC and .NET standard2.0 have all been passed. I'll fix the possible other Bugs after this PR.

PR #702, opened 2024-04-27
Interruptible Async Model Loading With Progress Monitoring

### LINKED ISSUE #699: [Feature] Allow async model loading and cancellation
In production environment, especially desktop apps, it's common to have a button (or any other way) to allow users to abort the model loading. Fortunately, llama.cpp has already added support for it https://github.com/ggerganov/llama.cpp/pull/4462. I think we should introduce this feature in LLamaSharp.

Similarly, async model loading is also important for applications based on LLamaSharp, which avoids blocking the main thread for a long time when loading a large model. I've found a similar work of it in the node.js binding of llama.cpp https://github.com/withcatai/node-llama-cpp/pull/178. We could also implement it by polling the progress callback.

martindevans:
- Added `LLamaWeights.LoadFromFileAsync`.
 - Updated examples to use `LoadFromFileAsync`
 - Async loading supports cancellation through a `CancellationToken`:
   - If loading is cancelled an `OperationCanceledException` is thrown.
   - If it fails for another reason a `LoadWeightsFailedException` is thrown.

This does _not_ expose the progress callback on the `ModelParams` object. However the code as written does support custom callbacks being set so that can be easily exposed in the future.

Resolve #699

### DIFF
--- a/LLama/Native/LLamaModelParams.cs
+++ b/LLama/Native/LLamaModelParams.cs
@@ -38,7 +38,7 @@ namespace LLama.Native
         // as NET Framework 4.8 does not play nice with the LlamaProgressCallback type
         public IntPtr progress_callback;
 #else
-        public LlamaProgressCallback progress_callback;
+        public LlamaProgressCallback? progress_callback;
 #endif
 
         /// <summary>
--- a/LLama.Examples/Examples/BatchedExecutorFork.cs
+++ b/LLama.Examples/Examples/BatchedExecutorFork.cs
@@ -19,7 +19,7 @@ public class BatchedExecutorFork
         string modelPath = UserSettings.GetModelPath();
 
         var parameters = new ModelParams(modelPath);
-        using var model = LLamaWeights.LoadFromFile(parameters);
+        using var model = await LLamaWeights.LoadFromFileAsync(parameters);
 
         var prompt = AnsiConsole.Ask("Prompt (or ENTER for default):", "Not many people know that");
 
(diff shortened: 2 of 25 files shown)


LLama/Native/LLamaModelParams.cs
```diff
@@ -38,7 +38,7 @@ public unsafe struct LLamaModelParams
         // as NET Framework 4.8 does not play nice with the LlamaProgressCallback type
         public IntPtr progress_callback;
 #else
-        public LlamaProgressCallback progress_callback;
+        public LlamaProgressCallback? progress_callback;
```
AsakusaRinne, line 41: Recently I'm wondering if it's better to use a modern progress interface here ([`IProgress<T>`](https://learn.microsoft.com/en-us/dotnet/api/system.iprogress-1?view=net-8.0)). In this way the `LLamaContextHandle` object would not be exposed to users who works with LLamaSharp high-level APIs. For us (or other experienced users), we won't use this parameter in a wrong way. But for some users who know little about our backend, it might be confusing.

I prefer to switching to `IProgress<float>` smoothly, maybe not in this PR though. :)

AsakusaRinne (approved):
Great work! You are the flash!

LLama/Native/LLamaModelParams.cs
martindevans: `IProgress` is a good idea 👍 

I intentionally didn't expose the raw progress callback to the user in this PR because it does feel a bit ugly - it's combining progress and cancellation together. We should probably expose it just in case someone wants to use it, but have better options available for most users.

I'll have a look at adding an `IProgress` parameter to `LoadFromFileAsync` now.

martindevans:
I added an `IProgress` parameter to the `LoadFromFileAsync` from method as well.

It's not _quite) reporting the raw values. If there are LoRAs to apply it reports model loading in the `[0 -> 0.9]` range and then reports LoRA loading progress in the `[0.9 -> 1.0]` range.

AsakusaRinne (approved):

PR #709, opened 2024-04-29
ci: add a workflow to check code format.

AsakusaRinne:
As the size of our project is increasing, it seems necessary to add an `.editorconfig` file to specify how the code should be formatted. Besides, a workflow is included in this PR to check the code format.

To ensure passing the check, the only thing needed to do is running the following command before the PR.

```cs
dotnet format
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,86 @@
+
+[*]
+charset = utf-8
+end_of_line = lf
+trim_trailing_whitespace = false
+insert_final_newline = false
+indent_style = space
+indent_size = 4
+
+# Microsoft .NET properties
+csharp_new_line_before_members_in_object_initializers = false
+csharp_preferred_modifier_order = public, private, protected, internal, file, new, static, abstract, virtual, sealed, readonly, override, extern, unsafe, volatile, async, required:suggestion
+csharp_style_prefer_utf8_string_literals = true:suggestion
+csharp_style_var_elsewhere = true:suggestion
+csharp_style_var_for_built_in_types = true:suggestion
+csharp_style_var_when_type_is_apparent = true:suggestion
+dotnet_naming_rule.enum_member_rule.import_to_resharper = True
+dotnet_naming_rule.enum_member_rule.resharper_description = Enum members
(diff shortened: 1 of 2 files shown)


LLama/Extensions/DictionaryExtensions.cs
```diff
@@ -10,7 +10,7 @@ internal static class DictionaryExtensions
             return GetValueOrDefaultImpl(dictionary, key, defaultValue);
         }
 #elif !NET6_0_OR_GREATER && !NETSTANDARD2_1_OR_GREATER
-    #error Target framework not supported!
+#error Target framework not supported!
```
martindevans, line 13: I prefer the old syntax - for the same reason you indent any other if statement!

martindevans:
Personally I'm not a big fan of auto formatters, I'm of the opinion they're a good way to make sure you're formatting isn't _bad_ but also a good way to ensure it is uniformly mediocre! Is it possible to add this in a "softer" way - e.g. generate warnings that are surfaced as part of CI, but aren't a hard failure?

As for the changeset itself, it's pretty huge and difficult to review! Can this be broken up into smaller PRs? e.g. apply it locally, and submit small groups of files with related changes. As an example, one thing I have avoided doing for a long time is replacing `namespace X { }` with `namespace X;` because that generates an enormous amount of noise in the change (touching every single line in every file).

AsakusaRinne:
TBH I'm not strongly pushing this changeset. It's okay for me to keep what the code was.

> add this in a "softer" way - e.g. generate warnings that are surfaced as part of CI, but aren't a hard failure?

I prefer not using `warning` because most people seldom care the warning! I think we could disable the workflow but leave `.editorconfig` there to let user know what is the format configuration of our code (VS and Rider both support formatting on saving depending on `.editorconfig`). The `.editorconfig` in this PR was generated from format auto-detection of jetbrains Rider.

martindevans:
> I think we could disable the workflow but leave .editorconfig there to let user know what is the format configuration of our code

Yeah I think this is a better way to go for now. That way we can submit smaller PRs over time to bring the overall codebase better in line with the config, and also adjust the config when it's wrong.

PR #710, opened 2024-04-29
ci: add workflow to check the spellings.

AsakusaRinne:
Typo (incorrect spelling) is a problem that's hard to be completely eliminated during the review. When we realize a spelling problem, it might have already come across several versions. This PR introduces a workflow to detect the possible typos to help us find them.

If a word or a file is incorrectly checked, you can modify `.github/_typo.toml` to skip check for it. Here's the [reference](https://github.com/marketplace/actions/typos-action).


LLama/LLamaExecutorBase.cs
```diff
@@ -209,7 +209,7 @@ protected virtual void HandleRunOutOfContext(int tokensToKeep)
         /// <summary>
         /// Try to reuse the matching prefix from the session file.
         /// </summary>
-        protected virtual void TryReuseMathingPrefix()
```
martindevans, line 212: This is _technically_ a breaking prefix. But I'm not too concerned about it in this case tbh.

martindevans:
Looks good! The number of typos this has caught already demonstrates it's worth!

LLama/LLamaExecutorBase.cs
AsakusaRinne: We could mention it in the next release note. :)

PR #711, opened 2024-04-29
Optional IHistoryTransform added to ChatSession.InitializeSessionFromHistoryAsync

Norne9:
This pull request introduces an optional parameter, `IHistoryTransform? transform`, to the `InitializeSessionFromHistoryAsync` method in the `ChatSession` class. This change accommodates models that require specific formatting of the chat history.

Some language models may require the chat history to be formatted in a specific way before being passed to the model for generating a response. The addition of the `IHistoryTransform` parameter allows for the flexibility to apply custom transformations to the chat history when initializing a new session.

This change maintains backward compatibility with existing code that does not provide a custom history transform by making the `transform` parameter optional and defaulting it to `null`.

### DIFF
--- a/LLama/ChatSession.cs
+++ b/LLama/ChatSession.cs
@@ -74,15 +74,21 @@ public class ChatSession
     /// </summary>
     /// <param name="executor">The executor for this session</param>
     /// <param name="history">History for this session</param>
-    /// <returns></returns>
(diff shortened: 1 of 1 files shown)


AsakusaRinne (approved):
LGTM, thank you for the contribution!

Issue #695, opened 2024-04-25
Android Backend


AmSmart:
I was trying to get this up and running with MAUI on Android but as far as I can tell, only desktop support is available for the time being.  Will the maintainers be open to a PR that merges android CPU support in?

martindevans:
Android support would be great!

If you're interested in working on it I'd suggest starting out by modifying [compile.yml](https://github.com/SciSharp/LLamaSharp/blob/master/.github/workflows/compile.yml) to produce Android binaries. That script is what we use to produce all of the binaries we distribute. Once that's merged we can start looking at what modifications are required for runtime support (if any).

AmSmart:
Sounds good, I'll start working on this

DGdev91:
So.... how' the progress on this?

Some time ago i managed to build up a working example on Android (just an experiment to see if it was possible to run an LLM inside Godot Engine, using the c# backend), just by building myself the libraries and loading them using the native lib classes.

The llama.cpp version used for v.0.15 seems like to have a bug wich makes it impossible to compile on android. 
The current upstream builds normally, but i can't make it work, not sure if i'm doing something wrong or it's just because the newer versions introduced some breaking changes (i know, i'm supposed to only use that exact version)

martindevans:
I introduced the changes into the build pipeline before 0.15 (see [here](https://github.com/SciSharp/LLamaSharp/blob/master/.github/workflows/compile.yml#L358)) but they're currently disabled, because when I came to do the 0.15 build it failed on Android. I think it was a temporary upstream issue with Android builds, so I'm expecting to be able to enable that for 0.16.

What we don't have yet is any runtime loading of those binaries. There needs to be an item which dynamically detects when we're on Android, checks what CPU it is running on, and loads the right binaries (see [here](https://github.com/SciSharp/LLamaSharp/tree/master/LLama/Native/Load)). That's the next piece of work to be done, if you're interested in contributing towards Android support :)

DGdev91:
> I introduced the changes into the build pipeline before 0.15 (see [here](https://github.com/SciSharp/LLamaSharp/blob/master/.github/workflows/compile.yml#L358)) but they're currently disabled, because when I came to do the 0.15 build it failed on Android. I think it was a temporary upstream issue with Android builds, so I'm expecting to be able to enable that for 0.16.
> 
[...]

PR #489, opened 2024-02-05
Create nuspec for OpenCL

jasoncouture:
Depends on #487 - Adds a NuGet package for OpenCL (CLBLAST)

PR #717, opened 2024-05-04
Android Backend Support

AmSmart:
As discussed in #695, Here's a PR that generates the shared library for Android. I have also added the new binaries to be copied in the CPU Backend ``.nuspec`` file. Please let me know if I am missing anything else.

I have tested the generated ``.so`` by loading manually into a .NET MAUI app which ran smoothly on my Android device. See video below

https://github.com/SciSharp/LLamaSharp/assets/26899350/d6afe4d9-21a4-455d-a74e-5a472249b2ea

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -51,12 +51,12 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           path: ./build/libllama.so
-          name: llama-bin-linux-${{ matrix.build }}-x64.so
+          name: llama-bin-linux-${{ matrix.build }}-x64
       - name: Upload Llava
         uses: actions/upload-artifact@v4
         with:
           path: ./build/examples/llava/libllava_shared.so
-          name: llava-bin-linux-${{ matrix.build }}-x64.so
+          name: llava-bin-linux-${{ matrix.build }}-x64
 
   compile-windows:
     name: Compile (Windows)
@@ -92,13 +92,13 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           path: .\build\bin\Release\llama.dll
-          name: llama-bin-win-${{ matrix.build }}-x64.dll
+          name: llama-bin-win-${{ matrix.build }}-x64
 
       - name: Upload Llava
         uses: actions/upload-artifact@v4
         with:
           path: .\build\bin\Release\llava_shared.dll
-          name: llava-bin-win-${{ matrix.build }}-x64.dll
+          name: llava-bin-win-${{ matrix.build }}-x64
 
   compile-clblast:
     name: Compile (clblast) - ${{ matrix.os }}
@@ -175,14 +175,14 @@ jobs:
           path: |
             .\build\bin\Release\llama.dll
             .\build\bin\Release\clblast.dll
-          name: llama-bin-win-clblast-x64.dll
+          name: llama-bin-win-clblast-x64
       - name: Upload llava artifacts (Windows)
         if: ${{ matrix.os == 'windows-latest' }}
         uses: actions/upload-artifact@v4
         with:
           path: |
             .\build\bin\Release\llava_shared.dll
-          name: llava-bin-win-clblast-x64.dll
+          name: llava-bin-win-clblast-x64
       - name: Upload artifacts (linux)
         if: ${{ matrix.os == 'ubuntu-22.04' }}
         uses: actions/upload-artifact@v4
@@ -190,14 +190,14 @@ jobs:
           path: |
             ./build/libllama.so
             # ./build/libclblast.so
-          name: llama-bin-linux-clblast-x64.so
+          name: llama-bin-linux-clblast-x64
       - name: Upload llava artifacts (linux)
         if: ${{ matrix.os == 'ubuntu-22.04' }}
         uses: actions/upload-artifact@v4
         with:
           path: |
             ./build/examples/llava/libllava_shared.so
-          name: llava-bin-linux-clblast-x64.so
+          name: llava-bin-linux-clblast-x64
 
           
   compile-cublas:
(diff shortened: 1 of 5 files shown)


martindevans:
Looks good, thanks for putting this together.

Could you please manually trigger a run of the "Update Binaries" action on your repo as a test run. To do that:
1. Go to actions (top bar of your repo)
2. Select "Update Binaries" on the left
3. Click "Run Workflow" at the top of the run list (right side)
4. Change the branch to **add-android-support**
5. Click run workflow
6. Link the run here

AsakusaRinne (changes requested):
Thank you for this contribution! The overall looks good except one concern. I prefer to putting the android native libraries in a new package `LLamaSharp.Backend.Android` because those who want to use android will never use libraries of win/linux/osx. For those who want to publish a cross-platform app, they could install both `LLamaSharp.Backend.Android` and `LLamaSharp.Backend.Cpu`. Please refer to #489 to see how to add a new backend package.

The following items will further improve it but are not required in this PR.

1. Add a workflow to run test in android. I know little about running C# apps on android but it seems that github action has already provided such support ([reference](https://github.com/marketplace/actions/run-android-instrumentation-tests)).
2. Add logics to support dynamic native library selection on android.

I'd like to work with item 2 after having this PR merged so please don't worry about it. :) Would you like to work with item 1 (in this PR or after this PR are both okay).

AmSmart:
Totally aligned. I was pondering that it deserved its own backend as well but wasn't sure. Will work on these over the next few days. Should share an update soon

cmpktheo:
Could you upload the MAUI app's source with the android runtime libraries to test it?

AmSmart:
Hi @cmpktheo Attaching a binary compliled for ``android-arm64-v8a``

cmpktheo:
Thanks @AmSmart !
I have been trying to build llama.cpp on a linux VM and test a MAUI Android app with LlamaSharp using a Visual Studio x86_64 emulator. 

I followed the android build instructions on llama.cpp page, but when I try to use the .so file, I get a null pointer calling the llama_model_default_params method (which basically returns a dummy structure) . It makes me think that there's either a flaw in my build process or some issue related to MAUI. This is my first attempt to build anything in MAUI so I might miss something.
I've created a windows MAUI app that loads the GPT2 guff model and works perfectly.

I would appreciate if you could share the build process so I can apply it to x86_64 architecture or a ready build x86_64 runtime and ideally a MAUI-Android sample app.

cmpktheo:
@martindevans  the issue seems to be with my build process since the native method returns null pointer.

AmSmart:
@cmpktheo You can check the ``compile.yml`` file from my PR. I've updated it to include other Android architectures.

AmSmart:
@AsakusaRinne I've been a bit busy since our last conversation but I've finally gotten some time to address some of the concerns you listed earlier and made some other changes as well.

**Things to note**

- I have created a different Nuspec for the Android backend as discussed
- I noticed there was a ``.targets`` file but it wasn't been used by any of the generated packages (I double checked this by unzipping existing packages on Nuget). That said, I needed the ``targets`` file to Embed the library and link for Android so I removed all the dormant build directives in the file and have just the Android ones in it. I plan to open another PR for iOS and it would need the ``targets`` file as well. If you do not like these change, or need a bit more clarification. I'm happy to explain more
- Since Android doesn't support additional enhancements (CUDA, CUBLAS, AVX etc) at the moment, there is really no need to manually resolve the native library. I have added an ``if`` statement to exclude Android from the Manual Resolution in ``NativeApi.Load.cs``
- Updated the ``compile.yml`` file to remove file extensions from directory names. I think it may confuse new contributors who may confuse the directories for files (I was confused for a moment).
- With regards to providing tests, I'm thinking of a world where we have a Demo/Example .NET Android Project instead. The UnitTests already provide coverage for the underlying code and a demo app should suffice. That said, we could still proceed to write tests against the demo app and have them run in the workflow. Let me know what you think, I'm happy to contribute this in another PR.
 


Let me know if there are anymore requested changes for this PR to be good to go.

LLama/LLamaSharp.Runtime.targets
```diff
@@ -2,133 +2,17 @@
     <PropertyGroup>
         <IncludeBuiltInRuntimes Condition="'$(IncludeBuiltInRuntimes)' == ''">true</IncludeBuiltInRuntimes>
     </PropertyGroup>
-    <ItemGroup Condition="'$(IncludeBuiltInRuntimes)' == 'true'">
 
-      <None Include="$(MSBuildThisFileDirectory)runtimes/deps/llama.dll">
```
AsakusaRinne, line 7: These targets are used in the unittest and example projects, for example:

https://github.com/SciSharp/LLamaSharp/blob/master/LLama.Examples/LLama.Examples.csproj#L2

https://github.com/SciSharp/LLamaSharp/blob/master/LLama.Unittest/LLama.Unittest.csproj#L2

Therefore, though I'm not sure whether this targets file should also be used in other backend packages, I don't think we should remove it. May I ask which behavior will be different if a targets file is used in the backend package?

LLama/Native/NativeApi.Load.cs
```diff
@@ -57,6 +57,11 @@ private static void SetDllImportResolver()
             // NativeLibrary is not available on older runtimes. We'll have to depend on
             // the normal runtime dll resolution there.
 #if NET5_0_OR_GREATER
+
+            // We don't need special dll resolution on Android
+            if (OperatingSystem.IsAndroid())
```
AsakusaRinne, line 62: The master branch has refactored the native library loading system so I'm afraid this code would not work. Don't worry, I can do it after having this PR merged.

.github/workflows/compile.yml
```diff
@@ -51,12 +51,12 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           path: ./build/libllama.so
-          name: llama-bin-linux-${{ matrix.build }}-x64.so
+          name: llama-bin-linux-${{ matrix.build }}-x64
```
AsakusaRinne, line 54: @martindevans Could you please take a look of this? I'm not sure what it's expected to do with the original code.

AsakusaRinne (changes requested):
Thank you for your effort! It looks good to me except a few minor issues.

> With regards to providing tests, I'm thinking of a world where we have a Demo/Example .NET Android Project instead. The UnitTests already provide coverage for the underlying code and a demo app should suffice. That said, we could still proceed to write tests against the demo app and have them run in the workflow. Let me know what you think, I'm happy to contribute this in another PR.

Sure, the workflow is not required in this PR. It's a good idea to add an example/demo first. The main motivation for the android test workflow is that our core team members do not have android environment to test it. Thus it will be hard for us to know whether a PR breaks this feature. But no hurry, adding an example is a good first step. :)

> Since Android doesn't support additional enhancements (CUDA, CUBLAS, AVX etc) at the moment, there is really no need to manually resolve the native library. I have added an if statement to exclude Android from the Manual Resolution in NativeApi.Load.cs

The native library loading system is also responsible for deciding the file path to load it. However, please don't worry, I'll do it after this PR.

cmpktheo:
> @cmpktheo You can check the `compile.yml` file from my PR. I've updated it to include other Android architectures.

Your compiled .so file for the android-arm64-v8a architecture, is ~12 MB.
When I run the Update Binaries workflow, the resulting .so file for the same architecture is ~17MB.
I am asking because, when I use the compiled .so file for the x86_64 architecture I get a [libc] Fatal signal 11 error, which means that there's an issue with the build process or MAUI/.NET.

How did you get the 12MB which is considerably smaller than my my run of Update Binaries workflow?

martindevans:
Since this PR has been stalled for a while I've taken the first step (generating the binaries) and added it in #861, hopefully we can start to make some progress on Android support :)

martindevans:
Some of this work was pulled out into a separate PR (#861). The rest of the work will need to be updated to fix the conflicts in new PRs, so I'll close this one. It should still serve as a handy reference for future PRs.

AmSmart:
Will pick up the rest of the effort in another PR as I find time.

martindevans:
Hopefully I've got the most complex bit out of the way for you with the updated build script :)

PR #718, opened 2024-05-06
Several updates to web project

Lamothe:
* Updated to .NET 8
* Updated to Bootstrap 5.3.3
* Use libman for client side packages
* Enabled razor runtime compilation in dev
* Added fav icon
* Removed the footer
* Added theme switcher


Lamothe:
For your consideration.  This will be the first of many updates that I want to make to the website sample.  Please note that .NET 7 is out of support in less than 2 weeks so I updated it to .NET 8 while I was in there.

Lamothe:
Agreed, especially because I'm only using 6 of those icons.  This is my first time using the new Bootstrap Icons, I've only used Font Awesome icons in the past.  There's a way that I can what I need from a CDN without pulling the entire module which includes the  individual icon files.  I'll do that for now and see if that helps.

LLama.Web/Common/ModelOptions.cs
```diff
@@ -1,117 +1,116 @@
...
+    /// Model friendly name
+    /// </summary>
+    public string Name { get; set; }
 
-        /// <summary>
-        /// Max context insta=nces allowed per model
-        /// </summary>
-        public int MaxInstances { get; set; }
+    /// <summary>
+    /// Max context insta=nces allowed per model
```
AsakusaRinne, line 16:
```suggestion
    /// Max context instances allowed per model
```

LLama.Web/Models/LLamaModel.cs
```diff
@@ -1,108 +1,111 @@
...
-        /// </summary>
-        public IModelParams ModelParams => _config;
-
-        /// <summary>
-        /// Gets the LLamaWeights
-        /// </summary>
-        public LLamaWeights LLamaWeights => _weights;
+        _config = modelParams;
+        _llamaLogger = llamaLogger;
+        _weights = LLamaWeights.LoadFromFile(modelParams);
```
AsakusaRinne, line 26: Async loading has been supported in #702, could you please introduce it here? Since async is not allowed in a class constructor, you could set the constructor as private and add a method `LLamaModel.CreateAsync` instead. For web app I think async loading is better because the UI would not be blocked when loading a large model.

AsakusaRinne (changes requested):
The overall looks good to me except a few concerns. Could you please also update [the LLama.Web readme](https://github.com/SciSharp/LLamaSharp/tree/master/LLama.Web) since this PR has changed the UI?

LLama.Web/Models/LLamaModel.cs
Lamothe: Will do.

SignalRT:
it looks ok to me. it seems that there is a conflict in ModelOptions.cs. Could you solve this one before merge the PR?

I was thinking on introducing the capability to use Multi-Modal models on the Web UI. I would like to know your plans to change the Web UI to not duplicate work.

AsakusaRinne (approved):

AsakusaRinne:
Merging this PR now, thank you a lot for the contribution!

PR #726, opened 2024-05-10
KernelMemory bug fix

zsogitbe:
While using WithLLamaSharpDefaults it was not possible to dispose GPU memory because the embedding was not defined well. This PR fixes that issue and also fixes some small problems with not setting all important model parameters.

### DIFF
--- a/LLama.KernelMemory/BuilderExtensions.cs
+++ b/LLama.KernelMemory/BuilderExtensions.cs
@@ -81,15 +81,15 @@ namespace LLamaSharp.KernelMemory
         {
             var parameters = new ModelParams(config.ModelPath)
             {
-                ContextSize = config?.ContextSize ?? 2048,
-                Seed = config?.Seed ?? 0,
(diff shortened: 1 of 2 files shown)


AsakusaRinne:
Hi, may I ask why this change will fix GPU memory disposing problem?

zsogitbe:
> WithLLamaSharpDefaults

Because in WithLLamaSharpDefaults the embedding generator is created but without an owner, so it does not get disposed... When using the LLamaSharpTextEmbeddingGenerator() constructor instead of the LLamaEmbedder() constructor the LLamaSharpTextEmbeddingGenerator will have full ownership and the embedder will be disposed.

AsakusaRinne (approved):

martindevans:
Is `config` allowed to be null when passed in? It's not marked with billable annotations (implying it's not) but it is using null coalescing to fall back to defaults (implying it is).

martindevans (approved):

PR #729, opened 2024-05-11
Fix context params defaults

### LINKED ISSUE #716: [BUG]: DefragThreshold default is not matching llama.cpp and probably not intended
### Description

`DefragThreshold` is 0 by default in LLamaSharp with the intent that it will disable defragmentation on the threshold excess.

https://github.com/SciSharp/LLamaSharp/blob/6bf010d719ec10c7b224d60deaf5b5c79a742a65/LLama/Abstractions/IContextParams.cs#L112

https://github.com/SciSharp/LLamaSharp/blob/6bf010d719ec10c7b224d60deaf5b5c79a742a65/LLama/Common/ModelParams.cs#L103

But actually it should be negative to disable this mechanics (-1). 0 means "do defrag when the smallest fragmentation discovered".

This may negatively (maybe sometimes positively) impact performance in some advanced cases, for example with batching.

See llama.cpp default https://github.com/ggerganov/llama.cpp/blob/3ab8b3a92ede46df88bc5a2dfca3777de4a2b2b6/llama.cpp#L11630

### Reproduction Steps

\-

\-

### Known Workarounds

_No response_

dlyz:
fixes #716 

- DefragThreshold is nullable now and disabled by default
- Seed is nullable now

There is also a possibility that the user will use special values directly instead of null. For example 0xFFFFFFFF for seed and it will lead to the same behavior as null (random seed). May not be obvious, but not sure what wee can do with that.

### DIFF
--- a/LLama.Web/Common/ModelOptions.cs
+++ b/LLama.Web/Common/ModelOptions.cs
@@ -1,4 +1,4 @@
-﻿using System.Text;
+using System.Text;
 using LLama.Abstractions;
 using LLama.Native;
 
@@ -32,7 +32,7 @@ namespace LLama.Web.Common
         public uint SeqMax { get; }
 
         /// <inheritdoc />
-        public uint Seed { get; set; } = 1686349486;
+        public uint? Seed { get; set; } = 1686349486;
 
         public bool Embeddings { get; }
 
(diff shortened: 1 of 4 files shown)


martindevans (approved):

martindevans:
Looks good, thanks for this!

PR #735, opened 2024-05-12
Less Sampler Allocations

martindevans:
Reduced the number of allocations required for sampling, by allowing passing in a temporary blob of memory to work in.
 - Using this new method in `LLamaContext` and `BaseSamplingPipeline`.
 - Using `Guidance` method in guidance example, instead of low level one working directly on logits.
 - Fixed `Guidance` method passing incorrectly sized span.


martindevans:
@Lyrcaxis I'd appreciate your review on this since you've been looking at sampling things recently.

LLama/Native/LLamaTokenDataArray.cs
```diff
@@ -342,7 +362,7 @@ public struct LLamaTokenDataArrayNative
         /// A pointer to an array of LlamaTokenData
         /// </summary>
         /// <remarks>Memory must be pinned in place for all the time this LLamaTokenDataArrayNative is in use</remarks>
-        public IntPtr data;
+        public unsafe LLamaTokenData* data;
```
AsakusaRinne, line 365: Is it possible to avoid exposing this member? It's a raw pointer so it's a bit dangerous to expose it to users.

AsakusaRinne (changes requested):
The overall looks good except two concerns. It seems that this PR will obviously improve the performance, which is inspiring!

LLama/Native/LLamaTokenDataArray.cs
martindevans: This is the _very_ low level structure that just directly represents the equivalent llama.cpp struct so it has to be like this (I actually consider an explicit pointer _safer_ than an IntPtr, since at least it carries type data with it. wheras an IntPtr is a `void*` in disguise).

AsakusaRinne: Right that makes sense. What about adding a comment to tell users not to modify the data? Though `LLamaTokenData*` is more type-safe but I'm afraid that users may write code like `tokenDataArray.data[0] = new LLamaTokenData(...);`. That seems to be bit shocking! :D

LLama/Sampling/BaseSamplingPipeline.cs
```diff
@@ -13,15 +13,21 @@ public abstract class BaseSamplingPipeline
     /// Grammar to constrain valid tokens
     /// </summary>
     public SafeLLamaGrammarHandle? Grammar { get; set; }
+    
+    private LLamaTokenData[]? _temporarySampling;
```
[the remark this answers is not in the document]
martindevans: done

LLama/Native/LLamaTokenDataArray.cs
martindevans: It's pretty hard to get your hands on a `LLamaTokenDataArrayNative` in the first place - the higher level `LLamaTokenDataArray` is what's normally exposed through APIs. You would have to deliberately construct one and then the only way you could use it would be to directly call the low level APIs.

AsakusaRinne (approved):

LLama/Native/LLamaTokenDataArray.cs
martindevans: @AsakusaRinne I realised I can do even better than a comment and changed the whole thing over to a span, so it's completely safe!

LLama/Native/LLamaTokenDataArray.cs
```diff
@@ -38,12 +38,32 @@ public LLamaTokenDataArray(Memory<LLamaTokenData> tokens, bool isSorted = false)
...
+        /// <param name="logits"></param>
+        /// <param name="temporary">Temporary memory which will be used to work on these logits. Must be at least as large as logits array</param>
+        /// <returns></returns>
+        public static LLamaTokenDataArray Create(ReadOnlySpan<float> logits, Memory<LLamaTokenData> temporary)
+        {
+            if (temporary.Length < logits.Length)
+                throw new ArgumentException("temporary memory is shorter than logits span");
+            var candidates = temporary.Slice(0, logits.Length);
+
+            var candidatesSpan = candidates.Span;
```
Lyrcaxis, line 60:
```suggestion

            // Get the memory of the logits and then its span so we can edit it.
            var candidates = temporary.Slice(0, logits.Length);
            var candidatesSpan = candidates.Span;
```

LLama/Native/LLamaTokenDataArray.cs
```diff
@@ -342,13 +362,28 @@ public struct LLamaTokenDataArrayNative
...
         /// Number of LLamaTokenData in the array
         /// </summary>
         public ulong size;
-
+        
+        /// <summary>
+        /// A pointer to an array of LlamaTokenData
+        /// </summary>
+        /// <remarks>Memory must be pinned in place for all the time this LLamaTokenDataArrayNative is in use</remarks>
+        public Span<LLamaTokenData> data
```
Lyrcaxis, line 376: I like that this is exposed! However what do you mean in the comment? Is it correct?
Might need some clarification if intended for end-users.

Lyrcaxis (approved):
Nice!

LLama/Native/LLamaTokenDataArray.cs
martindevans: I've put a slightly different comment here

LLama/Native/LLamaTokenDataArray.cs
martindevans: I've removed that comment, it's not really relevant any more.

When you could pass in a pointer directly, it was your responsibility to ensure that the pointer was fixed (i.e. `Memory.Pin()` or `fixed (blah)`) for all the time that memory was in use.

However, now that `Create` is the only way a `NativeTokenDataArray` can be created that's no longer relevant - `Create` handles it for you by pinning the memory and returning the handle.

PR #741, opened 2024-05-18
Removed .NET 7 now that it's reached EOL.

Lamothe:
Removing .NET 7 now that it is officially out of support, see https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core.


martindevans:
Thanks.

PR #749, opened 2024-05-22
ci: upload native libraries to huggingface when release new version.

AsakusaRinne:


martindevans:
I still don't understand why the binaries need to be uploaded from huggingface for autodownload to work - why can't they be pulled from this repo? That way no separate step is needed, we just need to know the git tag/commit ID to download e.g. `https://github.com/SciSharp/LLamaSharp/blob/c325ac91279428e8d88719357b93e43cd9c5ed52/LLama/runtimes/deps/avx512/llama.dll`

AsakusaRinne:
Using github is absolutely a good option and I'd like to accept such a PR if anyone want to add it as a second resource.

Here're the reasons why I chose huggingface.

1. AFAIK there's no mirror site for github with official mechanism. IIRC you're living in UK so visiting github is never a problem for you. However, people in some countries such as China, India and Iran are restricted to visiting github (and huggingface.co). Though some of them could use VPN on their PC, it might be impossible for them to use VPN in the production environment. If a developer in these countries is distributing an APP, it's even worse because most of the users with no CS background don't have a VPN. As a website built from an open-source project, huggingface has many mirror sites, such as `hf-mirror.com`. This means a lot for people in these countries. All they need to do is to set a mirror site url when using this feature.
2. File cache is significant in this feature. It's OK to do so with github as what I did in [HuggingfaceHub](https://github.com/AsakusaRinne/HuggingfaceHub). But the model downloading is also in my plan so using huggingface is obviously a big time saver for me. :)
3. Privacy: Either using huggingface or github, it's absolutely possible for users to switch to their own native libraries with a single string modification. However, I believe some people would not like to make their own native libraries public. While huggingface support downloading files with token, I didn't figure a way to download files from a private github repo with http request.

In conclusion, because of (1) and (3), I don't think we should drop the huggingface and take github instead. I'd like to discuss with you about which one to be used as default if both of them are supported. However, right now I don't have extra time to add support for it (that's why (2) is a reason!).

martindevans:
restrictions/mirroring are a good point, I'd never have thought of that!

PR #755, opened 2024-05-25
Add Rosetta2 Binaries

### LINKED ISSUE #752: LLamaSharp runtime binaries don't support Rosetta2
### Description

Hi there,

I'm attempting to integrate LLamaSharp with a C#-based game that is stuck with Rosetta emulation on M1 Macs. Unfortunately, the regular builds are compiled with AVX (I believe?) which Rosetta can't emulate, causing a crash. I compiled with minimal support from the Llama.cpp repo and it works fine. 

I'm not sure if this is something that needs to be addressed by the library itself, but felt worth noting here.

abhiaagarwal:
Closes #752. I make the assumption that "if x86 and doesn't have AVX, then it is rosetta2". My understanding is that all modern (>2015)-based Intel CPUs used on Macbooks have AVX, so checking for AVX under a x86 environment is a good enough proxy for "under Rosetta2".

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -335,7 +335,7 @@ jobs:
       - name: Rearrange Files
         run: |
           # Make all directories at once
-          mkdir --parents deps/{avx,avx2,avx512,osx-arm64,osx-x64,cu11.7.1,cu12.1.0,clblast}
+          mkdir --parents deps/{avx,avx2,avx512,osx-arm64,osx-x64,osx-x64-rosetta2,cu11.7.1,cu12.1.0,clblast}
 
           cp artifacts/llama-bin-linux-noavx-x64.so/libllama.so  deps/libllama.so
           cp artifacts/llama-bin-linux-avx-x64.so/libllama.so    deps/avx/libllama.so
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -278,6 +278,8 @@ jobs:
             defines: '-DCMAKE_OSX_ARCHITECTURES=arm64 -DLLAMA_METAL_EMBED_LIBRARY=ON'
           - build: 'x64'
             defines: '-DCMAKE_OSX_ARCHITECTURES=x86_64 -DLLAMA_METAL=OFF -DLLAMA_AVX=ON -DLLAMA_AVX2=ON'
+          - build: 'x64-rosetta2'
+            defines: '-DCMAKE_OSX_ARCHITECTURES=x86_64 -DLLAMA_METAL=OFF -DLLAMA_AVX=OFF -DLLAMA_AVX2=OFF'
     runs-on: macos-latest   
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


abhiaagarwal:
https://github.com/abhiaagarwal/LLamaSharp/actions/runs/9237231679/job/25413958437 Artifacts

SignalRT:
> https://github.com/abhiaagarwal/LLamaSharp/actions/runs/9237231679/job/25413958437 Artifacts

@abhiaagarwal, it seems that the step to Gather binaries fail in linked execution. I would appreciate If you can check this.

The other thing that seems strange to me is to name this as something related with rosetta. Any intel osx without avx support is included in this use case (including rosetta emulation). Am I right or I'm missing something?.

abhiaagarwal:
I'll fix the issue! Missing adding the folder name in `mkdir`.

> The other thing that seems strange to me is to name this as something related with rosetta. Any intel osx without avx support is included in this use case (including rosetta emulation). Am I right or I'm missing something?.

Yes, but at least in regards to Macbooks on Intel, non-AVX hardware doesn't really exist anymore. I did a search and the last chip in an Intel Macbook was in 2014. I named it Rosetta2 to make it more clear to anyone poking around *why* these binaries exist, but I don't really have a strong opinion otherwise. More than happy to rename it to `nonavx` if desired.

martindevans:
@abhiaagarwal Could you start another test run on your repo please. If that works we can merge this :)

abhiaagarwal:
@martindevans absolutely! Fixed the mkdir issue and triggered a new run here. https://github.com/abhiaagarwal/LLamaSharp/actions/runs/9294309458

abhiaagarwal:
Seems like the hardest thing I've ever done is properly creating directories... here's a run that succeeded. https://github.com/abhiaagarwal/LLamaSharp/actions/runs/9294794909

martindevans:
Looks good to me! @SignalRT I think this is ready to merge if it looks ok to you?

AsakusaRinne:
There's one more thing: could you please add the binaries to [this nuspec file](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/runtimes/build/LLamaSharp.Backend.Cpu.nuspec), so that it will get published in our next release?

Besides, @martindevans @SignalRT do you think it's a good idea to separate the MAC backend package from `LLamaSharp.Backend.Cpu`? Actually MAC binaries are not "pure" CPU ones because they may take use of metal.

martindevans:
> do you think it's a good idea to separate the MAC backend package from LLamaSharp.Backend.Cpu?

Seems like a good idea to me, it's a good way to shrink the CPU backend package down a bit.

abhiaagarwal:
If you can open a new branch on this repo, I can change this PR to merge against that? I'm not familiar with dotnet packaging, but a mac-specific backend is for sure a good idea.

SignalRT:
The PR it's ok for me. I would not separate the backend at this moment. In my opinion the point that rest is to add de libraries to the [nuspec](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/runtimes/build/LLamaSharp.Backend.Cpu.nuspec)

abhiaagarwal:
Updated the nuspec. Do y'all want me to squash merge?

SignalRT (approved):
It seems OK to me.

martindevans:
No need for you to squash, we can squash-and-merge :)

PR #757, opened 2024-05-26
LLava API Improvements

martindevans:
- Added new ways to create an image embedding which don't require a context
 - Added a way to get the embedding data directly from an image embed (one embedding vector at a time)

These are a couple of things that I've added while working on embeddings prompting support for the `BatchedExecutor`.


SignalRT:
@martindevans, It seems to be OK to me. I don't have the context of what are you doing with the BatchExecutor, but I don't see nothing extrange.

PR #751, opened 2024-05-24
June 2024 Binary Update

martindevans:
LLama.cpp commit: 1debe72737ea131cb52975da3d53ed3a835df3a6
Build Action: https://github.com/SciSharp/LLamaSharp/actions/runs/9216317481
Rosetta Build Action: https://github.com/SciSharp/LLamaSharp/actions/runs/9340131648

Testing:
- [x] Windows (CPU)
- [ ] Windows (CUDA 11)
- [x] Windows (CUDA 12)
- [x] Windows (OpenCL)
- [ ] Linux (CPU)
- [ ] Linux (CUDA 11)
- [x] Linux (CUDA 12)
- [ ] Linux (OpenCL)
- [x] MacOS (ARM64)
- [ ] MacOS (Rosetta)

PR #763, opened 2024-05-28
refactor: the directory structure.

AsakusaRinne:
Despite so many changes, this PR simply does the following things.

- Move the projects to three folders (src, test, samples) in both file system level and .NET solution level.
- Correct the file paths in the documentation and workflows.

Since we'll add the code generation (#738) and split the package into different levels, the number of projects will increase. Moving the projects to different folders will make the structure more clear. Besides, it's better to allow multiple projects for our example, making it more flexible.


martindevans:
Looks good. Only concern is this will probably break every other pending PR. Should we make an effort to get every other PR either closed or merged first?

AsakusaRinne:
Agreed. Is #751 expected to be merged soon?

martindevans:
It's had some testing on every platform, we could probably merge it right now?

PR #782, opened 2024-06-04
Remove OpenCL Backend.

martindevans:
Upstream is in the process of removing OpenCL: https://github.com/ggerganov/llama.cpp/pull/7735

Merge this only after https://github.com/ggerganov/llama.cpp/pull/7735 is merged!


martindevans:
Anyone who is currently using the OpenCL backend should consider taking the work started in https://github.com/SciSharp/LLamaSharp/pull/517 and finishing it, so we can release a Vulkan backend :)

PR #787, opened 2024-06-08
Generic Prompt Formatter

patrick-hovsepian:
- more generic default template history transformer
- EOT and EOS token to string
- minor refactor of ModelToken to readonly struct
- template convenience methods
- exposed metadata get by key native handle
- update example to use update default history transform and convenience anti prompt value from model

### DIFF
--- a/LLama.Examples/Examples/LLama3ChatSession.cs
+++ b/LLama.Examples/Examples/LLama3ChatSession.cs
@@ -1,38 +1,47 @@
-﻿using LLama.Abstractions;
-using LLama.Common;
+﻿using LLama.Common;
+using LLama.Transformers;
 
 namespace LLama.Examples.Examples;
 
-// When using chatsession, it's a common case that you want to strip the role names
-// rather than display them. This example shows how to use transforms to strip them.
+/// <summary>
+/// This sample shows a simple chatbot
+/// It's configured to use the default prompt template as provided by llama.cpp and supports
+/// models such as llama3, llama2, phi3, qwen1.5, etc.
+/// </summary>
 public class LLama3ChatSession
 {
     public static async Task Run()
     {
-        string modelPath = UserSettings.GetModelPath();
-
+        var modelPath = UserSettings.GetModelPath();
         var parameters = new ModelParams(modelPath)
         {
             Seed = 1337,
             GpuLayerCount = 10
         };
+
         using var model = LLamaWeights.LoadFromFile(parameters);
         using var context = model.CreateContext(parameters);
         var executor = new InteractiveExecutor(context);
 
         var chatHistoryJson = File.ReadAllText("Assets/chat-with-bob.json");
-        ChatHistory chatHistory = ChatHistory.FromJson(chatHistoryJson) ?? new ChatHistory();
+        var chatHistory = ChatHistory.FromJson(chatHistoryJson) ?? new ChatHistory();
 
         ChatSession session = new(executor, chatHistory);
-        session.WithHistoryTransform(new LLama3HistoryTransform());
+
+        // add the default templator. If llama.cpp doesn't support the template by default, 
+        // you'll need to write your own transformer to format the prompt correctly
+        session.WithHistoryTransform(new PromptTemplateTransformer(model, withAssistant: true)); 
+
+        // Add a transformer to eliminate printing the end of turn tokens, llama 3 specifically has an odd LF that gets printed somtimes
         session.WithOutputTransform(new LLamaTransforms.KeywordTextOutputStreamTransform(
-            new string[] { "User:", "Assistant:", "�" },
+            [model.Tokens.EndOfTurnToken!, "�"],
             redundancyLength: 5));
 
-        InferenceParams inferenceParams = new InferenceParams()
+        var inferenceParams = new InferenceParams()
         {
+            MaxTokens = -1, // keep generating tokens until the anti prompt is encountered
             Temperature = 0.6f,
-            AntiPrompts = new List<string> { "User:" }
+            AntiPrompts = [model.Tokens.EndOfTurnToken!] // model specific end of turn string
         };
 
         Console.ForegroundColor = ConsoleColor.Yellow;
@@ -40,10 +49,15 @@ public class LLama3ChatSession
 
         // show the prompt
         Console.ForegroundColor = ConsoleColor.Green;
-        string userInput = Console.ReadLine() ?? "";
+        Console.Write("User> ");
+        var userInput = Console.ReadLine() ?? "";
 
         while (userInput != "exit")
         {
+            Console.ForegroundColor = ConsoleColor.White;
+            Console.Write("Assistant> ");
+
+            // as each token (partial or whole word is streamed back) print it to the console, stream to web client, etc
             await foreach (
                 var text
                 in session.ChatAsync(
@@ -56,71 +70,8 @@ public class LLama3ChatSession
             Console.WriteLine();
 
             Console.ForegroundColor = ConsoleColor.Green;
+            Console.Write("User> ");
             userInput = Console.ReadLine() ?? "";
-
-            Console.ForegroundColor = ConsoleColor.White;
         }
-    }
-
-    class LLama3HistoryTransform : IHistoryTransform
-    {
-        /// <summary>
-        /// Convert a ChatHistory instance to plain text.
-        /// </summary>
-        /// <param name="history">The ChatHistory instance</param>
-        /// <returns></returns>
-        public string HistoryToText(ChatHistory history)
-        {
-            string res = Bos;
-            foreach (var message in history.Messages)
-            {
-                res += EncodeMessage(message);
-            }
-            res += EncodeHeader(new ChatHistory.Message(AuthorRole.Assistant, ""));
-            return res;
-        }
-
-        private string EncodeHeader(ChatHistory.Message message)
-        {
-            string res = StartHeaderId;
-            res += message.AuthorRole.ToString();
-            res += EndHeaderId;
-            res += "\n\n";
-            return res;
-        }
-
-        private string EncodeMessage(ChatHistory.Message message)
-        {
-            string res = EncodeHeader(message);
-            res += message.Content;
-            res += EndofTurn;
-            return res;
-        }
-
-        /// <summary>
-        /// Converts plain text to a ChatHistory instance.
-        /// </summary>
-        /// <param name="role">The role for the author.</param>
-        /// <param name="text">The chat history as plain text.</param>
-        /// <returns>The updated history.</returns>
-        public ChatHistory TextToHistory(AuthorRole role, string text)
-        {
-            return new ChatHistory(new ChatHistory.Message[] { new ChatHistory.Message(role, text) });
-        }
-
-        /// <summary>
-        /// Copy the transform.
-        /// </summary>
-        /// <returns></returns>
-        public IHistoryTransform Clone()
-        {
-            return new LLama3HistoryTransform();
-        }
-
-        private const string StartHeaderId = "<|start_header_id|>";
-        private const string EndHeaderId = "<|end_header_id|>";
-        private const string Bos = "<|begin_of_text|>";
-        private const string Eos = "<|end_of_text|>";
-        private const string EndofTurn = "<|eot_id|>";
     }
 }
--- a/LLama/LLamaTemplate.cs
+++ b/LLama/LLamaTemplate.cs
@@ -28,12 +28,12 @@ public sealed class LLamaTemplate
     /// <summary>
     /// Keep a cache of roles converted into bytes. Roles are very frequently re-used, so this saves converting them many times.
     /// </summary>
-    private readonly Dictionary<string, ReadOnlyMemory<byte>> _roleCache = new();
+    private readonly Dictionary<string, ReadOnlyMemory<byte>> _roleCache = [];
 
     /// <summary>
     /// Array of messages. The <see cref="Count"/> property indicates how many messages there are
     /// </summary>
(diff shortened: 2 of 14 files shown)


LLama.Examples/Examples/LLama3ChatSession.cs
```diff
@@ -1,49 +1,63 @@
...
+            [model.Tokens.EndOfTurnToken!, "�"],
             redundancyLength: 5));
 
-        InferenceParams inferenceParams = new InferenceParams()
+        var inferenceParams = new InferenceParams()
         {
+            MaxTokens = -1, // keep generating tokens until the anti prompt is encountered
             Temperature = 0.6f,
-            AntiPrompts = new List<string> { "User:" }
+            AntiPrompts = [model.Tokens.EndOfTurnToken!] // model specific end of turn string
```
martindevans, line 44: Is this actually necessary?

There's an `IsEndOfGeneration(Token)` method which _should_ be monitored in the executors and used to automatically end the inference loop. I know the stateless executors does this properly, not sure about the other executors.

LLama/LLamaTemplate.cs
```diff
@@ -189,6 +189,21 @@ public LLamaTemplate RemoveAt(int index)
 
         return this;
     }
+    
+    /// <summary>
+    /// Remove all messags from the template and resets internal state to accept/generate new messages
+    /// </summary>
+    public void RemoveAllMessages()
```
martindevans, line 196:
```suggestion
    public void Clear()
```

LLama/LLamaTemplate.cs
```diff
@@ -278,10 +291,29 @@ unsafe int ApplyInternal(Span<LLamaChatMessage> messages, byte[] output)
         }
     }
 
+    /// <summary>
+    /// Apply the template to the messages and return the resulting prompt as a string
+    /// </summary>
+    /// 
+    /// <returns>The formatted template string as defined by the model</returns>
+    public string ToModelPrompt()
```
martindevans, line 299: I deliberately avoided creating a method like this when creating LlamaTemplate originally. Calling this method will allocate a new (and potentially very large, since it contains the entire chat history) string every time it is called.

All the downstream uses of the template (e.g. tokenizing) should be built to accept a `Span<byte>` or `Span<char>`, so in principle this should never be needed. What's this used for?

martindevans: (sidenote: if we decide to keep it, it's probably better to call it `ToString()`)

LLama/Native/LLamaToken.cs
```diff
@@ -10,6 +10,9 @@ namespace LLama.Native;
 [DebuggerDisplay("{Value}")]
 public readonly record struct LLamaToken
 {
+    /// Token Value used when token is inherently null
```
martindevans, line 13:
```suggestion
    <summary>
    /// Token Value used when token is inherently null
    </summary>
```

LLama/Native/LLamaToken.cs
```diff
@@ -10,6 +10,9 @@ namespace LLama.Native;
...
+    public static readonly LLamaToken INVALID_TOKEN = -1;
```
martindevans, line 14: Correct naming for C# constants would be `InvalidToken`.

However, would it be better to replace this with a property like `IsInvalid => Value < 0`?

LLama/Native/SafeLlamaModelHandle.cs
```diff
@@ -221,11 +220,30 @@ private static int llama_model_meta_val_str_by_index(SafeLlamaModelHandle model,
...
+            var bytes = ArrayPool<byte>.Shared.Rent(bytesCount);
+
+            unsafe
+            {
+                fixed (char* keyPtr = key)
+                fixed (byte* bytesPtr = bytes)
+                fixed (byte* destPtr = dest)
+                {
+                    // Convert text into bytes
+                    Encoding.UTF8.GetBytes(keyPtr, key.Length, bytesPtr, bytes.Length);
```
martindevans, line 237:
```suggestion
                    Encoding.UTF8.GetBytes(keyPtr, key.Length, bytesPtr, byteCount);
```

The rented array will be any length, _at least_ the length you asked for. So it's not correct to use `bytes.Length`!

LLama/Native/LLamaToken.cs
martindevans: When I said that I was assuming this was used for _detecting_ invalid tokens. I see below that's not true, so just renaming it is fine :)

LLama/Native/SafeLlamaModelHandle.cs
```diff
@@ -576,13 +614,39 @@ public SafeLLamaContextHandle CreateContext(LLamaContextParams @params)
...
+            private readonly string? _eos;
 
             internal ModelTokens(SafeLlamaModelHandle model)
             {
                 _model = model;
+                _eot = LLamaTokenToString(EOT, true);
+                _eos = LLamaTokenToString(EOS, true);
+            }
+
+            private string? LLamaTokenToString(LLamaToken? token, bool isSpecialToken)
```
martindevans, line 630: Handling strings in this class feels like a bit of a mismatch to me (it's meant to just e a simple list of tokens). As far as I can see these new string properties are only used up in the Antiprompts (which may not be necessary).

If they're not critical to anything else can we remove this change from this PR for now, and discuss if we want something like this and where best to put it in another PR?

LLama/LLamaTemplate.cs
martindevans: I see below this is used in `PromptTemplateTransformer`. Could this entire method be moved there for now? That way we're not encouraging a sub-optimal use in the API (and the internals of `PromptTemplateTransformer` can be optimised if necessary later).

martindevans:
Please make sure that this compiles under netstandard 2.0. There are a few changes which I'm suspicious may break (e.g. removal of some `using LLamaSharp.Extensions`, which appear unused under NET8 but often contain polyfills for standard 2.0)

LLama/LLamaTemplate.cs
patrick-hovsepian: Good compromise, I'll move it

LLama/Native/SafeLlamaModelHandle.cs
patrick-hovsepian: fixed, thanks

LLama/Native/SafeLlamaModelHandle.cs
patrick-hovsepian: > Handling strings in this class feels like a bit of a mismatch to me (it's meant to just e a simple list of tokens). As far as I can see these new string properties are only used up in the Antiprompts (which may not be necessary).

This was my primary use-case, yes. Per your comment above, I need to explore the `IsEndOfGeneration`. Is the `InteractiveExecutor` sticking around? My motivation for adding this was because of the chat examples that used the interactive executor. I (naively) thought I could just plug in any model and it would work. I tried with Llama3 and Phi3 and got some odd behavior until I learned about chat templates. That made me think we could synthesize the history transformer to make a best guess and use the llama.cpp template formatter. To your other comments though, it's not great to have to constantly allocate memory to do the P/Invokes and I did consider if it's worth just porting the code to be dotnet native but it would stink to have to keep it in lockstep with that repo so I finally landed on the implementation in this PR. (this would have been helpful in the initial PR description 😅 )

To me, the `ModelToken` class seemed like a best fit and could cache the results to prevent further calls to the native library but I could remove it and just expose this function as a utility if that seems better

LLama.Examples/Examples/LLama3ChatSession.cs
patrick-hovsepian: > I know the stateless executors does this properly, not sure about the other executors.

Ah yeah see my last comment in regards to `LLamaTokenToString` I addressed them out of order so should have read this first. Exactly the problem I ran into is I guess it's not implemented for the InteractiveExecutor which is what the samples use.

patrick-hovsepian:
`dotnet build` does compile against all targets with the removed using statements that seemed unused. Would it be better to add a workflow that explicitly targets each framework to make it more automatic?

LLama.Examples/Examples/LLama3ChatSession.cs
martindevans: Ah ok, that's unfortunate. Does this definitely fix the problem? if so, it implies that special tokens are being rendered which seems a bit weird!

LLama/Native/SafeLlamaModelHandle.cs
martindevans: > Is the InteractiveExecutor sticking around

Probably (hopefully) not. My long term project for the past 6 months or so has been to build a ew foundation for better future executors.

> I (naively) thought I could just plug in any model and it would work

In the long term, I hope we can make that work! Many of the bits are there now - Automatic antiprompting (with end-of-generation tokens), automatic templating (with LlamaTemplate), maybe even automatic context shifting to handle out-of-context errors!

> To me, the ModelToken class seemed like a best fit and could cache the results to prevent further calls to the native library but I could remove it and just expose this function as a utility if that seems better

I'm ok with either way to be honest.

If we're going to keep it with the strings in `ModelTokens` could you switch ModelToken back to a `sealed class` instead of a struct - it's a bit too big to be a struct (and will probably grow if we add equivalent stirng properties for all the special tokens.

martindevans:
> a workflow that explicitly targets each framework to make it more automatic?

It would, and I thought we had one to be honest. I double checked before making that comment and was surprised to see we didn't!

LLama.Examples/Examples/LLama3ChatSession.cs
patrick-hovsepian: The rendering itself can be fixed by specifying the token as a keyword for `LLamaTransforms.KeywordTextOutputStreamTransform` 

I've tested these changes with llama3, phi3, and qwen 1.5 and it seems to work as expected for those

LLama/Native/SafeLlamaModelHandle.cs
patrick-hovsepian: > Probably (hopefully) not. My long term project for the past 6 months or so has been to build a ew foundation for better future executors.

Is that going to be here or another project entirely? I'd be interested in approaching this from the ground up if you're open to it 😸 

I'll revert the `ModelTokens` back to a sealed class and leave it as-is for now if that's cool

martindevans: Here. Over the past six months I've been laying the groundwork with `BatchedExecutor` (a low level executor that can act as a foundation for all executors).

I wrote up a lot of my thoughts recent in this comment [this comment](https://github.com/SciSharp/LLamaSharp/pull/740#issuecomment-2117693956). I ended up giving an overview of how I see the project being split up into layers. In particular the **Mid level** and **Puzzle Pieces** section show how I've put together a lot of the parts for new high level executors, now just need to figure out how to put them together! I'd be very happy with help there!

patrick-hovsepian: Oh cool, I'll give that thread a read to get caught up

LLama/LLamaTemplate.cs
```diff
@@ -101,6 +104,11 @@ public bool AddAssistant
             }
         }
     }
+    
+    /// <summary>
+    /// Get a span to the underlying bytes as specified by Encoding
+    /// </summary>
+    public ReadOnlySpan<byte> TemplateDataBuffer => _result;
```
martindevans, line 111: The reason I created `Apply` with the slightly weird use pattern (call it to do work and discover length, call it again to get results) was to ensure the internal `_result` array was never exposed.

This is to avoid a case like this where the buffer might get modified underneath you without you realising:

```csharp
t.Apply()
var template1 = t.TemplateDataBuffer;
t.RemoveAt(1);
t.Apply()
var template2 = t.TemplateDataBuffer;
// template1 was modified by the second call to Apply!
```

If we don't want to accept that risk, then this property needs to be removed for that reason.

On the other hand, if we _do_ want to accept that risk then we could change `Apply()` to return the span immediately, that leads to a much nicer interface!

```csharp
ReadOnlySpan<byte> Apply()
```

It's up to you which you go with. Having seen it in use I think I'd be slightly in favour of option 2 (as long as it is there is a warning in the method docs comments of course).

---

Sidenote: there's also a bug here, although from the above it probably doesn't matter!

`_result` is a big array that contains the template output. However, it's not _all_ valid data. `_resultLength` indicates how much of the array currently contains useful data. So this should be `_result.AsSpan(0, _resultLength);`.

patrick-hovsepian: I'm in favor of option 2 as well. I'd argue the internal buffer shouldn't be a class member at all. Every call to Apply would be idempotent and just return a new buffer and however callers chose to share or pass around that data is now up to them. I'd also be in favor of just returning the string here as is and not have to force them to also do an encoding translation or at the least an overload that does such

patrick-hovsepian: Sounds good, made the changes to support option 2.

I guess my gap is the use cases I've seen it's always been applied the last message added (sans the system prompt and user message seeding) likely because of the `InteractiveExecutor`. We'll likely be able to make additional improvements here since the history would be additive so re-templating the entire history on every call seems like overkill. Something that can be explored later.

martindevans: At the moment the templater relies entirely on llama.cpp which just supplies one method that takes a list and applies the template to it, so any kind of incremental updating is difficult with that particular API. I've been wondering about other upstreaming some improvements to llama.cpp to make it incremental, but that might a bit beynd my C++ abilities.

martindevans:
Looks good to me, thanks for all the hard work @patrick-hovsepian! I'll merge it once the CI finishes :)

PR #789, opened 2024-06-11
Model File Manager

patrick-hovsepian:
Create a simple manager to make administration of local models simpler
- Configure a list of directories that will be scanned to find gguf files that are ready to be loaded
- convenience method for loading a model that includes default params
- tests


LLama/Model/ModelManager.cs
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /// <summary>
+    /// Updates the passed in model params with default values
+    /// </summary>
+    /// <param name="modelParams"></param>
+    public static void DefaultModelConfigurator(ModelParams modelParams)
+    {
+        // Reasonable defaults
+        modelParams.GpuLayerCount = 12;
```
martindevans, line 276: `ModelParams` already has default values for all properties which are meant to be reasonable defaults. I think it'd be better to use those values.

LLama/Model/ModelManager.cs
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /// <inheritdoc />
+    public IEnumerable<LLamaWeights> GetLoadedModels()
+    {
+        return _loadedModelCache.Values;
+    }
+
+    /// <inheritdoc />
+    public bool TryGetLoadedModel(string modelId, out LLamaWeights model)
```
martindevans, line 243: One potential failure case I can see here is if someone were to retrieve weights from the `ModelManager` and then Dispose them. The cache would then contain disposed weights (which is fine) but then would return a disposed `LLamaWeights` object next time (which is not fine).

I think you probably need to add a check when you retrieve something from the `_loadedModelCache`, if the object has been disposed remove it from the cache and proceed from there as if it was never in the cache at all.

LLama/Model/ModelManager.cs
```diff
@@ -0,0 +1,309 @@
...
+        if (TryGetLoadedModel(modelId, out var model))
+        {
+            model.Dispose();
+            return _loadedModelCache.Remove(modelId);
+        }
+        return false;
+    }
+
+    /// <inheritdoc />
+    public void UnloadAllModels()
```
martindevans, line 301: This entire class should probably be `IDisposable` since it owns all of the `LLamaWeights` objects (with the Dispose implementation calling `UnloadAllModels()`

martindevans (changes requested):
I've left a few comments, the main issue is about being completely clear about ownership and resource disposal (models are expensive to load and keep loaded, so it's important that resource management is completely clear!)

LLama/Model/ModelManager.cs
patrick-hovsepian: I internally debated on this a bit and ultimately decided against it and figured the caller should be responsible for the model and dispose of it themselves or call unload explicitly. Do you feel strongly about this class being disposable itself? 

Main use case would be something like 
`using var model ModelCache.LoadModel(...);`

where they wouldn't call Unload explicitly. To your point above though, `TryGetLoadedModel` should guard against externally disposed models. and same with the call to load, if it's already in the cache.

LLama/Model/ModelManager.cs
```diff
@@ -0,0 +1,309 @@
...
+
+    /// <summary>
+    /// Currently loaded models
+    /// </summary>
+    /// <returns></returns>
+    public IEnumerable<LLamaWeights> GetLoadedModels();
+}
+
+/// <inheritdoc />
+public class ModelManager : IModelManager
```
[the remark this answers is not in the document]
patrick-hovsepian: sure, changed

LLama/Model/ModelManager.cs
patrick-hovsepian: good catch, addressed

LLama/Model/ModelManager.cs
martindevans: > I internally debated on this a bit and ultimately decided against it and figured the caller should be responsible for the model

I think the important thing is to be very clear about resource management, that's probably the most important problem that LLamaSharp has to solve in wrapping an unmanaged set of resources up for use in a managed language!

The problem with the cache not having ownership of the weights becomes _very_ unclear. If I call `LoadModel` from some code then it might be reasonable to assume I own those weights and should dispose them when done... but then the entire concept of a shared cache is useless (I unloaded the weights, so no one else can use them) or even harmful (I dispose the shared weights while some other bit of code is using them).

---

While typing up the above I did have another idea:

The lowest level thing that holds ownership of any resource is always a `SafeHandle` (e.g. [`SafeLlamaModelHandle`](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Native/SafeLlamaModelHandle.cs) for weights. A `SafeHandle` is a **reference counted** owner of a resource, caling dispose on a safe handle won't actually dispose the underlying resource until the count goes to zero.

We use this in `SafeLLamaContextHandle`. The context adds 1 to the weights ref count until the context is disposed. That way this code actually works:

```csharp
var weights = LoadWeightsFromSomewhere();
var ctx = weights.CreateContext();
weights.Dispose();
ctx.DoSomething(); // <-- this is fine, even thought it references the weights we just "disposed"
ctx.Dispose(); // <-- _now_ the weights are actually disposed
```

We could do the same thing in `LLamaWeights`. The cache can internally contain `SafeLLamaModelHandle` objects (which it owns, and disposes when the cache is disposed). When a request is made for some weights they get wrapped up in a new `LLamaWeights` objects (which increments the count), the caller unambiguously owns this object.

Note: if this is all sounds a bit scary and you'd rather not implement it I understand, I'm happy to have a look at it in a separate PR which we can merge first and then rebase this one onto.

patrick-hovsepian: Doesn't sound scary, let me explore a bit 😄

martindevans: LLamaContext should serve as a good example of how to do it

patrick-hovsepian: Alright, let me know what you think of the latest changes

LLama/Model/ModelCache.cs
```diff
@@ -0,0 +1,238 @@
...
+    {
+        var isCached = _loadedModelCache.TryGetValue(modelId, out var handle);
+        model = isCached
+            ? LLamaWeights.FromSafeModelHandle(handle)
+            : null!;
+        return isCached;
+    }
+
+    /// <inheritdoc />
+    public async Task<LLamaWeights> LoadModel(string modelPath,
```
AsakusaRinne, line 138: Seems `LoadModelAsync` will match the current namings better.

LLama/Model/ModelCache.cs
```diff
@@ -0,0 +1,238 @@
...
+            List<ModelFileMetadata> directoryModelFiles = [];
+            foreach (var file in Directory.EnumerateFiles(fullDirectoryPath))
+            {
+                if (!ExpectedModelFileTypes.Contains(Path.GetExtension(file)))
+                {
+                    continue;
+                }
+
+                // expected model file
+                var fi = new FileInfo(file);
```
AsakusaRinne, line 77: Please check whether this file is a symbolic link. If so, we should get the byte size from its target file.

If I'm not misunderstanding it, the semantic here is the real model size?

LLama/Model/IModelCache.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// Configured set of directories that are scanned to find local models
+    /// </summary>
+    /// <value></value>
+    public IEnumerable<string> ModelDirectories { get; }
+
+    /// <summary>
+    /// Add a directory containing model files
+    /// </summary>
+    /// <param name="directory"></param>
+    public void AddDirectory(string directory);
```
AsakusaRinne, line 25: I have a few concerns about the abstractions here. As this API appears in `IModelCache`, it is required to be implemented in all its derived classes. However, they are likely to have similar implementations about this method, which may lead to duplicated code in the future. Some other APIs such as `RemoveDirectory`, `RemoveAllDirectories`, etc. are also in this case.

What about adding an abstract class that provides default implementations for these APIs? `ModelCache` -> `ModelCacheBase` -> `IModelCache`. The method `GetModelsFromDirectories`could also be placed in the base class.

On my side, a different implementations of this interface might be needed when we want to support remote file (for example, huggingface).

AsakusaRinne (changes requested):
Thank you for your contribution! I left some comments and please let me know if any help is needed.

LLama/Model/IModelCache.cs
martindevans: That's a good point. To keep it as a minimal as possible the `IModelCache` interface should probably just have:
 - Get a model
 - Get list of all loaded models
 - Get available models (maybe)

LLama/LLamaWeights.cs
```diff
@@ -71,6 +71,23 @@ private LLamaWeights(SafeLlamaModelHandle weights)
...
+        /// <param name="handle"></param>
+        /// <returns></returns>
+        public static LLamaWeights FromSafeModelHandle(SafeLlamaModelHandle handle)
+        {
+            var model = new LLamaWeights(handle);
+
+            // Increment the model reference count while this weight exists.
+            // DangerousAddRef throws if it fails, so there is no need to check "success"
+            var success = false;
+            handle.DangerousAddRef(ref success);
```
martindevans, line 86: I don't see where this refcount is _decremented_?

Assuming I'm not missing it, the LLamaWeights class needs:
 - decrement it on `Dispose()` **if** this instance owns the safe handle (i.e. if it was created with this method).
 - finalizer to ensure that dispose gets called even if the object is not disposed.

LLama/LLamaWeights.cs
```diff
@@ -71,6 +71,23 @@ private LLamaWeights(SafeLlamaModelHandle weights)
             Metadata = weights.ReadMetadata();
         }
 
+        /// <summary>
+        /// Create from a "shared" handle
+        /// </summary>
+        /// <param name="handle"></param>
+        /// <returns></returns>
+        public static LLamaWeights FromSafeModelHandle(SafeLlamaModelHandle handle)
```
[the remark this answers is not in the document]
martindevans: For some reason GitHub won't let me turn that code block into a suggestion :/

LLama/LLamaWeights.cs
```diff
@@ -71,6 +71,23 @@ private LLamaWeights(SafeLlamaModelHandle weights)
...
+        {
+            var model = new LLamaWeights(handle);
```
martindevans, line 81: Rather than exposing this as a static constructor, would it make sense to expose it as an instance method on `LLamaWeights`? Something like `CreateShared`, which can be used to effectively clone the `LLamaWeights` object to a new one with shared ownership of the underlying `SafeLlamaModelHandle`?

LLama/Model/IModelCache.cs
```diff
@@ -60,7 +60,8 @@ public interface IModelCache : IDisposable
 
     // Model Load and Unload
     /// <summary>
-    /// Load a model file to be used for infernce
+    /// Load a model file to be used for inference
+    /// The caller assumes responsible for disposing this model
```
martindevans, line 64:
```suggestion
    /// Load a model file to be used for inference.
    /// The caller assumes responsibility for disposing this model.
```

LLama/Model/IModelCache.cs
```diff
@@ -86,15 +87,10 @@ public interface IModelCache : IDisposable
 
     /// <summary>
     /// Attempt to get a model that's expected to be loaded
+    /// The callers assumes responsiblilty for the lifetime of the model at this point if it exists in the cache
     /// </summary>
     /// <param name="modeId"></param>
     /// <param name="model"></param>
     /// <returns></returns>
```
martindevans, line 94:
```suggestion
    /// <summary>
    /// Attempt to get a model that's has already been loaded. The caller assumes responsibility for disposing the returned model.
    /// </summary>
    /// <param name="modeId">The modelId which was passed into <see cref="LoadModel"/></param>
    /// <param name="model">The model, if it was already loaded, otherwise null</param>
    /// <returns>true, if the model was available in the cache. Otherwise false.</returns>
```

LLama/Model/IModelCache.cs
```diff
@@ -86,15 +87,10 @@ public interface IModelCache : IDisposable
 
     /// <summary>
     /// Attempt to get a model that's expected to be loaded
+    /// The callers assumes responsiblilty for the lifetime of the model at this point if it exists in the cache
     /// </summary>
     /// <param name="modeId"></param>
     /// <param name="model"></param>
     /// <returns></returns>
     public bool TryGetLoadedModel(string modeId, out LLamaWeights model);
```
martindevans, line 95:
```suggestion
public bool TryGetLoadedModel(string modeId, out LLamaWeights? model);
```

LLama/LLamaWeights.cs
martindevans: Note: if you did this then the cache could go back to storing `LLamaWeights` directly. Every item returned from the cache would just call `CreateShared` on the item in the cache.

LLama/Model/ModelCache.cs
patrick-hovsepian: Checking for symbolic links would be good but it's non trivial to handle it cross platform. Can this be deferred and addressed separately? 

It's just a reflection of the size of the actual data on disk. I'm not sure if that maps directly to the model's underlying size or not. I think a future enhancement would be to read part(s) of GGUF header to get some more metadata about the model like name, architecture, etc but I kept it to just the file system info for now.

LLama/LLamaWeights.cs
patrick-hovsepian: It's done as part of `UnloadModel` the reason for the split is, I didn't want to change existing semantics of code already using this. Although, now as I'm typing this out, as long as it gets aded to both disposes and constructors, it should be okay? I think this method is still needed though to create from a cached handle

LLama/Model/ModelCache.cs
AsakusaRinne: It's okay to keep it now but `FileInfo.LinkTarget` is a cross-platform API to get the target file path with .NET version > NetStandard2.0

LLama/Model/IModelCache.cs
patrick-hovsepian: You both make good points, let me tweak this a bit

LLama/Model/ModelCache.cs
patrick-hovsepian: yeah, I found that too and didnt want to deal with 2.0 case 😄

LLama/LLamaWeights.cs
martindevans: I think it'd be neater to change the existing semantics. Also it's a more general improvement, if any `LlamaWeights` (whether or not it's from the cache) can be "cloned" to another object with a shared ref-countered ownership.

LLama/Model/ModelCache.cs
AsakusaRinne:
```suggestion
                var fi = new FileInfo(file);
                // TODO: check whether the file is a symbolic link
```

OK, please leave a comment here.

LLama/Model/IModelCache.cs
patrick-hovsepian: I still need to finalize the shared semantics a bit. I reworked the interface(s) a bit. Let me know what you think of the separation

LLama/LLamaWeights.cs
```diff
@@ -16,12 +16,21 @@ namespace LLama
     public sealed class LLamaWeights
         : IDisposable
     {
+        private bool _disposed = false;
+
+        ///
```
martindevans, line 21:
```suggestion
        /// <inheritdoc />
```

LLama/LLamaWeights.cs
```diff
@@ -204,11 +213,33 @@ public static async Task<LLamaWeights> LoadFromFileAsync(IModelParams @params, C
...
+        internal void Dispose(bool disposing)
+        {
+            if (_disposed)
+            {
+                return;
+            }
+
+            if (disposing)
+            {
+                NativeHandle.DangerousRelease();
```
martindevans, line 238: I think `DangerousRelease` needs to happen, whether or not this being called by the finalizer.

Example:
1. Allocate a SafeHandle
2. Wrap in LLamaWeights (increment count)
3. Drop LLamaWeights without properly calling dispose
4. `DangerousRelease` never gets called, so refcount never goes to zero!

LLama/LLamaWeights.cs
```diff
@@ -64,28 +73,28 @@ public sealed class LLamaWeights
...
+            NativeHandle.DangerousAddRef(ref success);
         }
 
+        #region Load
         /// <summary>
-        /// Create from a "shared" handle
+        /// Create from a "shared" handle. The `SafeLlamaModelHandle` will not be disposed and the model will not be unloaded until <b>all</b> such handles have been disposed.
         /// </summary>
         /// <param name="handle"></param>
         /// <returns></returns>
```
martindevans, line 94: Can this be made into a `Clone()` method (not sure about that name, open to suggestions). Basically an instance method on `LLamaWeights` which creates a new `LLamaWeights` that shares the same `NativeHandle` and `Metadata` fields. 

Alternatively, it could just be removed altogether, since it's now just a call to the constructor. However adding a clone method would be nice since it skips reading the metadata again.

LLama/LLamaWeights.cs
patrick-hovsepian: updated

LLama/LLamaWeights.cs
patrick-hovsepian: created clone method

LLama/Model/HuggingFaceModelRepo.cs
patrick-hovsepian, line 1: essentially for demo purposes. wanted to see how abstract the interface is

LLama/Model/IModelCache.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// Unload and dispose of a model with the given id
+    /// </summary>
+    /// <param name="modelId"></param>
+    /// <returns></returns>
+    public bool UnloadModel(string modelId);
+
+    /// <summary>
+    /// Unload all currently loaded models
+    /// </summary>
+    public void UnloadAllModels();
```
patrick-hovsepian, line 44: There's a few challenges with supporting this and I'm thinking maybe it's better to just drop these methods and have the caller explicitly call `Dispose` on the weight? 

A few questions
1. Should multiple instances of the same model be allowed to be loaded? I think yes. If so, we'll need a way to ensure calling unload is specific to a model.
2. Should we force model aliases in the cache to be unique? This would help the issue where I load multiple instances of the model and then calling dispose is guaranteed to operate on that same model. If we don't want to enforce this restriction, unloading the correct model because tricker and might require the original model to be passed in as opposed to the alias. 
3. Are we better off getting rid of this class altogether? My main goal was to have something like `IModelRepo` and quickly went out of scope with some of this but if we think it'll be useful I'm happy to leave it but I'm struggling a bit to justify it especially given the unload/dispose challenges.

LLama/LLamaWeights.cs
```diff
@@ -86,15 +81,43 @@ private LLamaWeights(SafeLlamaModelHandle handle)
...
+        /// Create a new instance of the model using same NativeHandle as this model.
+        /// Metadata is also copied from the existing model rather than read from the hanlde directly
+        /// The `SafeLlamaModelHandle` will not be disposed and the model will not be unloaded until <b>ALL</b> such handles have been disposed.
+        /// </summary>
         /// <returns></returns>
-        public static LLamaWeights FromSafeModelHandle(SafeLlamaModelHandle handle)
+        public LLamaWeights CloneFromHandleWithMetadata()
         {
-            return new LLamaWeights(handle);
+            var metadataClone = Metadata
```
martindevans, line 117: I don't think there's any need to clone `Metadata` here - it's an `IReadOnlyDictionary<string, string>` so it's already immutable (sorry if that was unclear before).

LLama/Model/HuggingFaceModelRepo.cs
martindevans: If this is just a test do we want to keep it in this PR? I think @AsakusaRinne was working on HF integrations for model loading, so you might want to check what the status of that work is and add something in a separate PR?

LLama/Model/ModelCache.cs
```diff
@@ -10,25 +10,35 @@
 
 namespace LLama.Model;
 
+internal class CachedModelReference
```
martindevans, line 13: Since the underlying `SafeHandle` already has a refcount is this needed? It looks like it's only used in the `UnloadModel`, but I'm not quite sure what for, so I might just be missing something about the lifetime management.

LLama/LLamaWeights.cs
martindevans: This has been done with the clone from metadata method

LLama/Model/IModelCache.cs
martindevans: > Should multiple instances of the same model be allowed to be loaded? I think yes. If so, we'll need a way to ensure calling unload is specific to a model.

Definitely. You can load a model with different settings (e.g. loras) which affect the output almost as if it's another model.

> Should we force model aliases in the cache to be unique?

Do you mean the `modelId` string? If so then I would think loading a model with an ID which is in the cache would either:
 - throw an error
 - return the existing model

I'd lean towards throwign the error out of those choices.

> Are we better off getting rid of this class altogether?

The model repo idea is interesting, and I think it's something Rinne has looked at as well. But yeah maybe it would be better to remove it from this PR, and make this PR all about the new shared lifetimes (which I think is a pretty big improvement in itself). The model repo can be done in a followup PR, fully taking advantage of the new lifetimes.

LLama/Model/ModelCache.cs
patrick-hovsepian: removed per discussion above

LLama/LLamaWeights.cs
patrick-hovsepian: reference used

LLama/Model/HuggingFaceModelRepo.cs
patrick-hovsepian: Curious what her thoughts are on that

LLama/Model/IModelCache.cs
patrick-hovsepian: If anything, I'd rather keep the model repo and the file system class than this but I've further reworked this to be more explicit. I haven't been able to finalize what a "good" api is for something like this (name aside) but the discussion here has helped add context.

LLama/Native/SafeLlamaModelHandle.cs
```diff
@@ -102,6 +110,7 @@ protected override bool ReleaseHandle()
             return true;
         }
 
+        // TODO: Move this to the model manager?
```
martindevans, line 113: todo needs removing

patrick-hovsepian: removed

LLama/Model/IModelSourceRepo.cs
patrick-hovsepian, line 1: have any thoughts on this @AsakusaRinne

LLama/Model/HuggingFaceModelRepo.cs
AsakusaRinne: Sorry for the late reply, I agree that it could be removed from this PR and added in a separate PR to `LLama.Experimental` project. With [HuggingfaceHub](https://github.com/AsakusaRinne/HuggingfaceHub), it's easy to download a model from the huggingface. It's easy to implement a remote model manager for gguf files but the APIs might change in the future if we want to support other formats (.safetensors, .bin) based on [GGMLSharp](https://github.com/SciSharp/GGMLSharp). So I would recommend putting it in the `LLama.Experimental` first.

AsakusaRinne (commented):
The overall of the code looks good. Since this PR added many new APIs, could you please add a document to introduce the usage of model cache and model repo? Please refer to #747 to see how to add a doc. :)

PR #795, opened 2024-06-18
Add the github build actions for Vulkan

martindevans:
See #794

PR #797, opened 2024-06-19
Vulkan support

m0nsky:
Based on #517 this PR adds support for the Vulkan backend. The codebase for the native library had changed quite a bit since that setup from february and I've rewritten the Vulkan API detection which now uses a regex to extract the Vulkan API version from `vulkaninfo --summary`. Tested on both Windows 10 and Ubuntu 22.04.

- On Windows, when Vulkan is available, it will be used by default (unless CUDA is available)
- On Linux, it requires `vulkan-tools` to be installed (`sudo apt install vulkan-tools`)

Both backends can be configured in the `NativeLibraryConfig` just like before:

```cs
NativeLibraryConfig
   .All
   .WithCuda(false)
   .WithVulkan(true)
```

I have tested the examples with both CUDA and Vulkan and it seems to work correctly. 

vulkan_example_small

**Note:**
Keep in mind that with llama.cpp commit `1debe72737ea131cb52975da3d53ed3a835df3a6` (which the LLamaSharp June 2024 binary update is based off) Vulkan currently crashes on multi-gpu setups (like mine) when `SplitMode` is set to `GPUSplitMode.None` which is the default. Setting this to `GPUSplitMode.Layer` works correctly. This has since then been fixed in [this PR](https://github.com/ggerganov/llama.cpp/pull/7552).

Submitting as draft for now as it depends on #795 to be merged first.


martindevans:
Note: #795 has now been merged.

m0nsky:
Now that both #795 and #799 have been merged, I have done some additional fixes/cleanups to the Vulkan build process. Because the `llama_cpp_commit` wasn't being taken in account, the binaries were causing a crash on a fresh build.

I've ran the updated build action on my local fork, and tested both llama and llava using Vulkan, and both are working as they should.

Should I remove the draft status for this PR?

martindevans:
> Should I remove the draft status for this PR?

Presumably this can't be merged until a binary update? If so we should probably leave it as a draft until I start that process.

m0nsky:
Yup you're right. We do need the corrected `llama_cpp_commit` for the Vulkan build before the binary update, so I will split that up into a separate PR.

martindevans (approved):

PR #799, opened 2024-06-19
LLaVa Vulkan build actions

m0nsky:
This PR adds the Vulkan build actions for `llava_shared.dll` and `libllava_shared.so`.
I ran a `Update Binaries` test run in my fork, which completed succesfully.


martindevans:
Looks good to me, thanks!

PR #800, opened 2024-06-19
Update CUDA 12 paths from cu12.1.0 to cu12.2.0

m0nsky:
I noticed the CUDA 12 paths (in `Update Binaries`) were recently changed from `cu12.1.0` to `cu12.2.0`. This means that starting from the next binary update, the CUDA 12 deps will be built to the `cu12.2.0` folder. However, `LLamaSharp.Runtime.targets` and `LLamaSharp.Backend.Cuda12.nuspec` are still looking for the `cu12.1.0` folder.

This PR updates those paths, and should be merged after the next binary update (because currently, the `cu12.2.0` folder is not yet shipped, so merging now will result in errors when building a fresh project)


martindevans:
Merging this into the july-2024-binaries branch, **not** master!

PR #830, opened 2024-07-06
Remove .NET 7.

Lamothe:
We shouldn't be using .NET 7 as it is now out of support.


martindevans (approved):

martindevans:
Looks good to me, thanks for this. 

I'll let @AsakusaRinne make the final decision, since they work on the `Experimental` project.

PR #836, opened 2024-07-10
Fix possible file locks due to race conditions during build & test

m0nsky:
Fixes an edge case where downloading the deps/models were sometimes causing file locks due to race conditions when building and testing the project. After applying the fixes, I have not been able to reproduce the issue.

It affects both the build targets for `LLamaSharp` (downloading the deps) and `LLama.Unittest` (downloading the models).

Note: this is based on, and will merge (after more testing) into `july-2024-binaries`


martindevans:
Tested in Visual Studio and `dotnet build` CLI with a clean clone, it works!

PR #842, opened 2024-07-10
Manually load GGML dependency for Linux/OSX

m0nsky:
After the introduction of the GGML dependency in the July 2024 binary update, the tests for Linux/OSX were failing because the NativeLibrary failed to load the GGML dependency.

There were two ways to tackle the issue:
- Specify `UseDllDirectoryForDependencies` in `NativeLibrary.TryLoad`, however, this does not seem to help on Linux / OSX
- Set the `LD_LIBRARY_PATH` environment variable to the current runtime directory, however, the `LD_LIBRARY_PATH` environment variable has to be set before the application starts, making it unusable in most cases, and incompatible with attempting to load multiple libraries

This PR introduces manually resolving the `GGML` dependency in `TryLoadLibrary` before loading the main library, for Linux & OSX.

I have ran the unit tests locally on Ubuntu 22.04 and they are now passing, and have also confirmed that building a standalone application where the `.so` files are placed in the same directory as the executable still work correctly.

I haven't ran the tests for OSX because I don't have access to it, but I hope those tests will also pass now, as they were reporting the same error. Maybe we can do this through GitHub?

Note: this is based on, and will merge into `july-2024-binaries`

### DIFF
--- a/LLama/Native/Load/NativeLibraryUtils.cs
+++ b/LLama/Native/Load/NativeLibraryUtils.cs
@@ -41,46 +41,37 @@ namespace LLama.Native
             {
                 // Prepare the local library file and get the path.
                 var paths = library.Prepare(systemInfo, config.LogCallback);
+                
                 foreach (var path in paths)
                 {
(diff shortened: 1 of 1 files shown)


martindevans:
Thank you for all your hard work on this!

PR #844, opened 2024-07-11
Use the correct lib prefix for GGML dependency on Linux / OSX

m0nsky:
Updates the `NativeLibraryUtils` to use the correct lib prefix (from `GetPlatformPathParts`) for loading the GGML dependency on Linux / OSX.

Tests are passing locally.


martindevans (approved):

martindevans:
Thanks!

PR #848, opened 2024-07-12
CI: make unzip target depend on download target

m0nsky:
Attempt to fix CI randomly failing due to "because the file does not exist or is inaccessible" when trying to unzip the deps from `LLamaBinaries`.

Microsoft [mentions](https://learn.microsoft.com/en-us/visualstudio/msbuild/target-build-order?view=vs-2022) the following for the target build order:

`In general, you shouldn't depend on the declaration order to specify what tasks run before other tasks.`

This PR splits up the download & unzip targets and makes the unzip target depend on the download target, so that they can never be executed in the wrong order, and adds download retries so that the build does not fail if there is a temporary issue when downloading the file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -352,3 +352,4 @@ site/
 **/appsettings.Local.json
 /LLama/runtimes/deps
 /LLama/runtimes/deps.zip
+/LLama/runtimes/version.txt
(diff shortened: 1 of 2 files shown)


martindevans:
Looks like this is still failing due to the zip file not existing.

martindevans:
Thankyou for all the hard work debugging this!

PR #850, opened 2024-07-12
Align ubuntu versions across workflow

m0nsky:
This PR aligns all Ubuntu versions in the compilation workflow by updating the `20.04` versions to `22.04` as previously both versions were mixed throughout the file.

Since the binaries are now distributed through SciSharp/LLamaSharpBinaries, I think this should go there instead of the main repo (?). I ran a `Update Binaries` action for this PR over [here](https://github.com/m0nsky/LLamaSharp/actions/runs/9898317234) which succeeded.

Submitting as a draft for now because:
- The change is not necessary
- I'm wondering if this raises the minimum OS for the end user to 22.04, or if this is not related at all

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -34,7 +34,7 @@ jobs:
             defines: '-DGGML_AVX2=OFF'
           - build: 'avx512'
             defines: '-DGGML_AVX512=ON'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


martindevans:
In https://github.com/SciSharp/LLamaSharp/pull/297 we specifically moved _back_ to 20.04, which fixed an issue with older versions of GLIBC in docker. I don't know enough about docker/Linux to say if this is still important though.

m0nsky:
Thanks, that's what I was wondering. The initial Vulkan PR targeted 22.04, so I kept that in the new Vulkan PR. That might cause the original docker compatibility issue to pop back up in the Vulkan backend, but for now I think I'll close this PR until it is confirmed. What do you think?

martindevans:
I think it's best to close it for now yeah. I'm happy to wait until someone reports a similar issue with Vulkan before changing the target there.

PR #851, opened 2024-07-12
Include debug symbols with NuGet packages

dpmm99:
Just adding symbols [per the docs](https://learn.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg) so users can step into the code more reliably.


martindevans:
Thanks!

Issue #856, opened 2024-07-18
[BUG]: Tokenization in 0.14.0 adds spaces


newsletternewsletter:
### Description

When tokenizing a text and decoding these tokens, one can see that tokenization now (as of version 0.14.0) adds one additional starting space to `text` for every call of `Context.Tokenize(text, addBos, special)`. This is especially bad if a text is tokenized with more than one call.
Version 0.13.0 did not exhibit such behavior. Or at least, it did not add spaces at the start of words, changing their token ids.

This seems fine for most models (I saw this when using `trollek/NinjaMouse-2.4B-32L-danube`), but when I use gemma-1.1-2b-it-Q6_K.gguf (from: bartowski/gemma-1.1-2b-it-GGUF) now, it's not working anymore. The prompt was:
```
<start_of_turn>user
Who are you?<end_of_turn>
<start_of_turn>model

```

Validating with `tokenize` from llama.cpp b2985 (used in LlamaSharp Version 0.13.0):
```
     2 -> '<bos>'
   106 -> '<start_of_turn>'
  2425 -> ' user'
235286 -> '\'
235254 -> 'n'
  6571 -> 'Who'
   708 -> ' are'
   692 -> ' you'
235336 -> '?'
   107 -> '<end_of_turn>'
   730 -> ' \'
235254 -> 'n'
   106 -> '<start_of_turn>'
  2091 -> ' model'
235286 -> '\'
235254 -> 'n'
```
Interestingly the token at position 2 with id 2425 `' user'` adds a starting space to `'user'` (id 1645).

But even the latest llama.cpp b3412 does not work correctly, look at token at position 2 with id 968 `' <'` :
```
     2 -> '<bos>'
   968 -> ' <'
  2997 -> 'start'
235298 -> '_'
   559 -> 'of'
235298 -> '_'
 15508 -> 'turn'
235313 -> '>'
  1645 -> 'user'
   108 -> '
'
  6571 -> 'Who'
   708 -> ' are'
   692 -> ' you'
181537 -> '?<'
   615 -> 'end'
235298 -> '_'
   559 -> 'of'
235298 -> '_'
 15508 -> 'turn'
235313 -> '>'
   108 -> '
'
235322 -> '<'
  2997 -> 'start'
235298 -> '_'
   559 -> 'of'
235298 -> '_'
 15508 -> 'turn'
235313 -> '>'
  2516 -> 'model'
   108 -> '
'
```

[...]

PR #862, opened 2024-07-22
Microsoft.KernelMemory version 0.68+ compatibility fix

SpaceAntelope:
fixes 859

### Issue details
The latest version of Microsoft.KernelMemory (0.68.240716.1 in  my case) adds **IReadOnlyList<string> GetTokens(string)** to interface **Microsoft.KernelMemory.AI.ITextTokenizer**

This breaks any project that would reference the latest packages of **LlamaSharp.kernel-memory** and **Microsoft.KernelMemory.Core** together, affecting mostly developers just getting into LLamaSharp.

### How it's solved in this commit
This commit provides a tentative implementation using **LLamaContext.Tokenizer** to get the tokens in embedding form and **StreamingTokenDecoder** to turn them back into (parts of) words and return them.

My assumptions for the overall expected behavior are based on the implementation of CountTokens in LLamaSharpTextEmbedingsGenerator and LLamaSharpTextGenerator, This means that it breaks on null input and returns an empty token that corresponds to the BOS embedding. Unit tests also check that the result of CountTokens matches the actual count of the tokens return from GetTokens.

### Other considerations
In the unit tests I trim the 'actual' result to match the 'expected' to account for the added empty space that corresponds to the BOS token. Issues such as https://github.com/SciSharp/LLamaSharp/issues/856 indicate that further clarity will emerge with respect to how this should be properly handled.

### DIFF
--- a/LLama.KernelMemory/LLamaSharpTextEmbeddingGenerator.cs
+++ b/LLama.KernelMemory/LLamaSharpTextEmbeddingGenerator.cs
@@ -3,6 +3,7 @@ using LLama.Common;
 using LLama.Native;
 using Microsoft.KernelMemory;
 using Microsoft.KernelMemory.AI;
+using Microsoft.KernelMemory.Context;
 
 namespace LLamaSharp.KernelMemory
 {
@@ -112,5 +113,24 @@ namespace LLamaSharp.KernelMemory
 
         /// <inheritdoc/>
         public int CountTokens(string text) => _embedder.Context.Tokenize(text, special: true).Length;
+
+        /// <summary>
+        /// Get the list of tokens for the input text
+        /// </summary>
+        /// <param name="text">Input string to be tokenized</param>
+        /// <returns>Read-only list of tokens for the input test</returns>
+        /// <remarks>
+        /// It throws if text is null and Includes empty stop token because addBos is left true to be consistent with the CountTokens implementation.</remarks>
+        /// <see cref="CountTokens(string)"/>
+        public IReadOnlyList<string> GetTokens(string text)
+        {
+            var context = _embedder.Context;
+            var embeddings = context.Tokenize(text, special: true);
+            var decoder = new StreamingTokenDecoder(context);
+            return embeddings
+                .Select(x => { decoder.Add(x); return decoder.Read(); })
+                .ToList()
+                .AsReadOnly();
+        }
     }
 }
--- a/LLama.Unittest/KernelMemory/ITextTokenizerTests.cs
+++ b/LLama.Unittest/KernelMemory/ITextTokenizerTests.cs
@@ -1,0 +1,81 @@
+using LLama.Common;
+using LLamaSharp.KernelMemory;
+using Microsoft.KernelMemory.AI;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+
+namespace LLama.Unittest.KernelMemory
+{
+    public abstract class ITextTokenizerTests
+    {
+        private readonly ITestOutputHelper _testOutputHelper;
+
+#pragma warning disable KMEXP00 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+        protected ITextTokenizer? _generator;
+#pragma warning restore KMEXP00 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+
+        protected InferenceParams _infParams;
+        protected LLamaSharpConfig _lsConfig;
+
+        public ITextTokenizerTests(ITestOutputHelper testOutputHelper)
+        {
+            _testOutputHelper = testOutputHelper;
+
+            _infParams = new() { AntiPrompts = ["\n\n"] };
+            _lsConfig = new(Constants.GenerativeModelPath) { DefaultInferenceParams = _infParams };
+
+            testOutputHelper.WriteLine($"Using model {Path.GetFileName(_lsConfig.ModelPath)}");
+        }
+
+        [Theory]
+        [InlineData("The quick brown fox jumps over the lazy dog")]
+        [InlineData("Well, here're some special characters!!!")]
+        [InlineData("And a little bit of unicode για να κρατήσουμε τα πράγματα ενδιαφέροντα")]
+        [InlineData("  \n  \r\n  \t   ")]
+        public void GetTokens_ShouldReturnListOfTokensForInputString(string? text)
+        {
+            var tokens = _generator!.GetTokens(text);
+            var tokensCount = _generator.CountTokens(text);
+
+            var expected = " " + text; // the placement of the space corresponding to BOS will vary by model
+            var actual = string.Join("", tokens);
+
+            _testOutputHelper.WriteLine($"Tokens for '{text}':");
+            _testOutputHelper.WriteLine(string.Join("", tokens.Select(x => $"({x})")));
+
+            Assert.Equal(expected, actual);
+            Assert.Equal(tokensCount, tokens.Count);
+        }
+
+        [Fact]
+        public void GetToken_ShouldThrowForNull()
+        {
+            string? text = null;
+
+            Assert.Throws<ArgumentNullException>(() => { _generator!.GetTokens(text!); });
+        }
+
+        [Fact]
+        public void GetToken_EmptyStringYieldsOneEmptyToken()
+        {
+            var text = "";
+            var expected = "";
+
+            var tokens = _generator!.GetTokens(text);
+            var tokensCount = _generator.CountTokens(text);
+            var actual = tokens.Single();
+
+            _testOutputHelper.WriteLine($"Tokens for '{text}':");
+            _testOutputHelper.WriteLine(string.Join("", tokens.Select(x => $"({x})")));
+
+            Assert.Equal(expected, actual);
+            Assert.Equal(tokensCount, tokens.Count);
+        }
+    }
+}
--- a/LLama.KernelMemory/LLamaSharp.KernelMemory.csproj
+++ b/LLama.KernelMemory/LLamaSharp.KernelMemory.csproj
@@ -27,7 +27,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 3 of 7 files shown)


LLama.KernelMemory/LLamaSharpTextEmbeddingGenerator.cs
```diff
@@ -112,5 +113,24 @@ public async Task<Embedding> GenerateEmbeddingAsync(string text, CancellationTok
...
+        /// <param name="text">Input string to be tokenized</param>
+        /// <returns>Read-only list of tokens for the input test</returns>
+        /// <remarks>
+        /// It throws if text is null and Includes empty stop token because addBos is left true to be consistent with the CountTokens implementation.</remarks>
+        /// <see cref="CountTokens(string)"/>
+        public IReadOnlyList<string> GetTokens(string text)
+        {
+            var context = _embedder.Context;
+            var embeddings = context.Tokenize(text, special: true);
+            var decoder = new StreamingTokenDecoder(context);
```
martindevans, line 129: The purpose of `StreamingTokenDecoder` is to handle the case where one single _character_ is represented by multiple _tokens_ (e.g. some emoji or various non-english characters).

For example let's say the character `A` is represented by the tokens `[1, 2, 3]`. The current code will:
1. Add `1` to the decoder, read an empty string (because `1` on it's own can't be decoded).
2. Add `2` to the decoder, read empty again
3. Add `3` to the decoder, finally read `A`

So the output would be `[ "", "", "A" ]` which obviously is a bit weird!

To be honest I'm not sure what the right fix is here, some tokens simply don't have a string representation - this seems like a fairly badly designed API by the KernelMemory team! We could just stick with the current implementation, that does at least output one string per token, even if some of those strings are empty.

LLama.KernelMemory/LLamaSharpTextEmbeddingGenerator.cs
```diff
@@ -112,5 +113,24 @@ public async Task<Embedding> GenerateEmbeddingAsync(string text, CancellationTok
...
+            return embeddings
+                .Select(x => { decoder.Add(x); return decoder.Read(); })
+                .ToList()
+                .AsReadOnly();
```
martindevans, line 133: No need to call `AsReadOnly()` here, `List<T>` is already `IReadOnlyList<T>`

LLama.Unittest/KernelMemory/ITextTokenizerTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            _testOutputHelper = testOutputHelper;
+
+            _infParams = new() { AntiPrompts = ["\n\n"] };
+            _lsConfig = new(Constants.GenerativeModelPath) { DefaultInferenceParams = _infParams };
+
+            testOutputHelper.WriteLine($"Using model {Path.GetFileName(_lsConfig.ModelPath)}");
+        }
+
+        [Theory]
```
martindevans, line 36: Could you try adding this string https://github.com/SciSharp/LLamaSharp/blob/master/LLama.Unittest/StreamingTextDecoderTests.cs#L45 to your test cases, you'll probably get some of the weird behaviour with empty strings I mentioned above. At least then it will be documented by the tests.

martindevans:
I've submitted a few review comments. The one with the empty strings I'm not really sure how best to handle, and if you want to go ahead with the current implementation I'm happy with that as long as there's at least a test covering this weirdness and a comment explaining what's going on.

LLama.Unittest/KernelMemory/ITextTokenizerTests.cs
SpaceAntelope: I get:
```
Using model llama-2-7b-chat.Q3_K_S.gguf
Tokens for '猫坐在垫子上 😀🤨🤐😏':
()( )()()(猫)()()(坐)(在)()()(垫)(子)(上)( )()()()(😀)()()()(🤨)()()()(🤐)()()()(😏)
```

Could this be a limitation of the generative model used for unit testing? llama-3 seems to do a bit better:
```
Using model Meta-Llama-3-8B-Instruct.Q8_0.gguf
Tokens for '猫坐在垫子上 😀🤨🤐😏':
()(猫)(坐在)()(垫)(子)(上)( 😀)()()(🤨)()()(🤐)()(😏)
```

I also noticed llama-2 has difficulties parsing the greek fragment and simply treats every letter as its own token, while llama-3 appears better at parsing it as a language:

```
Using model llama-2-7b-chat.Q3_K_S.gguf
Tokens for 'And a little bit of unicode για να κρατήσουμε τα πράγματα ενδιαφέροντα':
()( And)( a)( little)( bit)( of)( unicode)( )(γ)(ι)(α)( )(ν)(α)( )(κ)(ρ)(α)(τ)(ή)(σ)(ο)(υ)(μ)(ε)( )(τ)(α)( )(π)(ρ)(ά)(γ)(μ)(α)(τ)(α)( )(ε)(ν)(δ)(ι)(α)(φ)(έ)(ρ)(ο)(ν)(τ)(α)
```
vs
```
Using model Meta-Llama-3-8B-Instruct.Q8_0.gguf
Tokens for 'And a little bit of unicode για να κρατήσουμε τα πράγματα ενδιαφέροντα':
()(And)( a)( little)( bit)( of)( unicode)( για)( να)( κρα)(τή)(σ)(ουμε)( τα)( πρά)(γ)(ματα)( εν)(δια)(φέ)(ρον)(τα)
```

martindevans: The cause is that some individual characters are actually multiple tokens. The character `🤨` is the tokens `<0xF0><0x9F><0xA4><0xA8>` (using llama2 tokenizer, it'll vary between different tokenizers but the fundamental issue is the same). As an example if you call `GetTokens("🤨")` it will do:

1. Add `0xF0` to the StreamingTokenDecoder
2. Read() returns nothing, because that token can't be decoded on it's own
3. Add `0x9F` 
4. Read() returns nothing again
5. Add `0xA4`
6. Read() returns nothing again
7. Add `0xA8`
8. Read() finally returns "🤨"

The `GetTokens` method in KernelMemory assumes that each individual token has a string associated with it, but that's not true! That's why I commented it's bad API design by KernelMemory.

There are three ways I can see to handle this:
1. Remove the empty strings. This seems like the worst options to me, because then CountTokens will return a different count to what GetTokens returns!
2. Do as you've currently done, return empty strings for all the tokens that cannot be decoded. This way the count is correct, and the output is human readable.
3. Simply return the numeric value of every token, this way the output is less "weird", but completely unreadable to humans. In this case GetTokens would become:

```csharp
var tokens = context.Tokenize(text, special: true);
return tokens.Select(a => a.ToString()).ToArray();
```

I'm happy to go with 2 or 3, it's up to you.

LLama.KernelMemory/LLamaSharpTextEmbeddingGenerator.cs
SpaceAntelope: I arrived to StreamingTokenDecoder via the _obsolete_ message on the Detokenize method, which was a wrapper for StreamingTokenDecoder anyway. If there's a more appropriate way to decode tokens I'd be glad to switch.

Also, am I correct in referring to the result of the .Tokenize method as embeddings? In retrospect I think _numbers instead of chunks of letters, must be embeddings_ was quite an assumption on my part.

LLama.KernelMemory/LLamaSharpTextEmbeddingGenerator.cs
SpaceAntelope: Well that's an easy fix at least.

LLama.Unittest/KernelMemory/ITextTokenizerTests.cs
SpaceAntelope: I'd say 2, if anything returning a List<string> of numbers seems very counter intuitive and as you say the KernelMemory implementation returns strings. Hopefully in the future things will become clearer,

I sneaked a call to GetTokens in their CountTokens unit test and here's what it currently does:

https://github.com/microsoft/kernel-memory/blob/main/extensions/LlamaSharp/LlamaSharp.FunctionalTests/LlamaSharpTextGeneratorTest.cs
```
warn: Microsoft.KernelMemory.AI.LlamaSharp.LlamaSharpTextGenerator[0]
      Tokenizer not specified, will use GPT4Tokenizer. The token count might be incorrect, causing unexpected errors

[...]
Tokens for '猫坐在垫子上 😀🤨🤐😏'
(猫)(猫)(猫)(坐)(坐)(在)(垫)(垫)(垫)(子)(上)( 😀)(🤨)(🤨)(🤨)(🤐)(🤐)(🤐)(😏)(😏)
```

martindevans: Huh, looks like they've gone for a slightly different approach! They're duplicating the tokens, so if tokens `[1, 2, 3]` decode to `A` they're outputing `["A", "A", "A"]` instead of `[" ", " ", "A"]`.

For now, I'm happy to merge this as is if you are? I imagine this is an area of Kernel Memory that's going to get a bit of a redesign in the future!

SpaceAntelope:
@martindevans I pushed the relevant changes. I created a duplicate unit test with only the unicode cases and added this comment i(also referenced in the GetTokens implementations) :

```
  /* This is exactly the same test as the non-unicode cases. However, there are reasons why this
   * should be made a special case and may deviate in the future:
   * 
   * As of now there appears to be no final word as to how characters that consist of more than one 
   * numeric token should correspond to textual tokens, and results vary according to different 
   * models' tokenizers. For example, given a character 'Z' that corresponds to the numeric tokens {1,2,3} 
   * some (llama-2) will pad the length of the total number of tokens by returning spaces as tokens 
   * (i.e. ' ', ' ', 'Z') while others (GPT4Tokenizer) will pad with the character itself (i.e. 'Z','Z','Z').
   * 
   * This is very evident when tokenizing ideograms and emojis, but can arise with various unicode characters 
   * as well. See pull request for more relevant discussion https://github.com/SciSharp/LLamaSharp/pull/862
   *
   * Currently the method will remain consistent with the output of ITextTokenizer.CountTokens, meaning
   * any redundant tokens will not be ommited as long as they are counted by CountTokens.
   * 
   * StreamingTokenDecoder, while sufficiently useful for this task, was not designed with producing
   * output for one numeric token at a time in mind, so ITextTokenizer.GetTokens should not be considered 
   * an example of proper use.
   * 
   * Note: if this message is removed, also remove references to it in LLamaSharpTextEmbeddingGenerator.GetTokens
   * and LLamaSharpTextGenerator.GetTokens
   */
```

Issue #884, opened 2024-08-01
[BUG]: v0.14.0 hangs when initializing ModelParams


SpaceAntelope:
### Description

I feel that updating GPU drivers might work for most people, but since it's a serious problem and the solution is not necessarily obvious I decided to post a full bug report for anyone afflicted to find. 

The problem appears to be that the result of calling vulkaninfo in **/LLamaSharp/LLama/Native/Load/SystemInfo.cs** as a process and reading the output stream varies a lot. In my case in one machine it hangs, in another it returns empty and in the last one it works fine. Removing **CreateNoWindow = true**  from ProcessStartInfo seems to fix it.

I fixed the one (Win10 22H2) where it kept hanging by updating the GPU drivers which raised the vulkan instance version from 1.2.198 to 1.3.204. The machine (Win11 23H2) where the output stream comes up empty has v1.3.201, but since LLamaSharp accounts for dodgy vulkaninfo outputs this doesn't seem to cause a problem.

Note that vulkaninfo behaves the same when initialized as a System.Diagnostics.Process in a simple console app (i.e. it hangs).

### Reproduction Steps

With .NET8 SDK installed:

repro.csproj:
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <RootNamespace>vulkan_test</RootNamespace>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="llamasharp" Version="0.14.0" />
    <PackageReference Include="llamasharp.backend.cpu" Version="0.14.0" />
  </ItemGroup>

</Project>
```

Program.cs:
```
using System.Text.Json;

var modelParams = new LLama.Common.ModelParams("");

var json = JsonSerializer.Serialize(modelParams, new JsonSerializerOptions { WriteIndented = true });

Console.WriteLine(json);
```

**Expected**: Print the default values for ModelParams in json form.
**Actual**: If you're unlucky it hangs forever.

### Environment & Configuration

- Operating system: **Win10**
- .NET runtime version: **8.0.303**
- LLamaSharp version: **0.14.0**
- CUDA version (if you are using cuda backend): **the bug is hit before it attempts to load the model, it doesn't even complain about missing a backend package**
[...]

PR #885, opened 2024-08-01
Safer Vulkan System Info

martindevans:
Better handling of the vulkaninfo process, in case it hangs.

See #884


m0nsky:
Tested on Ubuntu 22.04 w/ Vulkan and everything is working fine here, I hope it fixes #884

PR #907, opened 2024-08-24
Remove OpenCL & add Vulkan to readme

m0nsky:
Since OpenCL has now been removed, and Vulkan has been introduced, this PR removes OpenCL from the readme and adds Vulkan (including the nuget links)


martindevans:
Thanks!

PR #943, opened 2024-10-08
Specify file open permissions needed as read only. 

scritch1sm:
Default open behavior looks to be read/write.  Resulting in failed file open of model in environments with strict read-only file system settings)

### DIFF
--- a/LLama/Native/SafeLlamaModelHandle.cs
+++ b/LLama/Native/SafeLlamaModelHandle.cs
@@ -129,7 +129,7 @@ namespace LLama.Native
             // - File exists (automatically throws FileNotFoundException)
             // - File is readable (explicit check)
             // This provides better error messages that llama.cpp, which would throw an access violation exception in both cases.
-            using (var fs = new FileStream(modelPath, FileMode.Open))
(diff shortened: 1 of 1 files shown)


martindevans:
This is a good idea, and I'll merge it after the tests have run. Thanks!

However, I'm not sure if it will entirely fix the problem. The `FileStream` there is not actually used to read the model, all we're doing is opening and immediately closing the file. This acts as a sanity check that the file exists and is readable before we pass off to llama.cpp to do the _actual_ work of reading the file.

The model is really loaded with `llama_load_model_from_file` (a couple of lines down). I don't know whether ot not that loads the file with read-only permissions.

scritch1sm:
makes sense, but FWIW in my particular use case (i can't share too much of
the particulars as it came up during work), it did resolve an issue.  the
filesystem in question simply didn't allow opening for write and would
throw an exception here, causing loading the model to fail outright here on
this line.  I compiled this in and was able to successfully load the model
in context and the application moved on with life.
I then ran into OTHER issues, but that's an unrelated other story.  =)

PR #951, opened 2024-10-15
Update "ContributingGuide" for latest llama.cpp

scritch1sm:


docs/ContributingGuide.md
```diff
@@ -32,29 +32,25 @@ As shown in [llama.cpp cmake file](https://github.com/ggerganov/llama.cpp/blob/m
...
 option(LLAMA_BUILD_SERVER               "llama: build server example")// Please disable it.
 ```
 
-Most importantly, `-DBUILD_SHARED_LIBS=ON` must be added to the cmake instruction and other options depends on you. For example, when building with cublas but without openblas, use the following instruction:
+Most importantly, `-DBUILD_SHARED_LIBS=ON` must be added to the cmake instruction and other options depends on you. For example, when building with CUDA but without openblas, use the following instruction:
 
 ```bash
 mkdir build && cd build
-cmake .. -DLLAMA_CUBLAS=ON -DBUILD_SHARED_LIBS=ON
+cmake .. -DGGML_CUDAS=ON -DBUILD_SHARED_LIBS=ON
```
martindevans, line 49: should this be `GGML_CUDA`?

llama.cpp
martindevans, line 1: Looks like you've committed an update to llama.cpp, can you undo that and just submit the changes to ContributingGuide.md. Thanks.

martindevans (changes requested):

llama.cpp
scritch1sm: ah crud.  sorry about that.  didn't notice that snuck in there.

PR #961, opened 2024-10-25
Custom Sampler Stages

martindevans:
Added support for custom sampler stages, written entirely in safe C#. These stages can be added to a sampler pipeline just like any other. See the example in `LLama.Examples\Examples\CustomSampler.cs` for an example of a custom pipeline using a custom stage.

The internals here are fairly complex, because we need to keep some manually managed memory (so we can have persistent pointers to it) and also make sure the GC doesn't collect things. The overall flow is as follows:
 - A custom sampler is written by the user implementing `ICustomSampler`
 - This is added to a pipeline with `AddCustom`
 - A `CustomSamplerHandle` is allocated, to own all of the various bits of memory
 - The `CustomSamplerHandle` holds a `GCHandle` which keeps **itself** alive. 
 - `CustomSamplerHandle` also allocates some memory to hold some native structures (`_samplerNativePtr` and `_samplerNativeInterfacePtr`) and passes pointers across to the native side.
 - When the native side calls `free` it will destroy the `GCHandle` keeping itself alive, free the native memory, and finally calls `ICustomSampler.Free`.

### DIFF
--- a/LLama.Examples/Examples/CustomSampler.cs
+++ b/LLama.Examples/Examples/CustomSampler.cs
@@ -86,7 +86,7 @@ namespace LLama.Examples.Examples
             // Make the most likely token impossible to pick
             tokenData.Data[0].Logit = float.NegativeInfinity;
             
-            // It's critically important to set this if the logits are no logner sorted after the custom 
+            // It's **critically** important to set this if the logits are no longer sorted after the custom 
             // sampler has run. If you're not sure, it's always safer to set it to false.
             //
             // In this case, because the first logit has just been set to negative infinity
--- a/LLama.Examples/Examples/CustomSampler.cs
+++ b/LLama.Examples/Examples/CustomSampler.cs
@@ -1,0 +1,114 @@
+using LLama.Common;
+using LLama.Examples.Extensions;
+using LLama.Native;
+using LLama.Sampling;
+
+namespace LLama.Examples.Examples
+{
+    public class CustomSampler
+    {
+        public static async Task Run()
+        {
+            var modelPath = UserSettings.GetModelPath();
+
+            var parameters = new ModelParams(modelPath);
+            using var model = await LLamaWeights.LoadFromFileAsync(parameters);
+
+            var ex = new StatelessExecutor(model, parameters);
+
+            Console.ForegroundColor = ConsoleColor.Yellow;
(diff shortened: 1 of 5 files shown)


LLama.Examples/Examples/CustomSampler.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            // Make the most likely token impossible to pick
+            tokenData.Data[0].Logit = float.NegativeInfinity;
+            
+            // It's **critically** important to set this if the logits are no longer sorted after the custom 
+            // sampler has run. If you're not sure, it's always safer to set it to false.
+            //
+            // In this case, because the first logit has just been set to negative infinity
+            // the token data is definitely not sorted!
+            tokenData.Sorted = false;
```
Lyrcaxis, line 94: Should also move *some* of this explanation on the method's documentation imo -- if it's THAT important
image

LLama/Native/SafeLLamaSamplerHandle.cs
```diff
@@ -589,5 +741,4 @@ public interface ICustomSampler
     /// Free all unmanaged resources held by this sampler
     /// </summary>
     void Free();
```
Lyrcaxis, line 743: Maybe rename to `Dispose()` and have `ICustomSampler` implement `IDisposable`?
..and potentially add some automated version of disposing it (?)

I'm aware this is kinda edge case though quite useful still.

Lyrcaxis (approved):

LLama/Native/SafeLLamaSamplerHandle.cs
martindevans: Good idea, I've made that change with `IDisposable`.

> and potentially add some automated version of disposing it (?)

It's kinda automated at the moment. When the pipeline is disposed all of the samplers are destroyed, including any custom ones implemented in C#. Ultimately it's all owned by `SafeLLamaSamplerChainHandle`, as long as that is disposed (or it's finalizer runs) all of the pipeline resources will be safely destroyed. i.e. the GC is the ultimate fallback thanks to the finalizer.

LLama.Examples/Examples/CustomSampler.cs
martindevans: I've rewritten these docs

PR #962, opened 2024-10-26
November Binary Update

martindevans:
Updated to [958367bf530d943a902afa1ce1c342476098576b](https://github.com/ggerganov/llama.cpp/commit/958367bf530d943a902afa1ce1c342476098576b), built with https://github.com/SciSharp/LLamaSharp/actions/runs/11509292954

This is a much smaller update that the last one. Only notable changes are:
 - Some changes in the fill-in-middle (FIM) token functions (no change to C# API)
 - New FIM tokens exposed, not sure what these are for so no useful comments, if anyone knows a tip would be appreciated!
 - Removed softmax sampler. It's deprecated upstream, so it's best to remove it quickly on our end since that's a new API no one will be depending on yet.
 - Removed tail free sampler. Also deprecated upstream (https://github.com/ggerganov/llama.cpp/pull/10071)

Testing:
- [x] Windows CPU
- [x] Windows CUDA
- [x] Windows Vulkan
- [x] Linux CPU
- [x] Linux CUDA
- [x] Linux Vulkan
- [x] MacOS CPU
- [x] MacOS Metal


m0nsky:
Unit tests passed on Windows CUDA.

- Windows CPU
- Windows CUDA
- Windows Vulkan
- Linux CPU
- Linux CUDA
- Linux Vulkan

PR #964, opened 2024-10-31
Add Microsoft.Extensions.AI support for IChatClient / IEmbeddingGenerator

stephentoub:
https://devblogs.microsoft.com/dotnet/introducing-microsoft-extensions-ai-preview/

### DIFF
--- a/LLama/Extensions/LLamaExecutorExtensions.cs
+++ b/LLama/Extensions/LLamaExecutorExtensions.cs
@@ -1,0 +1,169 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using LLama.Common;
+using LLama.Sampling;
+using Microsoft.Extensions.AI;
+
+namespace LLama.Abstractions;
+
+/// <summary>
+/// Extension methods to the <see cref="LLamaExecutorExtensions" /> interface.
+/// </summary>
+public static class LLamaExecutorExtensions
+{
+    /// <summary>Gets an <see cref="IChatClient"/> instance for the specified <see cref="ILLamaExecutor"/>.</summary>
+    /// <param name="executor">The executor.</param>
+    /// <param name="historyTransform">The <see cref="IHistoryTransform"/> to use to transform an input list messages into a prompt.</param>
+    /// <param name="outputTransform">The <see cref="ITextStreamTransform"/> to use to transform the output into text.</param>
+    /// <returns>An <see cref="IChatClient"/> instance for the provided <see cref="ILLamaExecutor" />.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="executor"/> is null.</exception>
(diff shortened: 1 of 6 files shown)


LLama/Extensions/SpanNormalizationExtensions.cs
```diff
@@ -81,6 +81,19 @@ public static Span<float> EuclideanNormalization(this Span<float> vector)
         return vector;
     }
 
+    /// <summary>
+    /// <b>In-place</b> divide every element by the euclidean length of the vector
```
martindevans, line 85: This one isn't in-place. I assume this is required because one of the new APIs only gives you a `ReadOnlySpan`, so it can't be implemented in place?

LLama/Extensions/LLamaExecutorExtensions.cs
```diff
@@ -0,0 +1,169 @@
...
+            {
+                AntiPrompts = antiPrompts,
+                TokensKeep = options?.AdditionalProperties?.TryGetValue(nameof(InferenceParams.TokensKeep), out int tk) is true ? tk : s_defaultParams.TokensKeep,
+                MaxTokens = options?.MaxOutputTokens ?? 256, // arbitrary upper limit
+                SamplingPipeline = new DefaultSamplingPipeline()
+                {
+                    AlphaFrequency = options?.AdditionalProperties?.TryGetValue(nameof(DefaultSamplingPipeline.AlphaFrequency), out float af) is true ? af : s_defaultPipeline.AlphaFrequency,
+                    AlphaPresence = options?.AdditionalProperties?.TryGetValue(nameof(DefaultSamplingPipeline.AlphaPresence), out float ap) is true ? ap : s_defaultPipeline.AlphaPresence,
+                    PenalizeEOS = options?.AdditionalProperties?.TryGetValue(nameof(DefaultSamplingPipeline.PenalizeEOS), out bool pe) is true ? pe : s_defaultPipeline.PenalizeEOS,
+                    PenalizeNewline = options?.AdditionalProperties?.TryGetValue(nameof(DefaultSamplingPipeline.PenalizeNewline), out bool pn) is true ? pn : s_defaultPipeline.PenalizeNewline,
```
martindevans, line 146: The spellchecker has a false positive on this line. If you add `pn` to the typos.toml file that should suppress it. (docs: https://github.com/marketplace/actions/typos-action#false-positives)

martindevans:
This looks good, thanks for the PR. The new `Microsoft.Extensions.AI` types look like they'll be really helpful as a foundation for us to build around in the future!

LLama/Extensions/SpanNormalizationExtensions.cs
stephentoub: Yup (or, rather, a `ReadOnlyMemory<T>`, but yea). Fixing comment.

LLama/Extensions/LLamaExecutorExtensions.cs
stephentoub: I'll just rename it. Thanks.

PR #981, opened 2024-11-13
Fixed `LLamaTokenDataArrayNative.Create`'s size initialization

Lyrcaxis:
This PR fixes an initialization issue of `LLamaTokenDataNativeArray`'s `_size` property, which is initialized at 0
image
but when `.Create()` is called, it attempts to set _size to some value that HAS to be larger than 0,
image
..which triggers this exception:
image


martindevans:
Thanks

PR #983, opened 2024-11-13
Updated tests model to "LLama 3.2 1B Instruct"

Lyrcaxis:
This PR makes the default model used for CI tests be the smaller `LLaMA-3.2-1B-Instruct-Q4_0` instead of the significantly larger `llama-2-7b-chat.Q3_K_S`. This should speed up CI tests and save a couple GB in space.


LLama.Unittest/LLamaEmbedderTests.cs
```diff
@@ -86,7 +86,7 @@ public async Task EmbedCompareEmbeddingModel()
     [Fact]
     public async Task EmbedCompareGenerateModel()
     {
-        await CompareEmbeddings(Constants.GenerativeModelPath);
+        await CompareEmbeddings(Constants.LLavaModelPath);
```
martindevans, line 89: This is meant to be using the generative model (hence the name) not the llava model

martindevans (changes requested):
Thanks a lot for this, it should really help out with test stability and speed in CI! I've pointed out a couple of minor issues that need fixing

PR #993, opened 2024-11-24
Non-deterministic default seed (+minor sampling parameters names & comments update)

Lyrcaxis:
Distribution seed for sampling used to always be fixed (42) by default, causing deterministic sampling.
This PR switches it to be randomly initialized with each new sampling pipeline instance.

Also changed some of the sampling pipeline's members' names and comments to better reflect their purpose.
- `AlphaFrequency` --> `FrequencyPenalty`
- `AlphaPresence` --> `PresencePenalty`
- `PenalizeEOS` --> `PreventEOS`
###### `PenalizeEOS` actually completely takes away the possibility for `EOS_TOKEN` to be sampled, causing infinite generation.

### DIFF
--- a/LLama/Sampling/DefaultSamplingPipeline.cs
+++ b/LLama/Sampling/DefaultSamplingPipeline.cs
@@ -25,15 +25,15 @@ public sealed class DefaultSamplingPipeline
     /// Number between -2.0 and 2.0. Positive values penalize new tokens based on their existing frequency in the text
     /// so far, decreasing the model's likelihood to repeat the same line verbatim.
     /// </summary>
-    public float AlphaFrequency
+    public float FrequencyPenalty
     {
         get => _alphaFreq;
         init
         {
             if (value < -2)
-                throw new ArgumentOutOfRangeException(nameof(value), "AlphaFrequency must be greater than -2");
+                throw new ArgumentOutOfRangeException(nameof(value), $"{nameof(FrequencyPenalty)} must be greater than -2");
             if (value > 2)
-                throw new ArgumentOutOfRangeException(nameof(value), "AlphaFrequency must be less than 2");
+                throw new ArgumentOutOfRangeException(nameof(value), $"{nameof(FrequencyPenalty)} must be less than 2");
             _alphaFreq = value;
         }
     }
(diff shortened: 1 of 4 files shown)


LLama/Sampling/DefaultSamplingPipeline.cs
```diff
@@ -111,7 +111,7 @@ public float AlphaPresence
     /// <summary>
     /// Seed to use for random sampling
     /// </summary>
-    public uint Seed { get; set; } = 42;
+    public uint Seed { get; set; } = (uint) new Random().Next(0, int.MaxValue);
```
martindevans, line 114: I think it would be better to use [`RandomNumberGenerator`](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.randomnumbergenerator?view=net-9.0) or a single static Random instance (protected by a lock). Afaik `new Random()` doesn't have very good distribution if you create many all at the same time.

LLama/Sampling/DefaultSamplingPipeline.cs
```diff
@@ -25,15 +25,15 @@ public sealed class DefaultSamplingPipeline
     /// Number between -2.0 and 2.0. Positive values penalize new tokens based on their existing frequency in the text
     /// so far, decreasing the model's likelihood to repeat the same line verbatim.
     /// </summary>
-    public float AlphaFrequency
+    public float FrequencyPenalty
```
martindevans, line 28: Can we keep the old properties (as well), but tagged with an `[Obsolete]` attribute (including a message suggesting to use the new properties). That way we can leave that in for a couple of versions to steer people towards the new properties before making the breaking change,

martindevans (changes requested):
Looks good, just a couple of minor nits :)

LLama/Sampling/DefaultSamplingPipeline.cs
Lyrcaxis: Would lean towards expecting the users to manually specify `Seed` on those cases. Thoughts?

martindevans: Even so, It's a potential footgun that's easy to fix, so we may as well!

Issue #995, opened 2024-11-25
[BUG]: loglevel 0 and 1 from llama.cpp doesn't seem to be supported


LoicDagnas:
### Description

Since [this PR](https://github.com/ggerganov/llama.cpp/pull/9418) on llama.cpp, log levels are not anymore between 2 and 5 but between 0 and 4 [here](https://github.com/ggerganov/llama.cpp/blob/master/ggml/include/ggml.h#L559). So I guess that the log level mapping should be adapted [here](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Native/LLamaLogLevel.cs).

### Reproduction Steps

Adding with `NativeLogConfig.llama_log_set(new LlamaSharpLogger());` this very naive custom logger:

```csharp
internal sealed class LlamaSharpLogger : ILogger
{
    public IDisposable BeginScope<TState>(TState state) => default;

    public void Log<TState>(
        LogLevel logLevel,
        EventId eventId,
        TState state,
        Exception exception,
        Func<TState, Exception, string> formatter)
    {

    }

    public bool IsEnabled(LogLevel logLevel) => true;
}
```

any call to `LlamaWeights.LoadFromFile("...")` will fails with an `ArgumentOutOfRangeException`.

### Environment & Configuration

- Operating system: Windows
- .NET runtime version: 8.0
- LLamaSharp version: 0.19
- CUDA version (if you are using cuda backend): 12
- CPU & GPU device: RTX 3000

martindevans:
Thanks for the detailed report!

You've located the correct lcoation to fix the problem, would you be interested in putting together a PR with the necessary changes?

martindevans:
This was resolved by #997

PR #997, opened 2024-11-26
The log levels defined on llama.cpp and LlamaSharp side were not aligned anymore (issue #995)

LoicDagnas:
Following this issue https://github.com/SciSharp/LLamaSharp/issues/995

After [this PR](https://github.com/ggerganov/llama.cpp/pull/9418) on the llama.cpp repo the loglevel defined on both llama.cpp and LlamaSharp sides were not aligned anymore. This misalignment made the model weight loading fail when a custom logger is used.

### Help
I am just wondering if it is correct to consider the `CONT` log level as equivalent to `NONE`.

### DIFF
--- a/LLama/Native/LLamaLogLevel.cs
+++ b/LLama/Native/LLamaLogLevel.cs
@@ -4,14 +4,26 @@ using Microsoft.Extensions.Logging;
 namespace LLama.Native
 {
     /// <summary>
-    /// Severity level of a log message
+    /// Severity level of a log message. This enum should always be aligned with
+    /// the one defined on llama.cpp side at
+    /// https://github.com/ggerganov/llama.cpp/blob/0eb4e12beebabae46d37b78742f4c5d4dbe52dc1/ggml/include/ggml.h#L559
     /// </summary>
     public enum LLamaLogLevel
     {
         /// <summary>
-        /// Logs that highlight when the current flow of execution is stopped due to a failure.
+        /// Logs are never written.
         /// </summary>
-        Error = 2,
+        None = 0,
+
+        /// <summary>
+        /// Logs that are used for interactive investigation during development.
+        /// </summary>
+        Debug = 1,
+
+        /// <summary>
+        /// Logs that track the general flow of the application.
+        /// </summary>
+        Info = 2,
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


martindevans:
It looks like the idea of `GGML_LOG_LEVEL_CONT` is simply to re-use whatever the level was for the last message. Setting that to `None` will presumably mean that some message using this will be cut in half (because the CONT part will be hidden, due to treating it as None).

The best way I can think of to handle this is to have a threadstatic field which stores the last log level used. I'll add a comment inline to show what I mean.

LLama/Native/LLamaLogLevel.cs
```diff
@@ -4,29 +4,41 @@
...
+        /// </summary>
+        Error = 4,
+        
+        /// <summary>
+        /// Continue log level is equivalent to None in the way it is used in llama.cpp.
+        /// </summary>
+        Continue = 5,
     }
 
     internal static class LLamaLogLevelExtensions
```
martindevans, line 44:
```suggestion
    internal static class LLamaLogLevelExtensions
    {
        [ThreadStatic] private static LogLevel _previous;
        
        public static LogLevel ToLogLevel(this LLamaLogLevel llama)
        {
            _previous = (llama) switch
            {
                LLamaLogLevel.None => LogLevel.None,
                LLamaLogLevel.Debug => LogLevel.Debug,
                LLamaLogLevel.Info => LogLevel.Information,
                LLamaLogLevel.Warning => LogLevel.Warning,
                LLamaLogLevel.Error => LogLevel.Error,
                LLamaLogLevel.Continue => _previous,
                _ => throw new ArgumentOutOfRangeException(nameof(llama), llama, null)
            };
            return _previous;
    }
```

martindevans: (Note that I haven't tested this, but hopefully it illustrates what I mean)

LoicDagnas: Got it, good idea

martindevans (approved):

martindevans:
Looks good to me, thanks for fixing this! Just waiting for the tests to run and then I'll merge it :)

martindevans:
I'll re-run it, unfortunately our CI is flakey sometimes

martindevans:
Looks like it worked this time.

PR #1025, opened 2024-12-31
Adds an option to apply the chat template to prompts when using StateslessExecutor

phil-scott-78:
Also updates LLamaStatelessExecutor.cs to use the new functionality as an example.


phil-scott-78:
this executor is used elsewhere, but I only updated this one example. Some of the other ones will use it combined with `ChatSession`, or at least `PromptTemplateTransformer` so they seem to be like they get the prompt templating via that.

martindevans:
Thanks!

PR #1034, opened 2025-01-02
Adds an exception for missing templates.

### LINKED ISSUE #1032: [BUG]: unknown templates generate exception
### Description

Using a model without a matching template, use `LlamaTemplate` passing in the model to pull the template from the model. Calling Apply results in an index out of range exception. The reason is that `llama_chat_apply_template` returns a -1 instead of the length when there isn't a matching model, and resulting in this line failing 
```csharp
output.AsSpan(0, outputLength).CopyTo(_result);
```

### Reproduction Steps

Using `Ministral-8B-Instruct-2410-Q6_K_L.gguf`, use `LlamaTemplate` passing in the model to pull the template from the model. The `StatelessModeExecute` is an easy way to reproduce now that it is tries to apply the system template, but anything using `LlamaTemplate` will get an exception.

- Operating system:  Windows 11
- .NET runtime version: .net 9
- LLamaSharp version: main branch
- CUDA version (if you are using cuda backend): 12

### Known Workarounds

Following the llama.cpp lead (https://github.com/ggerganov/llama.cpp/blob/master/src/llama.cpp#L23348), if we can't find a template then apply Chatml. Funny enough this doesn't work at all for the model I'm testing, but it *feels* like the proper behavior until llama.cpp gets updated to support it, I suppose. llama.cpp has `llama_chat_detect_template` they can call to complete circumvent even trying to apply the template which would probably be the move, but that's no exposed. in the meantime, this seems to work

```csharp
var outputLength = ApplyInternal(_nativeChatMessages.AsSpan(0, Count), output);
if (outputLength == -1)
{
    // worst case: there is no information about template, we will use chatml by default
    outputLength = ApplyChatmlInternal(_nativeChatMessages.AsSpan(0, Count), output);
}
// snip
unsafe int ApplyChatmlInternal(Span<LLamaChatMessage> messages, byte[] output)
{
    fixed (byte* customTemplatePtr = Encoding.GetBytes("chatml\0"))
    fixed (byte* outputPtr = output)
    fixed (LLamaChatMessage* messagesPtr = messages)
    {
        return NativeApi.llama_chat_apply_template(_model, customTemplatePtr, messagesPtr, (nuint)messages.Length, AddAssistant, outputPtr, output.Length);
    }
}
```

phil-scott-78:
Adds better exception handling for when there are missing templates with a model when calling `llama_chat_apply_template`

Funny enough, it looks like https://github.com/ggerganov/llama.cpp/pull/10572 added support for my missing template, but still nice to have a better exception down the road.

Closes #1032

### DIFF
--- a/LLama/Exceptions/RuntimeError.cs
+++ b/LLama/Exceptions/RuntimeError.cs
@@ -59,6 +59,25 @@ public class LLamaDecodeError
 }
 
 /// <summary>
+/// `llama_decode` return a non-zero status code
+/// </summary>
+public class MissingTemplateException
+    : RuntimeError
+{
+    /// <inheritdoc />
+    public MissingTemplateException()
+        : base("llama_chat_apply_template failed: template not found")
+    {
+    }
+    
+    /// <inheritdoc />
+    public MissingTemplateException(string message)
+        : base($"llama_chat_apply_template failed: template not found for '{message}'")
+    {
+    } 
+}
+
+/// <summary>
 /// `llama_get_logits_ith` returned null, indicating that the index was invalid
 /// </summary>
 public class GetLogitsInvalidIndexException
--- a/LLama/LLamaTemplate.cs
+++ b/LLama/LLamaTemplate.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text;
+using LLama.Exceptions;
 using LLama.Native;
 
 namespace LLama;
@@ -250,6 +251,19 @@ public sealed class LLamaTemplate
                 {
                     // Run templater and discover true length
                     var outputLength = ApplyInternal(_nativeChatMessages.AsSpan(0, Count), output);
+                    
+                    // if we have a return code of -1, the template was not found.
+                    if (outputLength == -1)
+                    {
+                        if (_customTemplate != null)
+                        {
+                            throw new MissingTemplateException(Encoding.GetString(_customTemplate));
+                        }
+                        else
+                        {
+                            throw new MissingTemplateException();    
+                        }
+                    }    
 
                     // If length was too big for output buffer run it again
                     if (outputLength > output.Length)


LLama/Exceptions/RuntimeError.cs
```diff
@@ -58,6 +58,25 @@ public LLamaDecodeError(DecodeResult returnCode)
     }
 }
 
+/// <summary>
+/// `llama_decode` return a non-zero status code
+/// </summary>
+public class MissingTemplateError
```
martindevans, line 64:
```suggestion
public class MissingTemplateException
```

Exception classes should end with `Exception` ([docs](https://learn.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#end-exception-class-names-with-exception)).

`RuntimeError` and `LLamaDecodeError` break this rule and should be fixed sometime!

LLama/LLamaTemplate.cs
```diff
@@ -250,6 +251,19 @@ public ReadOnlySpan<byte> Apply()
                 {
                     // Run templater and discover true length
                     var outputLength = ApplyInternal(_nativeChatMessages.AsSpan(0, Count), output);
+                    
+                    // if we have a return code of -1, the template was not found.
+                    if (outputLength == -1)
+                    {
+                        if (_customTemplate != null)
+                        {
+                            throw new MissingTemplateError(Encoding.GetString(_customTemplate));
```
martindevans, line 260:
```suggestion
                            throw new MissingTemplateException(Encoding.GetString(_customTemplate));
```

LLama/LLamaTemplate.cs
```diff
@@ -250,6 +251,19 @@ public ReadOnlySpan<byte> Apply()
...
+                        }
+                        else
+                        {
+                            throw new MissingTemplateError();
```
martindevans, line 264:
```suggestion
                            throw new MissingTemplateException();    
```

martindevans (changes requested):
Very minor naming nit, otherwise looks good :)

phil-scott-78:
i even felt a bit dirty typing it out, should have trusted my gut. Got it fixed up and pushed now.

martindevans (approved):

martindevans:
Thanks again!

PR #1036, opened 2025-01-04
January 2025 Update

martindevans:
Updated llama.cpp binaries to [0827b2c1da299805288abbd556d869318f2b121e](https://github.com/ggerganov/llama.cpp/commit/0827b2c1da299805288abbd556d869318f2b121e).

This introduces new binaries. Previously there were 2 for each platform, e.g. `ggml.dll, llama.dll` for CUDA. Now there are more: `ggml.dll, ggml-base.dll, ggml-cpu.dll, ggml-cuda.dll, llama.dll`.

Currently these are handled in the same way as the old system - each platform has it's own set of completely independent binaries. In the future this should be modified to dynamically load backends. See more details [here](https://github.com/ggerganov/llama.cpp/pull/10469/files#diff-201cbc8fd17750764ed4a0862232e81503550c201995e16dc2e2766754eaa57a) and [here](https://github.com/ggerganov/llama.cpp/pull/10256).

Testing:
- [x] Windows CPU
- [x] Windows CUDA
- [x] Windows Vulkan
- [x] Linux CPU
- [x] Linux CUDA
- [x] Linux Vulkan
- [ ] MacOS CPU
- [x] MacOS Metal


LSXAxeller:
Any chance this can be updated with the latest binaries? Mainstream got QRWKV Hybrid, Phi-3.5-MoE, Deepseek V3, C4AI Command-R 7B models, INT-8 and BF16 implementations.

martindevans:
The binaries here are from just one week ago, so hopefully most of those things should be covered.

Currently we need to fix the issues loading the binaries on Linux and MacOS, once that's figured out the next update should be easier 🤞

martindevans:
todo:
 - Update nuspec files for all backends

m0nsky:
Unit tests passed on Windows CUDA & Linux CUDA.

- Windows CPU
- Windows CUDA
- Windows Vulkan
- Linux CPU
- Linux CUDA
- Linux Vulkan

Lyrcaxis:
Works on my machine!
(osx-arm64 both partial and full offload)

PR #1038, opened 2025-01-08
Adding simple batch example

phil-scott-78:
I was really struggling to wrap my head around batched execution, got caught up in the weeds on an issue that turned out to be a bad prompt template in a gguf I was using. Anyways, I created a straight-forward "run these prompts as a batch" sample that was quite helpful to me while debugging, figured I'd toss it out there for inclusion with the samples if you see fit. If not, no worries, it served its value to me already.

https://github.com/user-attachments/assets/8b94f242-79af-4f02-93cd-55ae7995c666

PR #1045, opened 2025-01-17
Fix Batching Issues

martindevans:
In #1038 a new example was merged. However, this exposed a bug in the batching mechanism, sharing tokens when they shouldn't be shared. This PR completely removes the token sharing mechanism - tokens are only shared if the methods which explicitly add tokens to multiple sequences are used. While working on this I added some extra sanity checking to the batch, which I've left in place (`DEBUG` only).

I've also added more details explaining the `BatchedExecutorSimple` example, in particular explaining the NoKvSlot error which confuses a lot of people. @phil-scott-78 this may interest you :)


phil-scott-78:
Love it. New comments in the sample answered some questions I didn't even know I wanted to ask.

PR #1086, opened 2025-02-06
Update README.md

sangyuxiaowu:
Please use NativeLibraryConfig.All instead, or set configurations for NativeLibraryConfig.LLama and NativeLibraryConfig.LLavaShared respectively.


martindevans:
Thanks for fixing this!

PR #1098, opened 2025-02-13
Updates json gbnf to match llama.cpp example

phil-scott-78:
Example had a tendency to start just outputting whitespace until it ran out of tokens without a few of the constraints they added around whitespace generation. This gets it in line with example at https://github.com/ggerganov/llama.cpp/blob/master/grammars/json.gbnf


martindevans (approved):

martindevans:
Thanks for fixing this!

PR #1100, opened 2025-02-13
Add musl Compilation

Coopaguard:
- Add Musl compillation of LlamaCPP libs
- Detect Alpine OS
- package

Notes:
Vulkan & Cuda will not be supported for this time

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -76,6 +76,72 @@ jobs:
           name: llava-bin-linux-${{ matrix.build }}-x64.so
           if-no-files-found: error
 
+  compile-musl:
+    name: Compile (musl)
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - build: 'noavx'
+            defines: '-DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF'
+          - build: 'avx2'
+            defines: ''
+          - build: 'avx'
+            defines: '-DGGML_AVX2=OFF'
+          - build: 'avx512'
+            defines: '-DGGML_AVX512=ON'
+    runs-on: ubuntu-20.04
+    container:
+      image: alpine:latest
+    steps:
+      - name: Install dependencies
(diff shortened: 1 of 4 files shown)


martindevans:
Hi @Coopaguard, thanks for this PR. I don't personally know much at all about Musel or Alpine, could you explain a bit about what they're for? From a bit of research it looks like this would improve compatibility with Docker?

martindevans:
**Important Note**: BEFORE MERGING do a binary update that includes the new musl binaries in deps.zip!

Coopaguard:
> **Important Note**: BEFORE MERGING do a binary update that includes the new musl binaries in deps.zip!

Hi Martin,  I don't know how to do so.

Can't run the update Binaries action manually,
So I build it on my fork llamaSharp and get a deps.json file but don't how to add it to the projects through git.

Coopaguard:
> Hi @Coopaguard, thanks for this PR. I don't personally know much at all about Musel or Alpine, could you explain a bit about what they're for? From a bit of research it looks like this would improve compatibility with Docker?

Alpine Linux uses the musl C standard library instead of glibc, which is used by many other Linux distributions. Here's why it is often necessary to compile libraries with musl to use them on Alpine Linux.

Net8, Alpine & Docker 
When creating Docker containers for .NET 8 applications, Alpine Linux is often used as the base image due to its small size and security.

So yes this should allow us to create some docker images of LLama.WebAPI ou Lama.Web

martindevans:
> Hi @martindevans , I don't know how to do so.

Really sorry, I should have been clearerer there! I've updated the comment to clarify - that was a note to people reviewing/merging to do that before merging.

martindevans (approved):
Thanks for your work putting this together! I'll include this when I'm doing the next binary update in the coming weeks, and it should be included in the next version of LLamaSharp.

martindevans:
Changes have been merged into #1126 and will be merged when that PR is, so I'll close this one. Thanks for the hard work @Coopaguard ❤️

PR #1107, opened 2025-02-21
Unit tests that aim to verify the behavior and correctness of the sampling pipeline under various conditions

csharpconsultant:
This pull request adds several unit tests to the LlamaSharp library's SamplingTests.cs file. The added tests aim to verify the behavior and correctness of the sampling pipeline under various conditions. The tests include:

**SamplingWithTemperature:**

Tests the effect of changing temperature on the sampling process.

Compares low and high-temperature samples to ensure they are different.

**SamplingWithTopK:**

Tests the effect of TopK sampling with different values.

Ensures the sampled tokens and resulting text differ when using TopK=5 and TopK=50.

**FrequencyPenalty_ThrowsException_WhenValueIsLessThanMinusTwo:**

Verifies that an ArgumentOutOfRangeException is thrown when the frequency penalty is less than -2.

**FrequencyPenalty_ThrowsException_WhenValueIsGreaterThanTwo:**

Verifies that an ArgumentOutOfRangeException is thrown when the frequency penalty exceeds 2.

**PresencePenalty_ThrowsException_WhenValueIsLessThanMinusTwo:**

Verifies that an ArgumentOutOfRangeException is thrown when the presence penalty is less than -2.

**PresencePenalty_ThrowsException_WhenValueIsGreaterThanTwo:**

Verifies that an ArgumentOutOfRangeException is thrown when the presence penalty exceeds 2.

**DefaultValues_AreSetCorrectly:**

Verifies that the default values in the DefaultSamplingPipeline are set correctly.

**Seed_IsInitializedWithRandomValue:**

Verifies that the pipeline seed is initialized with a random value.

**Seed_IsInitializedWithSpecificValue:**

Verifies that the pipeline seed can be initialized with a specific value.

**SetMinKeep:**

Tests setting the MinKeep property to a specific value.

**GetMinKeepDefault:**

Verifies that the default value of the MinKeep property is set correctly.


LLama.Unittest/SamplingTests.cs
```diff
@@ -177,5 +178,231 @@ private static SafeLLamaSamplerChainHandle CreateChain(SafeLLamaContextHandle co
...
+        public void Seed_IsInitializedWithRandomValue()
+        {
+            // Arrange
+            var pipeline = new DefaultSamplingPipeline();
+
+            // Act
+            uint seed = pipeline.Seed;
+
+            // Assert
+            Assert.InRange(seed, 0u, uint.MaxValue);
```
martindevans, line 361: As far as I can see since `seed` is a `uint` it's impossible for this test to ever fail?

LLama.Unittest/SamplingTests.cs
```diff
@@ -177,5 +178,231 @@ private static SafeLLamaSamplerChainHandle CreateChain(SafeLLamaContextHandle co
...
+        [Fact]
+        public void SamplingWithTemperature()
+        {
+            using var context = new LLamaContext(_model, _params);
+            var tokens = _model.NativeHandle.Tokenize("The quick brown fox", false, false, Encoding.UTF8);
+
+            _batch.Add(token: tokens[0], pos: 0, sequence: LLamaSeqId.Zero, logits: true);
+            DecodeAndClear(context);
+
+            var logits = context.NativeHandle.GetLogits(numTokens: 1);
```
martindevans, line 193: The sampling part of this test could use _any_ logits, they don't actually have to come from an LLM. To avoid running the model more than necessary (which makes the test set very slow) could you just generate random logits (a big array of values between -1 and 1)? That also makes this test deterministic.

LLama.Unittest/SamplingTests.cs
csharpconsultant: Good point. What about testing for random seeds?  We could generate 1000 seeds, ensuring we don't have more than some defined number of collisions. The odds of getting two collisions in 1000 draws would be approximately 0.00032%.

LLama.Unittest/SamplingTests.cs
csharpconsultant: I removed the LLM and made the test deterministic.

LLama.Unittest/SamplingTests.cs
csharpconsultant: I removed this test and added a new one that confirms that the seed generation mechanism has a low probability of collision, ensuring reliable randomness in the sampling process.

csharpconsultant:
I made the following updates:

Removed the non-deterministic sampling with temperature test.
Made the TopK test deterministic.
Removed the uint seed range test that would always pass.
Added a test to verify that the DefaultSamplingPipeline generates seeds with a low collision probability.

PR #1110, opened 2025-02-26
Made Vocabulary's properties be initialized only ONCE on creation

Lyrcaxis:
Minor QOL improvement on the `Vocabulary` class with a slight performance increase.
Reason was I'm worried about llama.cpp internal deadlocks/mem. corruption based on an issue reported by @dpmm99.
image
I'm not 100% sure if this will solve the issue, but it's one less thing to worry about. Maybe we can narrow it down little by little.

### DIFF
--- a/LLama/Native/SafeLlamaModelHandle.cs
+++ b/LLama/Native/SafeLlamaModelHandle.cs
@@ -635,62 +635,60 @@ namespace LLama.Native
             /// <summary>
             /// Map of each token in this vocabulary to its string representation
             /// </summary>
-            internal readonly IReadOnlyDictionary<LLamaToken, string> TokenToString;
+            public readonly IReadOnlyDictionary<LLamaToken, string> TokenToString;
 
             /// <summary>
             /// Contains unique tokens that are supposed to end the generation (e.g.: EOS, EOT, etc)
             /// </summary>
-            internal readonly IReadOnlyList<LLamaToken> EOGTokens;
+            internal readonly HashSet<LLamaToken> EOGTokens;
 
             /// <summary>
             /// Contains unique tokens that exist for inference control rather than text output
             /// </summary>
-            internal readonly IReadOnlyList<LLamaToken> ControlTokens;
+            internal readonly HashSet<LLamaToken> ControlTokens;
 
-            internal Vocabulary(SafeLlamaModelHandle model)
+            internal unsafe Vocabulary(SafeLlamaModelHandle model)
             {
                 _model = model;
-                TokenToString = GetVocabCache();
 
                 // Cache the various properties that llama.cpp API exposes about the vocab
-                unsafe
-                {
-                    var vocabNative = llama_model_get_vocab(_model);
-                    Count = LLamaVocabNative.llama_vocab_n_tokens(vocabNative);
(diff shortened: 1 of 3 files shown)


Lyrcaxis:
rerun tests

LLama/Native/SafeLlamaModelHandle.cs
```diff
@@ -631,34 +632,65 @@ public sealed class Vocabulary
...
 
+            /// <summary>
+            /// Map of each token in this vocabulary to its string representation
+            /// </summary>
+            public readonly IReadOnlyDictionary<LLamaToken, string> TokenToString;
+
+            /// <summary>
+            /// Contains unique tokens that are supposed to end the generation (e.g.: EOS, EOT, etc)
+            /// </summary>
+            public readonly HashSet<LLamaToken> EOGTokens;
```
martindevans, line 643: publically exposing mutable collections isn't great. Would it be better just to make all three of these collections private, so there's no change to the public API, just the way it's implemented internally?

LLama/Native/SafeLlamaModelHandle.cs
```diff
@@ -669,232 +701,88 @@ internal Vocabulary(SafeLlamaModelHandle model)
...
-            {
-                get
-                {
-                    unsafe
-                    {
-                        return LLamaVocabNative.llama_vocab_n_tokens(VocabNative);
-                    }
-                }
-            }
+            public int Count { get; init; }
```
martindevans, line 704: I think these properties can all just be `{ get; }`?

martindevans:
@dpmm99 does this fix (or at least improve) the issue you were talking about in Discord?

LLama/Native/SafeLlamaModelHandle.cs
Lyrcaxis: Would something like this do? Or should I make them completely private?
image

dpmm99: Use FrozenSet or ImmutableHashSet?

dpmm99: Isn't it .NET Standard 2.0, not .NET Framework 2.0, that we're using? Both are supported there via a NuGet package.

Lyrcaxis: Oh, right! But it's still not available on .NET Standard 2.0 :P

martindevans: I think it's better just to make them private for this PR, since it's just meant to be fixing a bug. We can debate making them public (and what type to use) in another PR.

martindevans (approved):
Looks good 👍

dpmm99:
> @dpmm99 does this fix (or at least improve) the issue you were talking about in Discord?

Negative.

martindevans:
Since this didn't fix the bug shall we close this PR?

Lyrcaxis:
> Since this didn't fix the bug shall we close this PR?

I think it’s still a decent improvement — huge performance-wise depending on use case. But your call.

martindevans:
Oh if it's a large performance improvement as well let's merge it! I thought it was jsut a potential bugfix. Do you have any benchmarks for how much of a gain it is?

PR #1111, opened 2025-02-26
Cleaned up non-important warnings from the error list

Lyrcaxis:
LLamaSharp had [a lot of warnings that appear on each PR](https://github.com/SciSharp/LLamaSharp/pull/1110/files#file-llama-llamainteractexecutor-cs-L102), and bloat the error list.
Some of them are valid, but most of them are just nullables stuff, or missing documentations.
This PR takes care of the non-important warnings, bumping the visibility of current and future important ones.

For reference, here is how it looks on current master:
image

And here are the warnings that remain after this PR -- only obsolete warnings and warnings about await:
image

### DIFF
--- a/LLama/LLamaExecutorBase.cs
+++ b/LLama/LLamaExecutorBase.cs
@@ -395,7 +395,7 @@ namespace LLama
             public bool NeedToSaveSession { get; set; }
         }
 
-#pragma warning disable // Missing XML and irrelevant nullable warnings
+#pragma warning disable CS1591, CS8618 // Missing XML and irrelevant nullable warnings
         [JsonConverter(typeof(PolymorphicJSONConverter<ExecutorBaseState>))]
         public class ExecutorBaseState
         {
--- a/LLama/LLamaExecutorBase.cs
+++ b/LLama/LLamaExecutorBase.cs
@@ -395,6 +395,7 @@ namespace LLama
             public bool NeedToSaveSession { get; set; }
         }
 
+#pragma warning disable // Missing XML and nullable warnings that
         [JsonConverter(typeof(PolymorphicJSONConverter<ExecutorBaseState>))]
         public class ExecutorBaseState
         {
@@ -431,5 +432,6 @@ namespace LLama
             [JsonPropertyName("mirostat_mu")]
             public float? MirostatMu { get; set; }
         }
+#pragma warning restore
     }
 }
--- a/LLama/LLamaExecutorBase.cs
+++ b/LLama/LLamaExecutorBase.cs
@@ -395,7 +395,7 @@ namespace LLama
             public bool NeedToSaveSession { get; set; }
         }
 
-#pragma warning disable // Missing XML and nullable warnings that
+#pragma warning disable // Missing XML and irrelevant nullable warnings
         [JsonConverter(typeof(PolymorphicJSONConverter<ExecutorBaseState>))]
         public class ExecutorBaseState
         {
--- a/LLama/LLamaInstructExecutor.cs
+++ b/LLama/LLamaInstructExecutor.cs
@@ -25,8 +25,6 @@ namespace LLama
         private LLamaToken[] _inp_pfx;
         private LLamaToken[] _inp_sfx;
 
-        private ISamplingPipeline? _pipeline;
-
         /// <summary>
         /// 
         /// </summary>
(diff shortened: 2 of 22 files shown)


LLama/LLamaExecutorBase.cs
```diff
@@ -395,6 +395,7 @@ protected class InferStateArgs
             public bool NeedToSaveSession { get; set; }
         }
 
+#pragma warning disable // Missing XML and irrelevant nullable warnings
```
martindevans, line 398: I think it could be better to suppress the specific warnings, rather than _all_ warnings.

Lyrcaxis: Good catch ^^ There were so many warnings I might have accidentally lazied out a few

LLama/LLamaInteractExecutor.cs
```diff
@@ -69,15 +73,15 @@ public override Task LoadState(ExecutorBaseState data)
             if (data is InteractiveExecutorState state)
             {
                 _n_session_consumed = state.ConsumedSessionCount;
-                _embed_inps = state.EmbedInps.ToList();
+                _embed_inps = state.EmbedInps!.ToList();
```
martindevans, line 76: Same issue with suppressions as mentioned above.

LLama/LLamaInstructExecutor.cs
```diff
@@ -266,12 +264,12 @@ public class InstructExecutorState : ExecutorBaseState
             /// Instruction prefix tokens.
             /// </summary>
             [JsonPropertyName("inp_pfx")]
-            public LLamaToken[] InputPrefixTokens { get; set; }
+            public LLamaToken[]? InputPrefixTokens { get; set; }
```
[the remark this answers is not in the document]
Lyrcaxis: These were two separate warnings. This `?` here regards the `Should not leave constructor as null` warning.

LLama/Native/LLavaImageEmbed.cs
```diff
@@ -7,6 +7,13 @@ namespace LLama.Native;
 [StructLayout(LayoutKind.Sequential)]
 public unsafe struct LLavaImageEmbed
 {
+    /// <summary>
+    /// The embeddings of the embedded image. Should be a float[]
```
martindevans, line 11:
```suggestion
    /// The embeddings of the embedded image.
```
Span or Memory or any other buffer type is acceptable too!

LLama/Native/NativeApi.cs
```diff
@@ -18,11 +18,6 @@ public static void llama_empty_call()
             llama_max_devices();
         }
 
-        /// <summary>
-        /// Call once at the end of the program - currently only used for MPI
-        /// </summary>
-        public static extern void llama_backend_free();
```
martindevans, line 24: Please keep this (but feel free to make it private and suppress warnings about it being unused). When updating the binaries I try keep it one-to-one with llama.cpp, with nothing missing.

LLama/LLamaInstructExecutor.cs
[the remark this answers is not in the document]
Lyrcaxis: I think it's fine for it to be retrieved as null as long as it throws on the appropriate place (when it's used).
It deserializes from the file, so if it's null there, it sounds proper this class can keep them as nullable.

When using it, though, the `state.InputPrefixTokens!;` is a bit hacky indeed.
I think adding backing fields for this is a little overkill though.

Dunno. How would you suggest we approach this?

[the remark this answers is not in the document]
martindevans: I think you're right 👍

In this case any fixes to work around the nullability don't seem worth the extra work.

martindevans:
Thanks for doing this. We've been needing cleanup here for a long time!

LLama/LLamaInstructExecutor.cs
```diff
@@ -72,17 +70,17 @@ public override Task LoadState(ExecutorBaseState data)
...
                 _is_prompt_run = state.IsPromptRun;
                 _consumedTokensCount = state.ConsumedTokensCount;
-                _embeds = state.Embeds.ToList();
-                _last_n_tokens = new FixedSizeQueue<LLamaToken>(state.LastTokensCapacity, state.LastTokens);
-                _inp_pfx = state.InputPrefixTokens;
-                _inp_sfx = state.InputSuffixTokens;
+                _embeds = state.Embeds!.ToList();
+                _last_n_tokens = new FixedSizeQueue<LLamaToken>(state.LastTokensCapacity, state.LastTokens!);
+                _inp_pfx = state.InputPrefixTokens!;
+                _inp_sfx = state.InputSuffixTokens!;
```
GuyPago, line 79: I do have to say that you just made a lot of suppressing just to ignore the warnings, which might introduce some non indicative `NullReferenceException`s. I believe the warnings should rather remind to us provide a different behavior when its actually null, or at least provide some information to the user (or logger).

Lyrcaxis: I may have misunderstood the whole “nullables” concept, but from what I understand, they exist so developers can double check they indeed handle a possible null value properly before implementing it on a new system, and mark with “!” if they indeed did.

In reality, those values will NOT be null here based on all current implementations that the engine supports.

Let me know if you have a different view on the nullables, or if you have an alternative way to propose that handles this — or if you wanna try your hand at it I’d be happy to learn how you’ll handle this!

GuyPago: I’m thinking more long-term. If the implementation ever changes, or someone modifies the state-loading logic in the future, we might suddenly start seeing NullReferenceExceptions in unexpected places.

As for this specific comment context, `state.Embeds` is of  a non nullable type `LLamaToken[]` so I don't even think a suppression was necessary. As of `state.InputPrefixTokens`, if it can't be null in the current project implementation - why not just change its type to not nullable?

In general I agree with the inhouse double check approach you mentioned, I just hate to assume something will never happen so I made this comment because that was (almost) the only way you handled this warning across the PR, which made me wonder if a "double check" was made. Thank you for your contribution.

PR #1124, opened 2025-03-07
Added 'strict' parameter for template retrieval

Lyrcaxis:
This PR adds an additional check to the `LLamaTemplate` constructor and `Model.GetTemplate()`, causing it to throw an exception if a valid template was expected to be returned (`strict = true`) but could not be found.

This aims to avoid accidental confusion for users that try to use LLamaSharp's auto-template-formatting functionalities with a model whose template is either not supported or not found (e.g. base model vs instruct model).

The default behaviour can be disabled by passing `strict = false` when calling to retrieve the template.

### DIFF
--- a/LLama/Native/SafeLlamaModelHandle.cs
+++ b/LLama/Native/SafeLlamaModelHandle.cs
@@ -603,15 +603,24 @@ namespace LLama.Native
         /// Get the default chat template. Returns nullptr if not available
         /// If name is NULL, returns the default chat template
         /// </summary>
-        /// <param name="name"></param>
+        /// <param name="name">The name of the template, in case there are many or differently named. Set to 'null' for the default behaviour of finding an appropriate match.</param>
+        /// <param name="strict">Setting this to true will cause the call to throw if no valid templates are found.</param>
         /// <returns></returns>
-        public string? GetTemplate(string? name = null)
+        public string? GetTemplate(string? name = null, bool strict = true)
         {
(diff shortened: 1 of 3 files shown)


LLama/Native/SafeLlamaModelHandle.cs
```diff
@@ -603,15 +603,24 @@ internal IReadOnlyDictionary<string, string> ReadMetadata()
...
+        public string? GetTemplate(string? name = null, bool strict = true)
         {
             unsafe
             {
                 var bytesPtr = llama_model_chat_template(this, name);
                 if (bytesPtr == null)
-                    return null;
+                {
+                    if (strict)
+                        throw new Exception($"Tried to retrieve template for '{name}' but no templates were found.\n" +
```
martindevans, line 617: Can you use a specific exception type instead of a general `Exception`. Probably best to define a new exception in https://github.com/SciSharp/LLamaSharp/tree/master/LLama/Exceptions/.

Note: There is a `MissingTemplateException` already defined [here](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Exceptions/RuntimeError.cs#L64), but that's **not** the right thing to use (despite the name). That's for trying to use an unknown template with llama.cpp.

martindevans (changes requested):

martindevans (approved):
Looks good to me. Thanks!

PR #1155, opened 2025-04-05
docs: Added link to new package MaIN.NET

wisedev-pstach:
I’ve created a package that allows orchestrating agents, chat models, and RAG from different backends within .NET applications, using LlamaSharp for gaining local LLM capabilities as a local backend, and then allows it to be integrated with other LLM providers such as Ollama, Groq, OpenAI, etc. The aim of the package is to enable the creation of complex LLM (and other ml techniques in future i hope) workflows with minimal boilerplate, making it easier to experiment with different LLM integrations.

While the implementation is still a bit raw, I hope its already good-enough for prototyping and quick PoCs


martindevans:
Thanks :D

wisedev-pstach:
@martindevans my thanks, working with LLamaSharp was inspiring for me to learn a lot about LLMs

PR #1158, opened 2025-04-15
Fixed LLama.Web error starting session

### LINKED ISSUE #1080: [BUG]: LLamaSharp Web stopped working
### Discussed in https://github.com/SciSharp/LLamaSharp/discussions/1078

<div type='discussions-op-text'>

<sup>Originally posted by **hswlab** February  2, 2025</sup>
Has anyone gotten the latest version of the web project (0.21.0) to work yet? I downloaded the master repository in Visualstudio 2022, various files were downloaded when I rebuilt it, including some models in the LLama.Unittest\Models folder.  

The web project starts for me, but when I click on Begin Session, nothing happens, only the loading circle is visible. At first glance, there are no errors in the console output. 

I tried a slightly older project a few months ago and it worked straight away. Is it possible that some of the settings in appsettings.json are no longer up to date? For example, the preconfigured “llama-2-7b-chat.Q4_0.gguf” was not downloaded at all. There are only these models in my folder after project build.

image

image

Edit: I'm using a Win10 Machine with NVIDIA GeForce GTX 1070 and 4GHz i7-6700K CPU

...

Edit:

Oh, there seem to be a problem in a signalr function.
image

This Error happends in wwwroot\js\sessionConnectionChat.js if the following row is called:
image

I suppose, that usually the following Task in Hubs\SessionConnectionHub.cs should be called, but this is not happening for some reason. I can't tell why this signalR call is not working.
image

Are these params in "connection.invoke('LoadModel', sessionParams, sessionParams)" correct? It looks strange, because that "sessionParams" parameter is used twice.

EDIT:

I think someone forgot to set the second parameter “inferenceConfig” correctly. This also explains the TODO comment. If you compare the SessionConfig and InferenceOptions data types with each other, some properties are similar. The developer probably wanted to fill an InferenceOptions object at this point, but forgot to do so.

Does anyone know if this is still being developed? Unfortunately, the current state of the implementation prevents you from using the web application.</div>

tonib:
This should fix #1080

InferenceOptions.SamplingPipeline now is not required. Is not sent from client. This was the main issue for #1080
Added options to debug SignalR calls at appsettings.Development.json. If you dont have this enabled, SignalR serialization errors are not logged (even with Warning level)
Changed default web model to llava-v1.6-mistral-7b.Q3_K_XS.gguf, previous one (llama-2-7b-chat.Q4_0.gguf) did no exist

### DIFF
--- a/LLama.Web/Common/InferenceOptions.cs
+++ b/LLama.Web/Common/InferenceOptions.cs
@@ -20,6 +20,6 @@ namespace LLama.Web.Common
         public IReadOnlyList<string> AntiPrompts { get; set; } = Array.Empty<string>();
 
         /// <inheritdoc />
-        public required ISamplingPipeline SamplingPipeline { get; set; }
+        public ISamplingPipeline SamplingPipeline { get; set; } = new DefaultSamplingPipeline();
     }
 }
--- a/LLama.Web/appsettings.Development.json
+++ b/LLama.Web/appsettings.Development.json
@@ -3,7 +3,9 @@
   "Logging": {
     "LogLevel": {
       "Default": "Information",
-      "Microsoft.AspNetCore": "Warning"
+      "Microsoft.AspNetCore": "Warning",
+      "Microsoft.AspNetCore.SignalR": "Debug",
+      "Microsoft.AspNetCore.Http.Connections": "Debug"
     }
   }
 }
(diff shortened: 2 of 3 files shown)


martindevans:
Thanks for looking into this

martindevans:
Yeah I don't think that's your bug. I can't reproduce it locally either though :/

martindevans:
Even though the CI is failing I'll merge this one, I asked a few other maintainers and we're pretty confident the bug isn't in your changes.

PR #1170, opened 2025-04-23
Update LLamaEmbedder, Examples packages, and KernelMemory examples

zsogitbe:
- Embedding generation: Extension with Batch processing + Normalization (important  to have this built-in for KernelMemory).
- Examples had wrong nuget packages, updated to correct ones.
- Updated KernelMemory examples.

I have tested the code and it was fine with me. Please do test it at least once more. Thank you.

### DIFF
--- a/LLama/Native/NativeApi.cs
+++ b/LLama/Native/NativeApi.cs
@@ -290,6 +290,14 @@ namespace LLama.Native
         [DllImport(libraryName, CallingConvention = CallingConvention.Cdecl)]
         internal static extern void llama_kv_self_clear(SafeLLamaContextHandle ctx);
 
+        [Obsolete("Use `llama_kv_self_clear` instead")]
+        /// <summary>
+        /// Clear the KV cache. Both cell info is erased and KV data is zeroed
+        /// </summary>
+        /// <param name="ctx"></param>        
+        [DllImport(libraryName, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern void llama_kv_cache_clear(SafeLLamaContextHandle ctx);
+        
         /// <summary>
         /// Removes all tokens that belong to the specified sequence and have positions in [p0, p1)
         /// </summary>
--- a/LLama/Native/SafeLLamaContextHandle.cs
+++ b/LLama/Native/SafeLLamaContextHandle.cs
@@ -809,7 +809,8 @@ namespace LLama.Native
         /// </summary>
         public void KvCacheClear()
         {
-            NativeApi.llama_kv_self_clear(this);
+            //NativeApi.llama_kv_self_clear(this);
+            NativeApi.llama_kv_cache_clear(this);
         }
 
         /// <summary>
--- a/LLama/Native/SafeLLamaContextHandle.cs
+++ b/LLama/Native/SafeLLamaContextHandle.cs
@@ -809,8 +809,7 @@ namespace LLama.Native
         /// </summary>
         public void KvCacheClear()
         {
-            //NativeApi.llama_kv_self_clear(this);
-            NativeApi.llama_kv_cache_clear(this);
+            NativeApi.llama_kv_self_clear(this);
         }
 
         /// <summary>
--- a/LLama.Examples/Examples/KernelMemory.cs
+++ b/LLama.Examples/Examples/KernelMemory.cs
@@ -46,7 +46,7 @@ namespace LLama.Examples.Examples
 
             // Ask a predefined question
             Console.ForegroundColor = ConsoleColor.Green;
-            string question1 = "What formats does KM support";
+            string question1 = "What is Kernel Memory";
             Console.WriteLine($"Question: {question1}");
             await AnswerQuestion(memory, question1);
 
--- a/LLama.Examples/Examples/KernelMemorySaveAndLoad.cs
+++ b/LLama.Examples/Examples/KernelMemorySaveAndLoad.cs
@@ -54,7 +54,7 @@ public class KernelMemorySaveAndLoad
             await IngestDocuments(memory);
         }
 
-        await AskSingleQuestion(memory, "What formats does KM support?");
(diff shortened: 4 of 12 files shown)


LLama/Native/NativeApi.cs
```diff
@@ -290,6 +290,14 @@ public static void llama_log_set(NativeLogConfig.LLamaLogCallback logCallback)
         [DllImport(libraryName, CallingConvention = CallingConvention.Cdecl)]
         internal static extern void llama_kv_self_clear(SafeLLamaContextHandle ctx);
 
+        [Obsolete("Use `llama_kv_self_clear` instead")]
```
martindevans, line 293: This is exposed as `KvCacheClear` on `SafeLLamaContextHandle` now, it shouldn't be re-introduced here.

martindevans: Oh it's also an obsolete function in llama.cpp anyway!

LLama/Native/SafeLLamaContextHandle.cs
```diff
@@ -809,7 +809,8 @@ public int KvCacheCountTokens()
         /// </summary>
         public void KvCacheClear()
         {
-            NativeApi.llama_kv_self_clear(this);
+            //NativeApi.llama_kv_self_clear(this);
```
martindevans, line 812: Not sure why this change was made? `llama_kv_cache_clear` was renamed `llama_kv_self_clear` in llama.cpp recently, so that should always be used instead now.

martindevans (changes requested):
Overall looks pretty good, just a couple of minor changes needed with `llama_kv_self_clear`.

I haven't tested the embedder changes yet, they look reasonable and I'll try to make some time this weekend to test it out with long embeddings.

LLama/Native/SafeLLamaContextHandle.cs
zsogitbe: Yes, for some reason there are/were problems with the llama.cpp submodule. I had a slightly lower version (after downloading a version initially) and that version of llama.cpp does not have the llama_kv_self_clear yet. This was the reason for why I have reintroduced the old clear. I think that it would be better to keep it still in the code for this reason (it is marked as obsolete), but it is up to you to decide.

zsogitbe:
Test failed: It runs on my computer, I do not see any problem (nothing important has changed compared to before when all tests were OK). Try to restart the test. Thank you.

LLama/Native/SafeLLamaContextHandle.cs
martindevans: That was probably me forgetting to update the submodule again on the last update. If you keep an eye on the update PRs when they're available feel free to give me a poke if you spot that in the future.

For `llama_kv_cache_clear` vs `llama_kv_self_clear`, please put it back to the newer version. If it was a `public` API it'd make sense to keep it around for a while marked as `Obsolete` to give people a chance to migrate, but this is an `internal` function.

martindevans:
I've restarted the tests, but I'm not expecting them to pass right now. We seem to have an issue with one specific test at the moment that consistently fails in CI, but nobody can reproduce a failure locally :(

I'm probably just going to suppress that test if I get some time to work on it this weekend.

zsogitbe:
> I've restarted the tests, but I'm not expecting them to pass right now. We seem to have an issue with one specific test at the moment that consistently fails in CI, but nobody can reproduce a failure locally :(
> 
> I'm probably just going to suppress that test if I get some time to work on it this weekend.

OK! Take your time Martin, no problem. Thank you!

zsogitbe:
Martin, I have reworked the unit tests and experimented a bit with what is going wrong. I think that I have been able to correct several problems. The tests were switched off, could you please start the tests once more? Thank you.

zsogitbe:
I have corrected several problems in the code and I could make the macOSX tests pass (with skipping only 1), but the **Windows** and **Linux** KernelMemory tests (18) fail with cannot load the weights from disk:

`LLama.Exceptions.LoadWeightsFailedException : Failed to load model 'Models/Llama-3.2-1B-Instruct-Q4_0.gguf'`

I have no more ideas why the Windows and Linux KernelMemory tests are failing. **They are OK on my Windows computer (all tests pass)!**

One of the problems I have corrected is that I have completely rewritten the model download code (project file) because it was downloading the models every time it was compiling the code.

martindevans:
Could you split the csproj file downloading changes to a separate PR? The `SkipUnchangedFiles="true"` attribute should prevent duplicate downloads, so we'll want to look into that. Having it all in one PR will just slow down merging the embedder changes.

zsogitbe:
> Could you split the csproj file downloading changes to a separate PR? The `SkipUnchangedFiles="true"` attribute should prevent duplicate downloads, so we'll want to look into that. Having it all in one PR will just slow down merging the embedder changes.

That would be a lot of work Martin and I have invested already a lot of time in this (finding the problems and correcting them). It could also cause potential problems because of the split of the commits and then I would have even more work. Please try to check all together, if possible. You can easily test the download solution on your computer and the GitHub tests were successful on all platforms (the new code has found the downloaded models). I have tested the download on my computer with no models and also with existing models and it worked fine with me. Thank you. 

We will need to do some cleanup later (some commented out code) because some things are not very clear yet (for example why we do not need Embeddings=true anymore, etc.), but I would prefer to first integrate this PR without further changes, if possible. I have left these in to make sure that we do not forget to do some more research.

Thank you!

martindevans:
That's no problem, if it's not just a matter of cherry picking some commits to a new branch don't worry about it 👍

martindevans:
Sorry for the delay on this, I'd hoped to get the binary update PR in first but that's taking a long time!

zsogitbe:
> Sorry for the delay on this, I'd hoped to get the binary update PR in first but that's taking a long time!

Thank you Martin! Yes, I was expecting that this binary update will not be very simple (some key things changed... and it is an update with important improvements). Thank you so much that you do this!

PR #1138, opened 2025-03-27
add support for linux-arm64

nipeone:
#1137 I have completed the compile.yaml action, but I am unable to test other relevant code for the time being.

PR #1130, opened 2025-03-19
Feature/add android support

AmSmart:
# Description of Changes

- Updated compile Android to re-enable build and fix pipeline errors
- Added Android native libs to the CPU backend ``.nuspec`` file
- Updated ``.targets`` file to include new Android additions
- Added an extra condition to skip custom DLL resolution on Android because Android auto resolves libs based on generated apk
- Add Maui test Project (only Android enabled at the moment, will enable iOS in future PR)

Please let me know if any updates are required.

PR #1179, opened 2025-05-01
May Binary Update

martindevans:
Updated llama.cpp binaries to [ceda28ef8e310a8dee60bf275077a3eedae8e36c](https://github.com/ggml-org/llama.cpp/tree/ceda28ef8e310a8dee60bf275077a3eedae8e36c), compiled with [this run](https://github.com/martindevans/LLamaSharp/actions/runs/14763282855).

This PR includes work done by @nipeone in https://github.com/SciSharp/LLamaSharp/pull/1138 (adding Linux-ARM64 support) and by @AmSmart in https://github.com/SciSharp/LLamaSharp/pull/1130 (adding Android support).

Testing:
- [x] Windows CPU
- [x] Windows CUDA
- [x] Windows Vulkan
- [x] Linux CPU (x64)
- [ ] Linux CPU (ARM64)
- [ ] Linux CPU (musl)
- [x] Linux CUDA
- [x] Linux Vulkan
- [x] MacOS


martindevans:
@AmSmart looks like there are some installation issues for CI building the Android example. Not sure why they didn't show up before. Shall we just skip building the Android project in CI for now?

AmSmart:
@martindevans Probably fastest cause of action to skip. It ultimately depends on the planned release of this PR though. I've checked the errors and the issue is mostly minor, - I just need to set up the JDK so I can resolve it this weekend.

Also probably worth splitting out the build for Mobile into a separate job in the unit test workflow file. Things will be more optimal that way.

It is worth saying though, that we'll need to stop building the mobile workloads on Linux as .NET doesn't support MAUI on Linux at the moment. All mobile builds can be run in a single run with osx for example.

AmSmart:
I think the best course of action is to skip building Llama.Mobile for the desktop runs and I'll work on a PR to add a separate job that builds the Mobile project. I can work on that this weekend. What do you think?

BTW, how do you trigger the unit test runs? They don't seem to autorun when I update my branch.

AmSmart:
@Lyrcaxis The mobile project simply demonstrates success loading of the library. All the other samples in the repo apply to it just the same. It's fairly trivial to update the sample app a Q&A bot if we think that'll be nicer.

martindevans:
> BTW, how do you trigger the unit test runs? They don't seem to autorun when I update my branch.

Pushing to this PR auto trigger them. I _think_ that should work in a fork too. You may need to open a PR to your own master branch for the triggers to pick it up.

I've just done some work to exclude the mobile work on non-windows platforms, including removing it from the sln altogether. That's fixed Linux and MacOS. However Windows is still broken because it needs the JDK installed.

SignalRT:
Tested on MacOS.

My personal opinion is that the Mobile project should be on a different solution. It depends so much on specific SDK and your platform.

martindevans:
I've removed LLama.Mobile from the main solution for now. We can work out the best way to handle it in later PRs.

zsogitbe:
Martin, be careful with the [llama.cpp @ be7c303](https://github.com/ggerganov/llama.cpp/tree/be7c3034108473beda214fd1d7c98fd6a7a3bdf5)
You probably still need to update it. That is not [ceda28ef8e310a8dee60bf275077a3eedae8e36c](https://github.com/ggml-org/llama.cpp/tree/ceda28ef8e310a8dee60bf275077a3eedae8e36c) yet.

zsogitbe:
I have tested the code and it works well, also with new models. The binaries compiled from cpp are much smaller now (my ggml-cuda.dll is 45 MB only).

Thank you!

zsogitbe:
After further testing the KernelMemory example does not work anymore. There is a strange crash at `MemoryAnswer answer = await memory.AskAsync(question);` when it tries to retrieve the context size. Could you please test it on your computer?

martindevans:
Fixed the submodule, thanks for the reminder.

> I have tested the code and it works well
Which platforms did you test on? At the moment we've got an issue with the Linux CPU binaries failing for some people on Linux that we're having trouble narrowing down. So if you had success there that'd be an interesting datapoint.

martindevans:
> After further testing the KernelMemory example does not work anymore
I just tested `Kernel Memory: Document Q&A` with `Qwen3-30B-A3B Q4_K_M` and it worked for me, with some GPU load so it seems like it offloaded properly.

There aren't any config changes, just one extra field in the config struct which is always null. If you collapse `LLama.Mobile` in the diff you can see there actually aren't that many changes in this PR - it just looks big because that added a lot of files!

zsogitbe:
> Fixed the submodule, thanks for the reminder.
> 
> > I have tested the code and it works well
> 
> Which platforms did you test on? At the moment we've got an issue with the Linux CPU binaries failing for some people on Linux that we're having trouble narrowing down. So if you had success there that'd be an interesting datapoint.

Thank you!

I have found the problem. My modifications are not in your PR yet. They will probably get activated when you merge.

This is the bug I have identified before. The SplitMode BUG in llama.cpp. If you offload all layers to the GPU and use Native.GPUSplitMode.None the code will crash. I have showed you the C++ code before where this happens.

martindevans:
> If you off-load all layers (GpuLayerCount = -1) it will crash. See my remark above.

I just tried it with `GpuLayerCount = -1` and it still worked for me. However there was about 20GB of system memory in use as well as all of my VRAM, so maybe it spilled from VRAM into system RAM? I think that's something that can be turned off in the drivers, if you've got it off that'd explain the difference.

I just double checked the context size, since it's not specified in the config. If you trace that through KM it defaults to `2048` if none is specified so that's fine.

As a side note, there is some pretty suspect stuff happening [here](https://github.com/SciSharp/LLamaSharp/blob/master/LLama.KernelMemory/BuilderExtensions.cs#L70) though!
 - `context = weights.CreateContext(parameters);` creates a context
 - `var executor = new StatelessExecutor(weights, parameters);` internally creates a context
 - `new LLamaSharpTextEmbeddingGenerator(config, weights)` creates a context for embedding
 - `new LlamaSharpTextGenerator(weights, context, executor, config.DefaultInferenceParams)` accepts both an executor (with a context) and a context (but not necessarily the same context).

None of this has changed, so it's not the cause of your problem. But it does look like it could use a re-work! Would you be interested in working on that in the near future?

zsogitbe:
> I just tried it with `GpuLayerCount = -1` and it still worked for me. However there was about 20GB of system memory in use as well as all of my VRAM, so maybe it spilled from VRAM into system RAM? I think that's something that can be turned off in the drivers, if you've got it off that'd explain the difference.
> 
> I just double checked the context size, since it's not specified in the config. If you trace that through KM it defaults to `2048` if none is specified so that's fine.
> 
> As a side note, there is some pretty suspect stuff happening [here](https://github.com/SciSharp/LLamaSharp/blob/master/LLama.KernelMemory/BuilderExtensions.cs#L70) though!
> 
> * `context = weights.CreateContext(parameters);` creates a context
> * `var executor = new StatelessExecutor(weights, parameters);` internally creates a context
> * `new LLamaSharpTextEmbeddingGenerator(config, weights)` creates a context for embedding
> * `new LlamaSharpTextGenerator(weights, context, executor, config.DefaultInferenceParams)` accepts both an executor (with a context) and a context (but not necessarily the same context).
> 
> None of this has changed, so it's not the cause of your problem. But it does look like it could use a re-work! Would you be interested in working on that in the near future?

Yes, I think you're right. Many contexts are created and maintained everywhere. We either need to destroy them after use or ensure they are reusable. Since it is stateless, it can be leveraged in various situations. I'm particularly curious about the embedder's context, as it might require additional parameters. I'll try to make some time over the weekend to explore a more memory-efficient solution.

zsogitbe:
There is one way to do it: move `public int CountTokens(string text)` and `public IReadOnlyList<string> GetTokens(string text)` to the `StatelessExecutor` and access them from there. If this is OK for you, then I will try to implement it during the weekend!?

martindevans:
They both seems like thigns shouldn't need a context, can they be moved to the `LLamaWeights`? (btw let's move this discussion to a separate issue, since it's not related to this PR).

zsogitbe:
> They both seems like things shouldn't need a context, can they be moved to the `LLamaWeights`? (btw let's move this discussion to a separate issue, since it's not related to this PR).

I think that we better associate a context with the executor instead of the weights. I will try to propose something and then we can do the discussion there.

zsogitbe:
It was a good idea Martin! I have quickly put together a modification with streamlining contexts everywhere and it decreases GPU memory use with 30%! In my solution, both LLamaEmbedder and StatelessExecutor get their special CountTokens and GetTokens functions. These are being called and we can create the context on the fly and like this there is only 1 context in memory at all times.
I will clean it up and present the PR during the weekend.

martindevans:
Triggered a new build run, hopefully resolving the Linux CPU issues: https://github.com/martindevans/LLamaSharp/actions/runs/14956801082

PR #1183, opened 2025-05-10
Memory efficient context handling

zsogitbe:
Martin, in the LLamaEmbedderTests in CompareEmbeddings() I had to disable a Microsoft.Extensions.AI.IEmbeddingGenerator related code segment that does not work with the new efficient context handling. Please look at that code to decide what can be done to use it or we can also remove it. I guess that if we want that kind of functionality, then we will need to create a LLamaEmbedderService that is compatible with Microsoft.Extensions.AI.IEmbeddingGenerator.

I had to disable in SafeLlamaModelHandleTests -> MetadataValByKey_ReturnsCorrectly for Mac and Linux. On Windows it is OK. Please look at this also to decide if this is important. This has nothing to do with this PR, but I guess with llama.cpp.


LLama/LLamaEmbedder.cs
```diff
@@ -137,8 +150,54 @@ public async Task<IReadOnlyList<float[]>> GetEmbeddings(string input, Cancellati
...
 
         return (results, tokens.Length);
     }
+
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="text"></param>
+    /// <returns></returns>
+    public int CountTokens(string text)
```
martindevans, line 163: `CountTokens` and `GetTokens` methods are duplicated on `LLamaEmbedder` and `LLamaStatelessExecutor`. Also I don't think either of them actually requires a context (which is a very expensive object to create and destroy)!

Can these methods be moved up to `LLamaWeights` class instead? That's a more appropriate place for methods relating to tokens/vocabulary.

LLama/LLamaStatelessExecutor.cs
```diff
@@ -169,5 +169,44 @@ public async IAsyncEnumerable<string> InferAsync(string prompt, IInferenceParams
                     throw new LLamaDecodeError(returnCode);
             }
         }
+
+        /// <inheritdoc/>
+        public int CountTokens(string text)
```
martindevans, line 174: See other comment about these methods

martindevans (changes requested):

LLama/LLamaEmbedder.cs
zsogitbe: Martin, on a second thought, I think that you may be right (only the params need to be kept). I will look at it!

zsogitbe:
Moved the code to `LLamaWeights`. You were right Martin, it is better there.

LLama.Unittest/LLamaEmbedderTests.cs
```diff
@@ -42,37 +42,42 @@ private async Task CompareEmbeddings(string modelPath)
...
-        Assert.NotNull(generator.GetService<EmbeddingGeneratorMetadata>()?.DefaultModelId);
-        Assert.NotEmpty(generator.GetService<EmbeddingGeneratorMetadata>()?.DefaultModelId!);
-        Assert.Same(embedder, generator.GetService<LLamaEmbedder>());
-        Assert.Same(generator, generator.GetService<IEmbeddingGenerator<string, Embedding<float>>>());
-        Assert.Null(generator.GetService<string>());
-
-        var embeddings = await generator.GenerateAsync(
-        [
-            "The cat is cute",
+        if (false)
```
martindevans, line 45: Does this need resolving before merge?

LLama/LLamaWeights.cs
```diff
@@ -165,5 +166,35 @@ public LLamaToken[] Tokenize(string text, bool add_bos, bool special, Encoding e
...
+        /// Count the tokens in the input text
+        /// </summary>
+        /// <param name="text">input text</param>
+        /// <param name="parameters">context parameters</param>
+        /// <returns></returns>
+        public int CountTokens(string text, IContextParams parameters)
+        {
+            using var context = CreateContext(parameters);
+            var count = context.Tokenize(text, special: true).Length;
+            return count;
```
martindevans, line 180: There's a `Tokenize` method on the lower level model handle, no need to use a context: https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Native/SafeLlamaModelHandle.cs#L480

martindevans: (i.e. just writing a wrapper over that method should suffice)

LLama/LLamaWeights.cs
```diff
@@ -165,5 +166,35 @@ public LLamaToken[] Tokenize(string text, bool add_bos, bool special, Encoding e
...
+        /// <summary>
+        /// Get the list of tokens for the input text
+        /// </summary>
+        /// <param name="text">Input string to be tokenized</param>
+        /// <param name="parameters">Context parameters</param>
+        /// <returns>Read-only list of tokens for the input test</returns>
+        /// <remarks>
+        /// It throws if text is null and Includes empty stop token because addBos is left true to be consistent with the CountTokens implementation.</remarks>
+        /// <see cref="CountTokens(string, IContextParams)"/>
+        public IReadOnlyList<string> GetTokens(string text, IContextParams parameters)
```
martindevans, line 192: This implementation doesn't seem correct to me (I realise you're just moving it from `LlamaSharpTextGenerator` to `LLamaWeights`, but I don't really work on the KM stuff, so I haven't closely reviewed these methods before).

In general LLamaSharp is quite careful about never treating tokens as text, it's not safe for a number of reasons - for example a token could be half of a character, in which case it can't be decoded into text. That's what the `StreamingTokenDecoder` is for, you could add 10 tokens and get back just one character of text. At the very least, that means that GetTokens and CountTokens would have a mismatch.

Obviously KM needs _something_ back to satisfy the contract of `ITextTokenizer` etc, so I'm not really sure what the right answer is here. Maybe we should move back into KM, an an extension method on LLamaWeights? That way you can still use it as if it's here, but it's not part of the main lib. I'm open to other ideas though.

LLama.Unittest/LLamaEmbedderTests.cs
zsogitbe: `generator.GetService<EmbeddingGeneratorMetadata>()` uses the context and thus will fail because for a context efficient handling we do not keep the context. This was the main aim of this PR.  

The code in the test assumes that there is a context. I think that for the test code to work we would need some extra work to create an embedding service that keeps the context (this could be done in a next PR, if anybody is interested in to do it). The aim of the embedder in our code is different. My opinion is that the test code is wrong because it assume that the embedder is a live service, and it should not be for efficiently handling of GPU memory. There are two options, delete the test code or leave it in switched off with the TODO comment I have added.

LLama/LLamaWeights.cs
zsogitbe: I did not realize that there is a tokenize method on parent level. I will use that then and since it does not need a context I can move all code back to KM. The main aim of this PR was to remove the context that was create on several places because it unnecessarily fills GPU memory (saves about 30% of memory!).

LLama/LLamaWeights.cs
zsogitbe: The main aim of this PR was to remove the context that was created on several places in the KM code because it unnecessarily fills GPU memory (this saves about 30% of memory!). With using the native tokenize method and moving back the code to KM, I think that we have the right solution. Furthermore, I think that the code is only used in unit tests and probably will never be used by anybody. This is also the reason for why we should not keep a context there!

zsogitbe:
I have updated the code.

martindevans:
Just gave this a quick skim, looks pretty good after those last changes 👍 

I'll try to find time tomorrow to give it one alst in depth review and get it merged.

martindevans (approved):

martindevans:
Thanks for all the work on this!

zsogitbe:
> Thanks for all the work on this!

You too, Martin!

PR #1197, opened 2025-05-31
Conversation `-1` instead of `_end`

martindevans:
Using `-1` for start/end positions in `Conversation`. This eliminates the possibility of KV cache leaks due to bugs with `_end` getting out of sync.

### DIFF
--- a/LLama/Batched/Conversation.cs
+++ b/LLama/Batched/Conversation.cs
@@ -410,7 +410,7 @@ public sealed class Conversation
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


LLama/Batched/Conversation.cs
```diff
@@ -419,8 +419,7 @@ public void Remove(LLamaPos start, int count)
             if (count <= 0)
                 return;
 
-            var end = start.Value + count;
```
dpmm99, line 422: That would be breaking this Remove(start, count) API.

PR #1220, opened 2025-07-06
Multi modal (Just libraries)

SignalRT:
This are the changes to build and copy the right dynamic libraries as first step to introduce code changes.

@martindevans, CUDA compilation in Windows doesn´t work. If I see the problem correctly it requires CUDA 12.4. But I'm not sure, and I cannot test it on Windows.


martindevans:
It looks like llama.cpp no longer offers `CUDA 11.7` and `CUDA 12.2`, instead there are only builds for `CUDA 12.4`. Should we do the same?

martindevans:
Test run for CUDA 12.4 only started here: https://github.com/martindevans/LLamaSharp/actions/runs/16153517357, will PR that if it passes, then hopefully this PR should compile too

martindevans:
I've created a pull request to this branch (see [here](https://github.com/SignalRT/LLamaSharp/pull/106)). It changes over to CUDA 12.4 build.

martindevans:
I'm going to merge this and kick off a test build.

SignalRT:
@martindevans, I will be working on the changes during the next days.

martindevans:
The test run (https://github.com/SciSharp/LLamaSharp/actions/runs/16231321874) mostly passed, just a minor issue with file naming at the end. So we've got a working CUDA build for all your bit whenever they're ready :)

PR #1222, opened 2025-07-19
Fix for getting CUDA Version inside nvidia/cuda containers

Crelex:
Grab Version from Environment before parsing files on linux


martindevans:
Thanks!

PR #1256, opened 2025-09-23
Respect ChatOptions.Instructions in LlamaExecutorChatClient

stephentoub:


martindevans:
Thanks. The test failure looks like normal flakiness with the download so I'll merge this.

PR #1261, opened 2025-09-27
Mtmd implementation

SignalRT:
Prototype implementation:

- [x] Minimally tested on macOS. 
- [ ] Tested unsuccessfully with CUDA 13 (seems to be an issue in llama.cpp itself). 
- [ ] Unittest pending
- [ ] The test does not render images (commented out). 
- [ ] Audio and video handling not tested.


LLama/Batched/BatchedExecutor.cs
```diff
@@ -15,12 +16,8 @@ public sealed class BatchedExecutor
     : IDisposable
 {
     private int _nextSequenceId;
-    private readonly List<IBatch> _batchQueue = [];
```
martindevans: It looks like the changes from https://github.com/SciSharp/LLamaSharp/pull/1262 have been undone here? Probably an accident that requires rebasing?

LLama/Batched/Conversation.cs
```diff
@@ -21,6 +22,12 @@ public sealed class Conversation
     /// Indicates if this conversation has been "forked" and may share logits with another conversation.
     /// </summary>
     private bool _forked;
+    private readonly List<SafeMtmdEmbed> _mtmdEmbeds = new();
+    private int? _mtmdLogitsIndex;
```
martindevans: This only ever seems to be set to -1 or null?

LLama/Batched/Conversation.cs
```diff
@@ -206,6 +258,43 @@ private void AssertCanBePrompted()
...
+
+        if (Executor.ClipModel is null)
+            throw new InvalidOperationException("This conversation is not configured for multimodal prompts.");
+
+        var embed = Executor.ClipModel.LoadMedia(path);
+        _mtmdEmbeds.Add(embed);
+        _mtmdLogitsIndex = null;
+    }
+
+    public void QueueMedia(SafeMtmdEmbed embed)
```
martindevans: Could we add an overload of Prompt that takes a `Span<SafeMtmdEmbed>` instead of enqueuing them? This removes some state and makes the ownership of the embed objects clearer (e.g. at the moment you could call `QueueMedia`, then dispose the media, which would trigger an error on the next call to `Prompt` when you tried to use that disposed object).

LLama/Native/NativeApi.Mtmd.cs
```diff
@@ -0,0 +1,312 @@
...
+/// <summary>
+/// P/Invoke surface for MTMD (multimodal) helpers exposed by llama.cpp.
+/// </summary>
+public static partial class NativeApi
+{
+    /// <summary>
+    /// Convert a UTF-8 encoded native string pointer into a managed <see cref="string"/>.
+    /// Returns <c>null</c> when the pointer is zero.
+    /// </summary>
+    public static string? PtrToStringUtf8(IntPtr ptr)
```
martindevans: See other comment about duplicated code

LLama/SafeMtmdWeights.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using LLama.Native;
+
+namespace LLama;
+
+/// <summary>
+/// Lightweight wrapper around the MTMD native context and its helpers.
+/// </summary>
+public sealed class SafeMtmdWeights : IDisposable
```
martindevans: Rename to MtmdWeights for consistency with LLamaWeights?

Webslug:
Version 25.0 just breaks multi modal capabilities.

Qwen2.5-VL-3B won't work at all.  

How do we load the weights from other multimodal models?

System.DllNotFoundException: 'Unable to load DLL 'llava_shared' or one of its dependencies: The specified module could not be found. (0x8007007E)'

            string multiModalProj = "F:\\AI\\models\\Qwen2.5-VL-3B-Instruct-mmproj-f16.gguf";
            string modelPath = "F:\\AI\\models\\Qwen2.5-VL-3B-Instruct-q4_k_m.gguf";

            var parameters = new ModelParams(modelPath);
            NativeApi.llama_log_set((level, message) => { });
            Environment.SetEnvironmentVariable("LLAMA_LOG", "0");
            using var model = LLamaWeights.LoadFromFile(parameters);
            using var context = model.CreateContext(parameters);
            using var clipModel = LLavaWeights.LoadFromFile(multiModalProj);
            var executor = new InteractiveExecutor(context, clipModel, logger: null);

            var inferenceParams = new InferenceParams()
            {
                MaxTokens = 512,
                AntiPrompts = new List<string> { "\nUSER:" },
                SamplingPipeline = new DefaultSamplingPipeline { Temperature = 0.1f }
            };

LLama/Batched/BatchedExecutor.cs
SignalRT: I’ve rebased the branch.

SignalRT:
@martindevans , At this point, it would be helpful to have the final binaries so that we can perform a complete test and check for any issues with the rebase

m0nsky:
@SignalRT , we've updated the master branch with new binaries and a couple of fixes that should solve the macos/osx CI issues. If you update, the tests should be passing again!

SignalRT:
> @SignalRT , we've updated the master branch with new binaries and a couple of fixes that should solve the macos/osx CI issues. If you update, the tests should be passing again!

Branch updated; reviewing PR-breaking issues.

SignalRT:
> @SignalRT , we've updated the master branch with new binaries and a couple of fixes that should solve the macos/osx CI issues. If you update, the tests should be passing again!

Is it just me, or is the llama.cpp module not updated to the new binary version?

m0nsky:
I think it's updated? It seems to point to `86587da` on master, which is what the latest binaries were built with (but we recently rebuilt them with support for `macos-14` and `DCMAKE_BUILD_WITH_INSTALL_RPATH`)

SignalRT:
Now seems to be working with this llama.cpp binaries: https://github.com/ggml-org/llama.cpp/releases/tag/b7679
I tested the solution only in osx for now. 

I tested audio and video (image) models:

- gemma-3-4b-it-Q4_K_M.gguf
- Qwen2.5-Omni-3B-Q8_0.gguf
- Qwen2.5-VL-7B-Instruct-Q8_0.gguf
- Voxtral-Mini-3B-2507-Q4_K_M.gguf
- Qwen3-VL-2B-Instruct-Q8_0.gguf
- Qwen3VL-8B-Instruct-Q8_0.gguf

martindevans (approved):
LGTM! This is a huge amount of work, thanks very much @SignalRT ❤️

zsogitbe:
@[SignalRT](https://github.com/SignalRT) Thank you very much, huge work! ❤️

@[SignalRT](https://github.com/SignalRT) & [martindevans](https://github.com/martindevans) could you please explain the consequences? We all have llava_shared.dll now linked and using that with the former version of the library. Did you implement this as a breaking change, or as an extra option? What are the implications? Does this only change the internal working like in InteractiveExecutor, or also the interface to the outside world? Thank you!

SignalRT:
In recent updates, llama.cpp removed the LLaVA libraries, and the interface in the mtmd libraries has been reworked.
As a result, the integration points changed and the previous API surface no longer matches, so the interface breaks for existing code that depended on the old layout.

The good news is that the required changes are relatively small and mostly mechanical, even though the interface looks different after the refactor. You can see the adjustments reflected in:
	•	the updated examples (they show the new recommended usage patterns), and
	•	the [PR]for the web interface (it demonstrates the same capabilities end-to-end, and makes the supported functionality explicit).

So even though the interface changed enough to break compatibility, the precise code changes needed to adapt are limited, and the examples + the web UI PR provide a reference for the new flow and features.

zsogitbe:
Thank you, [SignalRT](https://github.com/SignalRT)!

I looked at the  [Examples](https://github.com/SciSharp/LLamaSharp/tree/master/LLama.Examples/Examples)/MtmdInteractiveModeExecute.cs and it seems that the only important changes we need are the following (please correct me if I’m wrong):

```csharp
using var clipModel = await MtmdWeights.LoadFromFileAsync(multiModalProj, model, mtmdParameters);
```

and

```csharp
var mediaMarker = mtmdParameters.MediaMarker ?? NativeApi.MtmdDefaultMarker() ?? "<media>";
```

[martindevans](https://github.com/martindevans), I think we need some kind of platform where breaking updates and major additions are explained by the people who introduce them, ideally alongside the PRs. Without this, users will install a new update, their code may break, or they may miss important new features.

PR #1262, opened 2025-09-28
Optimization - A queue with fixed storage size backed by a circular buffer

SignalRT:
This should provide some optimizations to the FixedQueue. My test are:

**BEFORE THE CHANGES:**

image

**AFTER THE CHANGES:**

image

### DIFF
--- a/LLama/Batched/BatchedExecutor.cs
+++ b/LLama/Batched/BatchedExecutor.cs
@@ -19,6 +19,7 @@ public sealed class BatchedExecutor
     private int _batchQueueHead;
     private int _batchedTokenCount;
     private bool _batchedTokenCountDirty = true;
+    private const int CleanupThreshold = 16;
     
     /// <summary>
     /// Set to 1 using interlocked exchange while inference is running
@@ -216,7 +217,7 @@ public sealed class BatchedExecutor
                 return;
             }
 
-            if (_batchQueueHead > 16 && _batchQueueHead > _batchQueue.Count / 2)
+            if (_batchQueueHead > CleanupThreshold && _batchQueueHead > _batchQueue.Count / 2)
             {
                 _batchQueue.RemoveRange(0, _batchQueueHead);
                 _batchQueueHead = 0;
--- a/LLama/Batched/BatchedExecutor.cs
+++ b/LLama/Batched/BatchedExecutor.cs
@@ -19,6 +19,7 @@ public sealed class BatchedExecutor
     private int _batchQueueHead;
     private int _batchedTokenCount;
     private bool _batchedTokenCountDirty = true;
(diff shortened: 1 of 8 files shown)


LLama/Common/FixedSizeQueue.cs
```diff
@@ -85,15 +122,35 @@ IEnumerator IEnumerable.GetEnumerator()
...
+            if (start + count <= Capacity)
+            {
+                return new ReadOnlySpan<T>(_buffer, start, count);
+            }
+
+            _window ??= new T[Math.Min(Capacity, Math.Max(4, count))];
+            if (_window.Length < count)
+            {
+                Array.Resize(ref _window, Math.Min(Capacity, Math.Max(_window.Length * 2, count)));
+            }
```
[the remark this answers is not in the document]
SignalRT: Solved

LLama/Batched/BatchedExecutor.cs
```diff
@@ -16,7 +15,12 @@ public sealed class BatchedExecutor
     : IDisposable
 {
     private int _nextSequenceId;
-    private readonly List<IBatch> _batchQueue = [ ];
+    private readonly List<IBatch> _batchQueue = [];
```
martindevans, line 18: This has introduced a lot of complexity to do with managing the queue and the cached token count into `BatchedExecutor`, it could probably be factored out into other classes in the future.

LLama/Common/FixedSizeQueue.cs
```diff
@@ -6,75 +6,117 @@
...
             Capacity = size;
-            _storage = new();
+            _buffer = new T[size];
+            _start = 0;
+            _count = 0;
         }
 
         /// <summary>
-        /// Fill the quene with the data. Please ensure that data.Count &lt;= size
+        /// Fill the queue with existing data. Please ensure that data.Count <= size
```
martindevans, line 63:
```suggestion
        /// Fill the queue with existing data. Please ensure that data.Count &lt;= size
```

LLama/Common/FixedSizeQueue.cs
```diff
@@ -83,17 +125,40 @@ IEnumerator IEnumerable.GetEnumerator()
             return GetEnumerator();
         }
 
+        /// <summary>
+        /// Returns up to <paramref name="count"/> of the most recent items as a contiguous span.
+        /// </summary>
         internal ReadOnlySpan<T> AsSpan(int count)
```
martindevans, line 131: `AsSpan` doesn't seem to be used anywhere, so this could be removed if you want. (I have no objection if you would prefer to keep it).

martindevans (approved):
LGTM, just a couple of minor nits

PR #1274, opened 2025-10-05
Fix some warnings.

Lamothe:


LLama/LLamaExecutorBase.cs
```diff
@@ -262,7 +262,7 @@ protected virtual void TryReuseMatchingPrefix()
         /// <param name="inferenceParams"></param>
         /// <param name="args"></param>
         /// <returns></returns>
-        protected abstract Task<(bool, IReadOnlyList<string>)> PostProcess(IInferenceParams inferenceParams, InferStateArgs args);
```
martindevans, line 265: This is a breaking change for anything inheriting from executor base. So it needs to stay async.

martindevans:
Thanks!

Issue #1277, opened 2025-10-10
[BUG]: macOS ARM64 Native Library Loading Failure


jchristn:
### Description

**IMPORTANT** Claude Code helped write the issue summary, so please forgive any factual inaccuracies that I couldn't personally verify.  I couldn't find a duplicate bug other than issue #1276 but I assume that might be Unity-specific.

## Summary

**Issue**: LlamaSharp.Backend.Cpu v0.25.0 fails to load on macOS ARM64 due to broken `@rpath` references in the native library dependencies.

**Impact**: Could not initialize on macOS ARM64 without manual post-build fixes.

### System Information
```
macOS Version: Sonoma 15.6
Architecture: arm64 (Apple Silicon M2 Max)
Processor: Apple M2 Max
.NET Version: 9.0.8
.NET RID: osx-arm64
```

### Package Versions
```xml
<PackageReference Include="LLamaSharp" Version="0.25.0" />
<PackageReference Include="LLamaSharp.Backend.Cpu" Version="0.25.0" />
```

### Observed Behavior

When attempting to use LlamaSharp on macOS ARM64, the application fails during native library initialization with the following exception:

```
System.TypeInitializationException: The type initializer for 'LLama.Native.NativeApi' threw an exception.
 ---> LLama.Exceptions.RuntimeError: The native library cannot be correctly loaded. It could be one of the following reasons:
1. No LLamaSharp backend was installed. Please search LLamaSharp.Backend and install one of them.
2. You are using a device with only CPU but installed cuda backend. Please install cpu backend instead.
3. One of the dependency of the native library is missed. Please use `ldd` on linux, `dumpbin` on windows and `otool`to check if all the dependency of the native library is satisfied. Generally you could find the libraries under your output folder.
4. Try to compile llama.cpp yourself to generate a libllama library, then use `LLama.Native.NativeLibraryConfig.WithLibrary` to specify it at the very beginning of your code.

   at LLama.Native.NativeApi..cctor()
   at LLama.Native.NativeApi.llama_max_devices()
```

### Expected Behavior

The native library should load successfully when:
1. LLamaSharp.Backend.Cpu NuGet package is installed
2. Application is built and published
3. `NativeLibraryConfig.All.WithLibrary()` is called with the correct path

This works correctly on Windows x64.  I have yet to test on Linux.

---

### Issue 1: Broken @rpath References

The `libllama.dylib` library in the NuGet package uses `@rpath` to reference its dependencies:

```bash
[...]

PR #1279, opened 2025-10-13
set RPATH to "@loader_path" / "$ORIGIN" to ensure executables and dynamic libraries search for dependencies in their origin directory

SignalRT:
Linked with the issue:
https://github.com/SciSharp/LLamaSharp/issues/1277#issue-3501172244

I seems that should be done something similar on LINUX (not done in this PR)

[Reference llama.cpp PR](https://github.com/ggml-org/llama.cpp/pull/14309#issue-3164332869)

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -460,11 +460,13 @@ jobs:
       matrix:
         include:
           - build: 'arm64'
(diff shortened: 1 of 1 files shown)


jchristn:
Hi @SignalRT many thanks - please pardon the dumb question (not familiar with you and the team's approach to releases) - is this something slated for a particular version and if so any idea when I should be on the lookout?  If not I understand.  Thank you for your time!

Issue #1281, opened 2025-10-16
[BUG]: IndexOutOfRangeException with structured GBNF grammar output


SerialKicked:
### Description

Hi,

I think i've tried about everything to get grammar to work properly to no avail. I have read the docs and looked at the issues, but for some reason I can't click the checkboxes. docs/Examples/GrammarJsonResponse.md is outdated btw.

I'm using 0.25 on a Stateless executor: [Here's a link to the adapter i wrote](https://github.com/SerialKicked/Lethe-AI-Sharp/blob/master/Adapters/LlamaSharpAdapter.cs) 

The same grammar that works on KoboldCpp always gives me an error when running LLamaSharp instead. A *System.IndexOutOfRangeException: 'Index was outside the bounds of the array.'* to be precise.

I'm probably doing something wrong, but I really don't see what. I tried with different Grammar Optimization values out of curiosity, but nope. Oh, and yeah, normal generation works perfectly fine. 

Any help would be super welcome because I managed to get structured output to work for both KoboldAPI, and OpenAI-compatible backends. So, that'd very neat to get feature parity for LLamaSharp.

### Reproduction Steps

**So, example:**

I have a BasicBitch class
```csharp
public class BasicBitch
{
    public string Response { get; set; } = string.Empty;
    public int Confidence { get; set; } = 0;
}
```

The simplest possible JSON Schema -> GBNF conversion gives:

````
root ::= "{" space Response-kv "," space Confidence-kv "}" space
Response ::= string | null
Response-kv ::= "\"Response\"" space ":" space Response
Confidence-kv ::= "\"Confidence\"" space ":" space integer

char ::= [^"\\\x7F\x00-\x1F] | [\\] (["\\bfnrt] | "u" [0-9a-fA-F]{4})
integer ::= ("-"? integral-part) space
integral-part ::= [0] | [1-9] [0-9]{0,15}
null ::= "null" space
number ::= ("-"? integral-part) ("." [0-9]+)? ([eE] [-+]? [0-9]+)? space
space ::= | " " | "\n"{1,2} [ \t]{0,20}
string ::= "\"" char* "\"" space
````

Finally (ignore the implied data):
```csharp
var pipeline = new DefaultSamplingPipeline()
{
    Temperature = (float)input.Temperature,
    TopP = (float)input.Top_p,
    TopK = input.Top_k,
    TypicalP = (float)input.Typical,
    MinP = (float)input.Min_p,
[...]

PR #1282, opened 2025-10-19
Fixes Extended grammar optimization mode would crash if TopK sampling is set to zero.

SerialKicked:
Fixes the crash I identified in #1281 between grammar optimization and TopK sampling. In summary, now the `GrammarOptimizationMode.Extended` code runs only if TopK is set to something above zero.


martindevans (approved):

martindevans:
Thanks for fixing this. I'm not sure what's wrong the the docs updater action, but I don't think it's anything wrong with this PR.

SerialKicked:
You're very welcome, I'm just happy I can run structured output properly :)

(And yeah, I doubt it has anything to do with my change)

Issue #1299, opened 2025-12-05
Adding Jinja support to LLamaSharp


vvdb-architecture:
### Description

I've just finished implementing a Jinja interpreter, actually a re-implementation of the complete language accepted by  [Google's Minja project](https://github.com/google/minja/blob/main/include/minja/minja.hpp) 

It's fully-managed C# code in a single assembly. There are also XUnit tests. It supports a variety of methods to do bindings for the Jinja templates, including anonymous types, value types and Json elements.

Jinja support is currently missing, so I'd like to add this to the list of assemblies provided by LLamaSharp.

The reason why I'm hesitating is because I don't see a simple way to integrate this further. 
As far as I understand the architecture, there is [LLamaTemplate](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/LLamaTemplate.cs) which is called by `ToModelPrompt()` (a static function), so the only way to integrate Jinja capability is via a complete re-implementation of `ILLamaExecutor`. Is this the way to proceed?

I can of course just add the assembly as a first PR while I'm figuring out the best way to proceed. 
Let me know.

Vincent

vvdb-architecture:
Abandoned.

PR #1300, opened 2025-12-09
Added Jinja template processor

vvdb-architecture:
This adds a Jina template processor, language compatible with  [Google's minja project](https://github.com/google/minja/blob/main/include/minja/minja.hpp) as outlined in #1299 

The unit tests have been added as well, and show what can be supported and the kind of contexts that can be used.

Basically, it's a 3-step process:
1. Parse the template string into a `TemlpateNode` instance. This can be done only once (cache it if you want to re-use it)
2. Convert the bindings you supplied into a `Context` instance
3. Render the template node using the context

The absolute minimum method is:

~~~csharp
    private static string Render(string templateStr, object? bindings, Options? options = null)
    {
        var root = Parser.Parse(templateStr, options ?? Options.Default); //1
        var context = Context.Make(bindings); //2
        return root.Render(context); //3
    }
~~~

The `bindings` can be either an user defined type, an anonymous type, a dictionary mapping strings to objects, or one or more  `JsonElement` instances. 
For example, given:

~~~csharp
    private class TestBinding1
    {
        [JsonPropertyName("z")]
        public int[][] Z { get; } = [[1, 10], [2, 20]];
    }
~~~

The following are equivalent:

~~~csharp
        // binding using anonymous type
        Assert.Equal("1,10;2,20;", Render(@"    
{%- for x, y in z -%}
        {{- x }},{{ y -}};
{%- endfor -%}
", new { z = new[] { [1, 10], new int[] { 2, 20 } } }, null));

        // binding using dictionary
        Assert.Equal("1,10;2,20;", Render(@"    
{%- for x, y in z -%}
        {{- x }},{{ y -}};
{%- endfor -%}
", new Dictionary<string, object>() { { "z", new[] { [1, 10], new int[] { 2, 20 } } } }, null));

        // binding using explicit type
        Assert.Equal("1,10;2,20;", Render(@"    
{%- for x, y in z -%}
        {{- x }},{{ y -}};
{%- endfor -%}
", new TestBinding1(), null));

        // binding using JsonElement
        Assert.Equal("1,10;2,20;", Render(@"    
{%- for x, y in z -%}
        {{- x }},{{ y -}};
{%- endfor -%}
", JsonSerializer.SerializeToElement(new TestBinding1()), null));

        // binding using JSon string
        Assert.Equal("1,10;2,20;", Render(@"    
{%- for x, y in z -%}
        {{- x }},{{ y -}};
{%- endfor -%}
", JsonSerializer.Deserialize<JsonElement>("{\"z\":[[1,10],[2,20]]}"), null));

~~~

### DIFF
--- a/LLama.Jinja/ArgumentsExpression.cs
+++ b/LLama.Jinja/ArgumentsExpression.cs
@@ -51,7 +51,7 @@ internal sealed class ArgumentsExpression
         sb.Append("Args: [");
         for (var i = 0; i < Args.Count; i++)
         {
-            sb.Append(Args[i].ToString());
+            sb.Append(Args[i]);
             if (i < Args.Count - 1)
                 sb.Append(", ");
         }
@@ -60,7 +60,7 @@ internal sealed class ArgumentsExpression
         {
             sb.Append(KwArgs[i].Name);
             sb.Append(": ");
-            sb.Append(KwArgs[i].Expression.ToString());
+            sb.Append(KwArgs[i].Expression);
             if (i < KwArgs.Count - 1)
                 sb.Append(", ");
         }
--- a/LLama.Jinja/ArgumentsValue.cs
+++ b/LLama.Jinja/ArgumentsValue.cs
@@ -2,8 +2,8 @@ namespace LLamaSharp.Jinja;
 
 internal class ArgumentsValue
 {
-    public List<Value> Args;
-    public List<(string Name, Value Value)> Kwargs;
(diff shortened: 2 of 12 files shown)


martindevans:
This is a huge amount of new code that would be a huge maintanence burden, is it possible to build this as external package that extends LLamaSharp, rather than as a core part of LLamaSharp? If necessary we can modify the core package to add hooks that your external package uses.

vvdb-architecture:
Abandoned

PR #1316, opened 2026-01-19
 End‑to‑end multimodal chat with document parsing, media uploads, audio recording, and streaming markdown rendering

SignalRT:
**Summary:**
  This PR delivers a full multimodal chat pipeline in LLama.Web: PDF and Word document ingestion with text extraction, image and audio uploads, native in‑browser audio recording (preview/attach/discard), plus streaming response
  rendering with Markdown support.

  **Key Features:**

  - Streaming chat responses rendered incrementally.
  - Markdown rendering in the UI (including code blocks, lists, etc.).
  - Multimodal inference pipeline with MTMD support wired into session execution.
  - PDF ingestion with text extraction and truncation safeguards.
  - Word (DOCX) ingestion with text extraction from document XML.
  - Image uploads supported end‑to‑end (validation, storage, rendering in chat).
  - Audio uploads supported end‑to‑end (validation, storage, playback in chat).
  - In‑browser audio recording (MediaRecorder) with preview + attach/discard workflow.
  - Capability‑aware UI (shows whether text/vision/audio are supported per model).
  - Download models automatically and shows the progress

  Implementation Highlights

  - Attachment service handles file validation, storage, and extraction (PDF/DOCX).
  - Model session builds prompts with attached media and enforces capability checks.
  - Chat UI renders images/audio and guides users on supported inputs.
  - Captures audio and converts it to a browser file for existing upload flow.
  - Streaming tokens update the UI while Markdown is rendered on the fly.

**Capability to upload images and ask about the images**

image

**Model auto-download + Capability to upload files and ask about the files**
image


martindevans:
One thing that I'm not sure about is the media queue in the `SafeMtmdModelHandle`. Why is it an implicit queue instead of an explicit parameter passed into the tokenize call?

Alternatively, if it is necessary for some reason, could it be moved up one layer into the MtmdModel, instead of SafeModelModelHandle? That way the SafeHandle remains a minimal wrapper around llama.cpp, with additional behaviour added for convenience at the higher level wrapper.

martindevans:
Other than that one comment, looks good to me!

SignalRT:
> One thing that I'm not sure about is the media queue in the `SafeMtmdModelHandle`. Why is it an implicit queue instead of an explicit parameter passed into the tokenize call?
> 
> Alternatively, if it is necessary for some reason, could it be moved up one layer into the MtmdModel, instead of SafeModelModelHandle? That way the SafeHandle remains a minimal wrapper around llama.cpp, with additional behaviour added for convenience at the higher level wrapper.

That's a convenience API.

So my preference would be:

1. Keep explicit media passing as the primary API.
2. Treat the implicit queue as optional convenience only.
3. Move that convenience up out of SafeMtmdModelHandle if we keep it at all.

martindevans:
@SignalRT There are a couple of merge conflicts, once those are resolved I think this is ready to merge?

PR #1347, opened 2026-03-04
Fix musl detection using RID instead of distro name

m0nsky:
Changes `alpine` to `musl` in the NativeLibraryUtils.

The runtime identifier on musl based distros (for example, alpine) is linux-musl-x64, not alpine-*, so the previous StartsWith("alpine") check failed, causing the loader to switch to linux-x64. Musl tests are passing after this change (tested on my fork).


martindevans:
Thanks!

PR #1354, opened 2026-03-06
Remove setting hardcoded value for n_seq_max in IContextParamsExtensions

Ed-Pavlov:
It's already set from passed argument `result.n_seq_max = @params.SeqMax;`

Overwriting it with hardcoded value is removed
result.n_seq_max = Math.Min(Math.Max(10U, result.n_ctx / 8U), 64U /*0x40*/);

Ref: https://github.com/SciSharp/LLamaSharp/discussions/1353


martindevans:
Thanks!

Issue #1336, opened 2026-02-19
[BUG]: Qwen3 VL + HandleRunOutOfContext()


aropb:
### Description

The error occurs when filling in the context.

Tried the options InferenceParams:
TokensKeep: -1, 0, value

Code:
...
InteractiveExecutor executor = new(context, logger);
...
await foreach (string text in session.ChatAsync(new ChatHistory.Message(AuthorRole.User, query), false, inferenceParams, cancellationToken))
{
}

Critical error:

D:\a\LLamaSharp\LLamaSharp\src\llama-kv-cache.cpp:398: GGML_ASSERT(hparams.n_pos_per_embd() == 1 && "seq_add() is only supported for n_pos_per_embd() == 1") failed

### Environment & Configuration

- Operating system: Windows 11
- .NET runtime version:  NET 10.0.3
- LLamaSharp version: 0.26.0
- CUDA version (if you are using cuda backend): 12.9
- CPU & GPU device:  GPU RTX 5090

aropb:
More details, the error occurs on the model (VL models):
Qwen3-VL-30B-A3B-Instruct-Q6_K_L.gguf
Qwen3-VL-30B-A3B-Instruct-Q4_K_M.gguf <- official Qwen3 VL 
Qwen3-VL-8B-Instruct-F16.gguf <- official Qwen3 VL 

But this one doesn't have it (text models):
Qwen3-30B-A3B-Instruct-2507-Q6_K_L.gguf

Mtmd is not used!

LLamaSharp Binaries:
https://github.com/SciSharp/LLamaSharpBinaries/releases/tag/506bb6e01009058f355

aropb:
the error occurs here (class StatefulExecutorBase):

**Context.NativeHandle.MemorySequenceAdd(LLamaSeqId.Zero, tokensToKeep + n_discard, _pastTokensCount, -n_discard);**

Execution data:

            Context.NativeHandle.MemorySequenceRemove(LLamaSeqId.Zero, 3488, 5584);
            Context.NativeHandle.MemorySequenceAdd(LLamaSeqId.Zero, 5584, 7680, -2096); 


        protected virtual void HandleRunOutOfContext(int tokensToKeep)
        {
            // if we run out of context:
            // - take the tokensToKeep first tokens from the original prompt (via n_past)
            // - take half of the last (n_ctx - tokensToKeep) tokens and recompute the logits in batches
            var n_left = _pastTokensCount - tokensToKeep;
[...]

PR #1373, opened 2026-04-19
Context Overflow Fix

zsogitbe:
Hi @martindevans and @aropb,

Thank you for digging into this issue. The proposed solution (`MemoryClear` + Re-prefilling) in https://github.com/SciSharp/LLamaSharp/issues/1336 correctly identifies how to prevent the engine from crashing on models that return `MemoryCanShift == false` (like Qwen 3.5 with 2D RoPE embeddings).

However, after reviewing how the upstream `llama.cpp` ecosystem handles this, I’d like to propose an architectural modification in this PR. 

### The Core Issue
The truncation and shifting logic (like `--no-context-shift`) does not actually exist in the core `llama.h` C API; it is application-level logic built into `llama-cli` (`main.cpp`). Because LLamaSharp's `StatelessExecutor` and `InteractiveExecutor` act as our application-layer equivalents, implementing the fix here is structurally correct. 

However, hardcoding `n_discard = n_left / 2;` forces an extremely opinionated behavior. Silently amputating 50% of the KV cache when it fills up introduces two major risks for developers:
1.  **Semantic Destruction:** Blindly dropping exactly half of the tokens can cut code blocks, JSON payloads, or sentences in half, causing instant hallucinations.
2.  **Silent Latency Spikes:** The developer will experience a massive, multi-second TTFT (Time To First Token) delay while `DecodeAsync` re-evaluates the remaining thousands of tokens, with no explicit warning as to why the delay happened.

### The Proposed Compromise
Instead of making silent truncation the default behavior, we should implement a `ContextOverflowStrategy` on `IInferenceParams`. 

* **Default:** `ThrowException`. If the model's math blocks shifting, we fast-fail. This ensures enterprise developers know they hit a limit and must implement proper RAG/Summarization. (This mimics the `llama-cli` `--no-context-shift` flag).
* **Opt-In:** `TruncateAndReprefill`. We run the PR's `MemoryClear` logic, but we allow the developer to configure the `ContextTruncationPercentage` (defaulting to 10% instead of 50% to prevent excessive context loss). 

### Explanation of Implementation Choices
1.  **`TokensToKeep` Indexing:** In the original PR, `RemoveRange` used `tokensToKeep - 1`. This was structurally unsafe. If `tokensToKeep` was `0` (e.g., no system prompt), it would result in a `-1` index and instantly crash the application. The new implementation safely calculates the `startIndex`.
2.  **`Math.Clamp` on Percentage:** A user might accidentally pass `1.5f` (150%) or `0.0f` to the truncation percentage. The implementation guarantees the math will always drop *some* tokens but never *all* tokens, preventing infinite loops.
3.  **`ContextOverflowException`:** By subclassing a specific exception, developers can wrap their generation loops in a `try/catch (ContextOverflowException)` block to specifically catch and handle this event with custom semantic summarization.

### What's Included in this PR
* Added `ContextOverflowStrategy` enum and `ContextOverflowException`.
* Updated `IInferenceParams` and `InferenceParams` with the new configuration properties.
* Updated `StatefulExecutorBase.cs` to respect the new strategy.
* Updated `LLamaStatelessExecutor.cs` to try native memory shifting first, and safely catch the `MemoryCanShift` exception to trigger the fallback logic.
* Added/Updated Unit Tests to verify the exact overflow boundaries for both Stateless and Stateful executors.

### Test Output
To ensure this is robust, I tested the exact context boundaries dynamically based on what `llama.cpp` allocates at runtime. Both executors now successfully intercept the overflow and throw the correct exception before native crashes occur:

```text
[DEBUG] Actual ContextSize: 256, Prompt length: 254
Successfully caught expected exception: The context window is full and the current strategy 
is set to ThrowException. To automatically truncate and manage context, set 
InferenceParams.OverflowStrategy to ContextOverflowStrategy.TruncateAndReprefill.

Successfully caught expected exception in InstructExecutor: The context window is full and 
the current model architecture does not support native memory shifting. To automatically 
truncate and re-prefill the context, set InferenceParams.OverflowStrategy to 
ContextOverflowStrategy.TruncateAndReprefill.
```

Looking forward to hearing your thoughts on this approach!

### DIFF
--- a/LLama/Abstractions/IInferenceParams.cs
+++ b/LLama/Abstractions/IInferenceParams.cs
@@ -1,5 +1,6 @@
-using System.Collections.Generic;
+using LLama.Common;
 using LLama.Sampling;
+using System.Collections.Generic;
 
 namespace LLama.Abstractions
 {  
@@ -36,5 +37,19 @@ namespace LLama.Abstractions
 		/// Controls the behavior of decoders like <see cref="StreamingTokenDecoder" />
 		/// </remark>
 		public bool DecodeSpecialTokens { get; set; }
-	}
+
+        /// <summary>
+        /// Defines the strategy the executor should use when the context window is full 
+        /// and the model architecture (e.g., models with 2D RoPE embeddings) does not 
+        /// support native memory shifting.
+        /// </summary>
+        ContextOverflowStrategy OverflowStrategy { get; set; }
+
+        /// <summary>
+        /// The percentage of past tokens to discard when <see cref="OverflowStrategy"/> 
+        /// is set to <see cref="ContextOverflowStrategy.TruncateAndReprefill"/>. 
+        /// For example, 0.1f represents dropping the oldest 10% of the conversational context.
+        /// </summary>
+        float ContextTruncationPercentage { get; set; }
+    }
 }
--- a/LLama/Exceptions/ContextOverflowException.cs
+++ b/LLama/Exceptions/ContextOverflowException.cs
@@ -7,7 +7,33 @@ namespace LLama.Exceptions
     /// cannot mathematically support native memory shifting, or when the 
     /// ContextOverflowStrategy.ThrowException is used.
     /// </summary>
-    public class ContextOverflowException() : Exception("The context window is full and the current strategy is set to ThrowException. To automatically truncate and manage context, set InferenceParams.OverflowStrategy to ContextOverflowStrategy.TruncateAndReprefill.")
+    public class ContextOverflowException : Exception
     {
+        private const string DefaultMessage = "The context window is full and the current strategy is set to ThrowException. To automatically truncate and manage context, set InferenceParams.OverflowStrategy to ContextOverflowStrategy.TruncateAndReprefill.";
+
+        /// <summary>
+        /// Initializes a new instance of the ContextOverflowException class with a default error message.
+        /// </summary>
+        public ContextOverflowException() : base(DefaultMessage)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the ContextOverflowException class with a specified error message.
+        /// </summary>
+        /// <param name="message">The message that describes the error.</param>
+        public ContextOverflowException(string message) : base(message)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the ContextOverflowException class with a specified error message 
+        /// and a reference to the inner exception that is the cause of this exception.
+        /// </summary>
+        /// <param name="message">The message that describes the error.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception.</param>
+        public ContextOverflowException(string message, Exception innerException) : base(message, innerException)
+        {
+        }
     }
 }
--- a/LLama/Exceptions/ContextOverflowException.cs
+++ b/LLama/Exceptions/ContextOverflowException.cs
@@ -1,0 +1,14 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LLama.Exceptions
+{
+    /// <summary>
+    /// Thrown when the KV cache context is full and the model architecture 
+    /// cannot mathematically support native memory shifting.
+    /// </summary>
+    public class ContextOverflowException(string message) : Exception(message)
+    {
+    }
+}
--- a/LLama/Exceptions/ContextOverflowException.cs
+++ b/LLama/Exceptions/ContextOverflowException.cs
@@ -1,14 +1,13 @@
 using System;
-using System.Collections.Generic;
-using System.Text;
 
 namespace LLama.Exceptions
 {
     /// <summary>
     /// Thrown when the KV cache context is full and the model architecture 
-    /// cannot mathematically support native memory shifting.
+    /// cannot mathematically support native memory shifting, or when the 
+    /// ContextOverflowStrategy.ThrowException is used.
     /// </summary>
-    public class ContextOverflowException(string message) : Exception(message)
+    public class ContextOverflowException() : Exception("The context window is full and the current strategy is set to ThrowException. To automatically truncate and manage context, set InferenceParams.OverflowStrategy to ContextOverflowStrategy.TruncateAndReprefill.")
     {
     }
 }
--- a/LLama/LLamaExecutorBase.cs
+++ b/LLama/LLamaExecutorBase.cs
@@ -239,6 +239,13 @@ namespace LLama
             Context.NativeHandle.MemorySequenceAdd(LLamaSeqId.Zero, tokensToKeep + n_discard, _pastTokensCount, -n_discard);
             _pastTokensCount -= n_discard;
 
+            // Stop saving the session if we run out of context. 
+            // Note: A more advanced (but riskier and more complex) solution would be to physically trim 
+            // the _session_tokens list and adjust _n_session_consumed to perfectly match the newly 
+            // shifted native memory. This would allow session saving to continue safely, but requires 
+            // precise index tracking to avoid off-by-one errors. For now, we abort saving to prevent corruption.
+            _pathSession = string.Empty;
+
             return Task.CompletedTask;
         }
 
--- a/LLama/LLamaExecutorBase.cs
+++ b/LLama/LLamaExecutorBase.cs
@@ -200,47 +200,56 @@ namespace LLama
         /// </summary>
         /// <param name="tokensToKeep">The number of tokens from the initial prompt to preserve (e.g., system prompt).</param>
         /// <param name="inferenceParams">The parameters controlling the inference and overflow strategy.</param>
-        /// <exception cref="ContextOverflowException">Thrown when the overflow strategy is set to ThrowException or truncation bounds are invalid.</exception>
+        /// <exception cref="ContextOverflowException">Thrown when the overflow strategy is set to ThrowException.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when tokensToKeep is invalid.</exception>
         /// <exception cref="LLamaDecodeError">Thrown if the native context decoding fails during the re-prefill phase.</exception>
         protected virtual async Task HandleRunOutOfContext(int tokensToKeep, IInferenceParams inferenceParams)
         {
             // 1. Fast Fail if not configured to auto-truncate
             if (inferenceParams.OverflowStrategy == ContextOverflowStrategy.ThrowException)
             {
-                throw new ContextOverflowException(
-                    "The context window is full and the current model architecture does not support native memory shifting. " +
-                    "To automatically truncate and re-prefill the context, set InferenceParams.OverflowStrategy to ContextOverflowStrategy.TruncateAndReprefill."
-                );
+                throw new ContextOverflowException();
             }
 
             // 2. Calculate tokens safely
             var n_left = _pastTokensCount - tokensToKeep;
             if (n_left <= 0)
             {
-                throw new ContextOverflowException("Cannot truncate context: tokensToKeep exceeds or equals the current context size.");
+                throw new ArgumentOutOfRangeException(nameof(tokensToKeep), "Cannot truncate context: tokensToKeep exceeds or equals the current context size.");
             }
 
             // Clamp the percentage between 1% and 99% to prevent math errors or total wipeouts
             var percentage = Math.Max(0.01f, Math.Min(0.99f, inferenceParams.ContextTruncationPercentage));
             var n_discard = (int)(n_left * percentage);
 
-            // Sanity check: always discard at least 1 token if we are truncating
-            if (n_discard < 1) n_discard = 1;
+            // Sanity check: always discard at least 1 token, but never more than we have available.
+            n_discard = Math.Max(1, Math.Min(n_discard, n_left));
 
             // 3. Remove the oldest non-kept tokens. 
-            // If tokensToKeep is 10, we keep indexes 0-9, and start removing from index 10.
             int startIndex = Math.Max(0, tokensToKeep);
             _session_tokens.RemoveRange(startIndex, n_discard);
 
-            // 4. Clear the native KV Cache and perform a complete re-prefill
-            LLamaBatch batch = new();
-            Context.NativeHandle.MemoryClear();
-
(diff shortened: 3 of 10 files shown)


martindevans:
I really like this idea, it's much better than hardcoding a specific behaviour.

I haven't had a chance to review this in depth yet - but my first thought on reading is: do you think it might be feasible to factor the code handling the overflow out to a separate class? Something like an `IOverflowHandler` interface, with `OverflowThrows` and `OverflowShifts` implementations built into LLamaSharp. That way end users can add completely custom context overflow handling.

(Note: If you don't think that's feasible or you simply don't want to work on it that's fine! I'm happy to go ahead with reviewing this as-is if you prefer).

LLama/LLamaStatelessExecutor.cs
```diff
@@ -138,32 +141,66 @@ public async IAsyncEnumerable<string> InferAsync(string prompt, IInferenceParams
...
-                    n_past -= n_discard;
+                    try
+                    {
+                        // First, attempt the fast native memory shift (works for standard models like Llama 2/3)
+                        Context.NativeHandle.MemorySequenceRemove(LLamaSeqId.Zero, tokensKeep, tokensKeep + n_discard);
+                        Context.NativeHandle.MemorySequenceAdd(LLamaSeqId.Zero, tokensKeep + n_discard, n_past, -n_discard);
+                        n_past -= n_discard;
+                        all_tokens.RemoveRange(tokensKeep, n_discard);
+                    }
+                    catch (Exception ex) when (ex.Message.Contains("MemoryCanShift"))
```
martindevans: I think `Context.NativeHandle.MemoryCanShift` should be a better way to detect this case.

aropb:
> I haven't had a chance to review this in depth yet - but my first thought on reading is: do you think it might be feasible to factor the code handling the overflow out to a separate class? Something like an `IOverflowHandler` interface, with `OverflowThrows` and `OverflowShifts` implementations built into LLamaSharp. That way end users can add completely custom context overflow handling.

It seems redundant to me. Alternatively, you can make the method virtual.

aropb:
I usually calculate the size of the context as the sum of the prompt + reserve for the response (2-4К). TokensKeep = -1, so that the prompt is saved when the context is shifted. In this case, the 50% shift works well. A lower value will result in a more frequent shift.

zsogitbe:
> I really like this idea, it's much better than hardcoding a specific behaviour.
> 
> I haven't had a chance to review this in depth yet - but my first thought on reading is: do you think it might be feasible to factor the code handling the overflow out to a separate class? Something like an `IOverflowHandler` interface, with `OverflowThrows` and `OverflowShifts` implementations built into LLamaSharp. That way end users can add completely custom context overflow handling.
> 
> (Note: If you don't think that's feasible or you simply don't want to work on it that's fine! I'm happy to go ahead with reviewing this as-is if you prefer).

Glad that you like the approach! Concerning your `IOverflowHandler` idea Martin, I think that I tend to agree with aropb, because the reason for handling the overflow would be, for example, to create a summary, then rerun the prompt, and this handler would not be the right place to do that logic, I think, but correct me if I am wrong. In any case, even if it is decided to do this `IOverflowHandler` interface, I would like to suggest to do this in a follow-up PR. let us focus now on making this logic perfect first.

> I think Context.NativeHandle.MemoryCanShift should be a better way to detect this case.

Yes, we can change that to a Context.NativeHandle.MemoryCanShift test instead of catching the exception.

@martindevans, @aropb, I will wait until you did a full review and test and you give all of your remarks, before adjusting the PR.

LLama/Exceptions/ContextOverflowException.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace LLama.Exceptions
+{
+    /// <summary>
+    /// Thrown when the KV cache context is full and the model architecture 
+    /// cannot mathematically support native memory shifting.
+    /// </summary>
+    public class ContextOverflowException(string message) : Exception(message)
```
martindevans: The `message` should be built here, instead of needing to be passed in wherever the exception is throw.

LLama/Exceptions/ContextOverflowException.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LLama.Exceptions
+{
+    /// <summary>
+    /// Thrown when the KV cache context is full and the model architecture 
+    /// cannot mathematically support native memory shifting.
```
martindevans: Also mention that this may be thrown if `ContextOverflowStrategy.ThrowsException` is used.

LLama/LLamaExecutorBase.cs
```diff
@@ -198,21 +198,50 @@ public void SaveSessionFile(string filename)
...
+                    "To automatically truncate and re-prefill the context, set InferenceParams.OverflowStrategy to ContextOverflowStrategy.TruncateAndReprefill."
+                );
+            }
+
+            // 2. Calculate tokens safely
             var n_left = _pastTokensCount - tokensToKeep;
-            var n_discard = n_left / 2;
+            if (n_left <= 0)
+            {
+                throw new ContextOverflowException("Cannot truncate context: tokensToKeep exceeds or equals the current context size.");
```
martindevans: This feels like it should be a different kind of exception, since it's a misconfiguration of tokensToKeep. Maybe `ArgumentOutOfRangeException`.

LLama/LLamaExecutorBase.cs
```diff
@@ -198,21 +198,50 @@ public void SaveSessionFile(string filename)
...
+            }
+
+            // Clamp the percentage between 1% and 99% to prevent math errors or total wipeouts
+            var percentage = Math.Max(0.01f, Math.Min(0.99f, inferenceParams.ContextTruncationPercentage));
```
martindevans, line 231: Might want to add another safety check here that there's at least one token left and one token of space.

LLama/LLamaExecutorBase.cs
```diff
@@ -198,21 +198,50 @@ public void SaveSessionFile(string filename)
...
-            _pastTokensCount -= n_discard;
-            // stop saving session if we run out of context
-            _pathSession = string.Empty;
+            // 3. Remove the oldest non-kept tokens. 
+            // If tokensToKeep is 10, we keep indexes 0-9, and start removing from index 10.
+            int startIndex = Math.Max(0, tokensToKeep);
+            _session_tokens.RemoveRange(startIndex, n_discard);
+
+            // 4. Clear the native KV Cache and perform a complete re-prefill
+            LLamaBatch batch = new();
```
martindevans: Can this do the same branching as stateless, where it either shifts (as before) or clears and prefills (as now) depending on if the model is capable of shifting. Most models can be shifted, and it is vastly cheaper than rerunning prefill!

martindevans (changes requested):

martindevans:
> Concerning your IOverflowHandler idea

There are three main things I want to address with the idea of overflow strategies:

1. User extensibility - New overflow handling strategies can be implemented without extending the library or creating a new executor type.
2. Configuration - At the moment there's no place to store per-strategy configuration, so we end up storing it on the InferenceParams and simply documenting that some properties don't do anything except in certain cases.
3. Code duplication - At the moment the complex shifting behaviour is implemented twice (BaseExecutor and StatelessExecutor) and is not compatible at all with the BatchedExecutor. If it were moved out to overflow strategy classes that code wouldn't be deuplicated. There's an added bonus here that the executor classes have become a complex mess - factoring any code out of them is a win!

> In any case, even if it is decided to do this IOverflowHandler interface, I would like to suggest to do this in a follow-up PR. let us focus now on making this logic perfect first.

Agreed 👍

zsogitbe:
> > Concerning your IOverflowHandler idea
> 
> There are three main things I want to address with the idea of overflow strategies:
> 
> 1. User extensibility - New overflow handling strategies can be implemented without extending the library or creating a new executor type.
> 2. Configuration - At the moment there's no place to store per-strategy configuration, so we end up storing it on the InferenceParams and simply documenting that some properties don't do anything except in certain cases.
> 3. Code duplication - At the moment the complex shifting behaviour is implemented twice (BaseExecutor and StatelessExecutor) and is not compatible at all with the BatchedExecutor. If it were moved out to overflow strategy classes that code wouldn't be deuplicated. There's an added bonus here that the executor classes have become a complex mess - factoring any code out of them is a win!
> 
> > In any case, even if it is decided to do this IOverflowHandler interface, I would like to suggest to do this in a follow-up PR. let us focus now on making this logic perfect first.
> 
> Agreed 👍

@martindevans, this is a fantastic idea Martin, and a textbook use case for the Strategy pattern. It cleanly solves the issue of overloading `InferenceParams`, removes the duplicated fallback logic between `StatelessExecutor` and `StatefulExecutorBase`, and makes the API much more extensible for users who want to build custom overflow handlers. 

The main piece of work here will be getting the internal "plumbing" right. Because `StatelessExecutor` (which uses `List<LLamaToken> all_tokens`) and `StatefulExecutorBase` (which tracks `_session_tokens` and `_embed_inps`) manage their token histories differently, the `IOverflowHandler` interface will need to be designed carefully. It has to safely mutate those different history states and update the `n_past` count without breaking the internal tracking of either executor. 

p.s.: I have updated the code above!

martindevans (approved):

zsogitbe:
I have corrected the bugs found by Copilot. Unfortunately, I had to disable the MemoryCanShift = false path in the StatefulExecutorBase in HandleRunOutOfContext because as Copilot pointed out we need to track _history_tokens... This will be a future PR.

**Explanation**: I am holding off on implementing the manual history tracking fix in this PR for StatefulExecutorBase. Rebuilding the KV cache manually for models that don't support native memory shifting is highly complex and risks silent state corruption, especially when dealing with session caching and multimodal inputs. Keeping the ContextOverflowException guard as-is forces the calling application to handle the overflow gracefully (e.g., starting a new chat) rather than risking a desynced model state. We can revisit this if we see high user impact, but stability comes first.

martindevans:
Thanks for fixing that feedback, it all looks ok to me 👍 

Could you create an issue for that todo linking this PR and explaining the details of the work required? I'll mark it as do-not-close.

martindevans:
copilot has raised a couple of issues, but I can't assess them since I don't fully understand the internals of the higher level executors. Feel free to dismiss them if you think they're invalid.

zsogitbe:
> copilot has raised a couple of issues, but I can't assess them since I don't fully understand the internals of the higher level executors. Feel free to dismiss them if you think they're invalid.

I will look at them and apply the necessary changes if needed! This is a specialized model trained on similar code, so the comments may be relevant.

OK! Added the corrections.

LLama/LLamaStatelessExecutor.cs
```diff
@@ -138,32 +141,70 @@ public async IAsyncEnumerable<string> InferAsync(string prompt, IInferenceParams
                 tokens.Add(id);
 
                 // when run out of context
-                // based on this logic: https://github.com/ggerganov/llama.cpp/blob/master/examples/main/main.cpp#L497
                 if (n_past + tokens.Count >= Context.ContextSize)
                 {
+                    if (inferenceParams.OverflowStrategy == ContextOverflowStrategy.ThrowException)
+                    {
+                        throw new ContextOverflowException();
+                    }
```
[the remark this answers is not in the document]
zsogitbe: This is standard continuous batching behavior and closely mirrors how llama.cpp handles it. Moving the overflow check before sampling is unnecessary and would likely complicate your generation loop without providing any tangible benefit.

LLama/LLamaExecutorBase.cs
```diff
@@ -198,21 +198,48 @@ public void SaveSessionFile(string filename)
...
+            n_discard = Math.Max(1, Math.Min(n_discard, n_left));
 
+            // 4. Fast path: attempt the fast native memory shift
             Context.NativeHandle.MemorySequenceRemove(LLamaSeqId.Zero, tokensToKeep, tokensToKeep + n_discard);
             Context.NativeHandle.MemorySequenceAdd(LLamaSeqId.Zero, tokensToKeep + n_discard, _pastTokensCount, -n_discard);
-
             _pastTokensCount -= n_discard;
-            // stop saving session if we run out of context
-            _pathSession = string.Empty;
+
```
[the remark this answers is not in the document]
zsogitbe: OK! We will stop saving the session if we run out of context.

LLama/Abstractions/IInferenceParams.cs
```diff
@@ -36,5 +37,19 @@ public interface IInferenceParams
...
+        /// Defines the strategy the executor should use when the context window is full 
+        /// and the model architecture (e.g., models with 2D RoPE embeddings) does not 
+        /// support native memory shifting.
+        /// </summary>
+        ContextOverflowStrategy OverflowStrategy { get; set; }
+
+        /// <summary>
+        /// The percentage of past tokens to discard when <see cref="OverflowStrategy"/> 
+        /// is set to <see cref="ContextOverflowStrategy.TruncateAndReprefill"/>. 
+        /// For example, 0.1f represents dropping the oldest 10% of the conversational context.
```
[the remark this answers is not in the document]
zsogitbe: Not doing this now.

LLama/LLamaStatelessExecutor.cs
```diff
@@ -138,32 +141,70 @@ public async IAsyncEnumerable<string> InferAsync(string prompt, IInferenceParams
...
                     var tokensKeep = inferenceParams.TokensKeep;
 
                     // number of tokens to keep when resetting context
-                    // Ported from https://github.com/ggerganov/llama.cpp/blob/60325fa56f61c228464c9f065db3aa6a61f2156e/examples/main/main.cpp#L334
-                    if (tokensKeep < 0 || tokensKeep > tokens.Count)
+                    if (tokensKeep < 0 || tokensKeep > all_tokens.Count)
                     {
-                        tokensKeep = tokens.Count;
+                        tokensKeep = all_tokens.Count;
                     }
```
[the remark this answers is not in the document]
zsogitbe: OK! I will correct this.

martindevans:
Thanks for investigating those review items.

PR #1386, opened 2026-05-11
Implement Sequence ID pooling in BatchedExecutor to prevent native SeqMax overflow and crashes

zsogitbe:
### Description

**The Problem**
Under sustained traffic or continuous batching scenarios, the `BatchedExecutor` eventually causes the native `llama.cpp` backend to deadlock, reject tokens, or throw a segmentation fault.

**The Root Cause**
Currently, `BatchedExecutor` assigns sequence IDs using a strictly incrementing counter (`_nextSequenceId++`). When a `Conversation` is disposed, its tokens are properly removed from the KV cache via `MemorySequenceRemove`, but its Sequence ID is permanently abandoned.

In `llama.cpp`, the `SeqMax` (`n_seq_max`) parameter dictates the static allocation of arrays within the native KV cache. It expects sequence IDs to act as strictly bounded array indices (from `0` to `SeqMax - 1`). Once the C# `_nextSequenceId` counter exceeds `SeqMax`, passing that ID to the native backend results in an out-of-bounds memory access.

**The Solution**
This PR replaces the strictly incrementing counter with a Sequence ID pool.

1. **`BatchedExecutor`**: Updated to track active IDs (e.g., using a thread-safe `HashSet` or similar structure) and assign the lowest available sequence ID. Added a `ReleaseSequenceId` method.
2. **`Conversation`**: Updated the `Dispose()` method to call `Executor.ReleaseSequenceId(ConversationId)` after clearing the KV cache.

**Benefits**

* **Native Memory Safety:** Guarantees that sequence IDs will never exceed the configured `SeqMax` limit as long as concurrent conversations stay within bounds.
* **Indefinite Uptime:** Allows the `BatchedExecutor` to run continuously under high-traffic workloads without requiring the host application to dangerously destroy and recreate the massive native context to reset the ID counter.

### DIFF
--- a/LLama/Batched/BatchedExecutor.cs
+++ b/LLama/Batched/BatchedExecutor.cs
@@ -14,7 +14,54 @@ namespace LLama.Batched;
 public sealed class BatchedExecutor
     : IDisposable
 {
-    private int _nextSequenceId;
+    /// <summary>
+    /// Tracks the sequence IDs currently in use by active conversations.
+    /// This pool ensures that IDs are reused and never exceed the native backend's SeqMax allocation.
+    /// </summary>
+    private readonly HashSet<int> _activeSequenceIds = new();
+
+    /// <summary>
+    /// Allocates the lowest available Sequence ID for a new conversation.
+    /// </summary>
+    /// <returns>A unique sequence ID bounded by the maximum number of concurrent active conversations.</returns>
+    /// <exception cref="InvalidOperationException">Thrown if no sequence IDs can be allocated.</exception>
+    internal LLamaSeqId GetNextSequenceId()
+    {
+        // LOCK REQUIRED: Prevent race conditions if multiple conversations are created simultaneously
+        lock (_activeSequenceIds)
+        {
+            // Linearly search for the lowest available ID.
+            // Because IDs are recycled when conversations are disposed, this will naturally 
+            // stay bounded below the host's maximum concurrency limit (SeqMax).
+            for (int i = 0; i < int.MaxValue; i++)
+            {
+                if (!_activeSequenceIds.Contains(i))
+                {
+                    _activeSequenceIds.Add(i);
+                    return (LLamaSeqId)i;
+                }
+            }
+        }
+
+        // Fallback safety (practically unreachable unless int.MaxValue concurrent users are active)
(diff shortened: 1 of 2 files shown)


LLama/Batched/BatchedExecutor.cs
```diff
@@ -14,7 +14,54 @@ namespace LLama.Batched;
...
+    /// Allocates the lowest available Sequence ID for a new conversation.
+    /// </summary>
+    /// <returns>A unique sequence ID bounded by the maximum number of concurrent active conversations.</returns>
+    /// <exception cref="InvalidOperationException">Thrown if no sequence IDs can be allocated.</exception>
+    internal LLamaSeqId GetNextSequenceId()
+    {
+        // LOCK REQUIRED: Prevent race conditions if multiple conversations are created simultaneously
+        lock (_activeSequenceIds)
+        {
+            // Linearly search for the lowest available ID.
```
martindevans, line 33: A better way to manage this could be to split the state into two parts:
 - "High water mark": The minimum seq ID that has never been used
 - "Free set": A set of unused sequences IDs which are less than the high water mark

That way you can always retrieve a new ID in O(1):
 - Remove from free set
 - If free set is empty, return "high water mark" (and increment it by one)

I don't think this is really a problem though, this search as implemented should be very fast. I'm happy to merge as is if you don't want to implement the alternative.

martindevans (approved):
Looks good, thanks for fixing this.

LLama/Batched/BatchedExecutor.cs
martindevans: (Disposing a conversation always just adds to the free set)

zsogitbe: Let us keep it simple for now.

zsogitbe:
Martin, when you merge this could you please also check the "[llama.cpp @ 73c9eb8](https://github.com/ggerganov/llama.cpp/tree/73c9eb8ceda397b651dbb6661b2935f0283a2b1d)" link in the master branch. I think you forgot to update it to the version applied with the former main update. Thanks a lot!

martindevans:
> could you please also check the [submodule]

Oops, I keep forgetting to do that. I fixed it in https://github.com/SciSharp/LLamaSharp/pull/1387

PR #1391, opened 2026-05-24
Update c0c7e147e7efa6c5858754b47259ba4880f8a906

martindevans:
Updated to llama.cpp [c0c7e147e7efa6c5858754b47259ba4880f8a906](https://github.com/ggml-org/llama.cpp/tree/c0c7e147e7efa6c5858754b47259ba4880f8a906).

Testing:
 - [x] Windows CUDA
 - [ ] Windows Vulkan
 - [ ] Linux CUDA
 - [ ] Linux Vulkan


martindevans:
This is using a set of test binaries, will re-open with final binaries soon (should be identical).

PR #1408, opened 2026-07-17
Disable GGML_F16C GGML_BMI2 for noavx buid (Fixes #1407)

### LINKED ISSUE #1407: [BUG]: Process killed by SIGILL on old SSE 4.2 only CPU
### Description

Running my app in Docker on a Synology DS920+ (Intel Celeron J4125, Gemini Lake), the process is killed by SIGILL while loading model weights, even though the `noavx` backend is selected.  

I confirmed debug log says `runtimes/linux-x64/native/noavx/libggml-cpu.so` is the library actually loaded, so this is not a library-selection problem. The `noavx` binary itself contains instructions this CPU doesn't have.

### Reproduction Steps

1. Use `NativeLibraryConfig.All.WithAvx(AvxLevel.None)` on a CPU without AVX/F16C/BMI2 (Gemini Lake / Goldmont Plus class; flags go up to SSE4.2)
2. Load any GGUF model
3. Process exits with code 132 (SIGILL). dmesg:

```
traps: .NET TP Worker[29868] trap invalid opcode ip:7f19b4d0e0bd sp:7f19b6ffb900 error:0 in libggml-cpu.so[7f19b4d04000+be000]
```

- LLamaSharp 0.27.0 + LLamaSharp.Backend.Cpu 0.27.0
- Linux x64, Docker on Synology DS920+ (Celeron J4125)
- CPU flags: SSE4.1 / SSE4.2 only (no AVX, no F16C, no FMA, no BMI1/2)
- .NET 10, `WithAvx(AvxLevel.None)`, noavx variant confirmed loaded via native load logs

### Known Workarounds

The route cause is that the `noavx` build in `.github/workflows/compile.yml` passes only `-DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF`. ggml's CMake enables `GGML_F16C` and `GGML_BMI2` by default even with `GGML_NATIVE=OFF`, so the "noavx" binary still contains F16C (VEX-encoded) and BMI2 instructions. The fp16 to fp32 conversion during weight loading is the first code that executes them.

Adding these two flags to the noavx build should fix it:

```
-DGGML_F16C=OFF -DGGML_BMI2=OFF
```

As a local workaround, building llama.cpp at the release's pinned commit with those flags and loading it via NativeLibraryConfig.All.WithLibrary(...) should avoid this.

 (But not yet tested on my side. the analysis above is based on the CI build flags and ggml's CMake defaults)

butameron:
This PR disables the GGML_F16C and GGML_BMI2 options for the noavx builds in CI.

Without these options, the "noavx" binaries were still compiled with `-mf16c -mbmi2`,
so they crashed with SIGILL on old CPUs that have no F16C/BMI2 support such as Celeron J4125.

Fixes #1407

Tested locally with the following commands:

```
mkdir build && cd build
cmake .. -DGGML_NATIVE=OFF -DLLAMA_BUILD_TESTS=OFF -DLLAMA_OPENSSL=OFF \
  -DBUILD_SHARED_LIBS=ON -DLLAMA_BUILD_UI=OFF -DLLAMA_BUILD_APP=OFF \
  -DLLAMA_BUILD_EXAMPLES=OFF -DLLAMA_BUILD_SERVER=OFF \
  -DCMAKE_INSTALL_RPATH='$ORIGIN' -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON \
  -DCMAKE_PLATFORM_NO_VERSIONED_SONAME=ON \
  -DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF -DGGML_F16C=OFF -DGGML_BMI2=OFF
cmake --build . --config Release -j $(nproc) --target ggml ggml-base ggml-cpu llama mtmd
```

This changes the flags used to compile `ggml-cpu` from `-mbmi2 -mf16c -msse4.2` to `-msse4.2`.

I confirmed this by directly overwriting the `*.so` files above to `/path/to/app/runtimes/linux-x64/native/noavx`, and it successfully runs on a Synology DS920+ Docker environment.

I haven't tested the other environments, but I added the same options to each of them for consistency.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - build: 'noavx'
-            defines: '-DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF'
+            defines: '-DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF -DGGML_F16C=OFF -DGGML_BMI2=OFF'
             os: ubuntu-22.04
             arch: x64
           - build: 'avx2'
@@ -95,7 +95,7 @@ jobs:
       matrix:
         include:
           - build: 'noavx'
-            defines: '-DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF'
+            defines: '-DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF -DGGML_F16C=OFF -DGGML_BMI2=OFF'
           - build: 'avx2'
             defines: ''
           - build: 'avx'
@@ -160,7 +160,7 @@ jobs:
       matrix:
         include:
           - build: 'noavx'
-            defines: '-DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF'
+            defines: '-DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF -DGGML_F16C=OFF -DGGML_BMI2=OFF'
           - build: 'avx2'
             defines: ''
           - build: 'avx'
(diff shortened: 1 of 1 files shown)


martindevans (approved):
Thanks for putting this together, looks good to me!

PR #1415, opened 2026-08-14
Fix broken star history chart

Dessalines39394:
The star history chart in the README is currently broken due to GitHub stargazer API restrictions. This change moves it to a working alternative provider that requires no API token, so the chart renders correctly again.


martindevans:
Thanks for fixing this.

I can see from your contribution history that you've submitted this same PR to many other projects - are you in any way associated with the new star history provider? Please disclose this if so.

Issue #1422, opened 2026-08-21
Getting a Illegal instruction (core dumped) when trying to run


blackdow1:
Getting the Illegal instruction (core dumped) when I try to run llamasharp. Here is my process of trying to run this repo: 
-dotnet new console -n LlamaSharpProject
-cd LLamaSharpProject
-dotnet add package LLamaSharp (I did not see the notice at the bottom of the readme at first, so i also went back and did the process over with --version 0.24.0 appended)
-dotnet add package LLamaSharp.Backend.Cpu (appended --version 0.24.0 on other attempt)
-dotnet build
-dotnet run (it fails silently, get no feedback from console.)
I then try dotnet bin/Debug/net10.0/LlamaSharpProject.dll and that is when i see the core dump message.

I am running this on an old laptop. These are its flags from lscpu command:
Flags:                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pg
                             e mca cmov pat pse36 clflush dts acpi mmx fxsr sse 
                             sse2 ht tm pbe syscall nx lm constant_tsc arch_perf
                             mon pebs bts rep_good nopl cpuid aperfmperf pni dte
                             s64 monitor ds_cpl est tm2 ssse3 cx16 xtpr pdcm lah
                             f_lm pti dtherm ida

I am able to run models fine using ggml's llama.cpp. I am running the Qwen3-0_6B-Q4_0
Any help would be greatly appreciated.

zsogitbe:
Hi @blackdow1, your `lscpu` flags point directly to the root of the problem.

The `Illegal instruction (core dumped)` error is happening because your laptop's CPU is a bit older and doesn't support **AVX** or **AVX2** instructions (looking at your flags, your CPU caps out at `ssse3`).

The native libraries bundled inside the `LLamaSharp.Backend.Cpu` NuGet package are pre-compiled with AVX support enabled to ensure fast performance for most modern hardware. When your CPU encounters one of those instructions, it crashes.

The reason standard `llama.cpp` works for you is likely because you built it from source on that machine, meaning your C++ compiler automatically turned off AVX instructions to match your hardware.

**How to fix this:**
To run LLamaSharp on this specific machine, you'll need to use a backend compiled without AVX.

modelpath-dev:
I'll work on this. Could you assign it to me?

I'll look at the current code, add a focused test, and send a small PR.

martindevans:
[...]

Issue #1360, opened 2026-03-21
[Feature]: Cuda 13 support


aropb:
### Background & Description

Building binary files for cuda 13.

aropb:
@martindevans  Is it possible to do this?

martindevans:
If llama.cpp supports CUDA 13 then it can probably be added to the list [here](https://github.com/SciSharp/LLamaSharp/blob/master/.github/workflows/compile.yml#L404) to produce the binaries. That would be the first step.

However, CUDA builds are _very_ slow, so I'm hesitant to add a new one. What does CUDA 13 add over our current CUDA 12 binaries?

aropb:
This is the current version of CUDA, the 12th version is already archived.

https://developer.nvidia.com/cuda-downloads

You don't even have to make nuget packages, Llamasharp Binaries are enough.

The latest version is CUDA 12 -> 12.9
https://developer.nvidia.com/cuda-12-9-1-download-archive

manutiedra:
> What does CUDA 13 add over our current CUDA 12 binaries?

CUDA 13 is a big improvement:
https://www.linkedin.com/pulse/whats-new-cuda-13-why-matters-bojan-tunguz-ph-d--e4ele

Some apps I had improvements with CUDA 13 so It will be great to have CUDA 13 supported in LlamaSharp

aropb:
Is it planned to add a build for CUDA13?
I think cuda11 can be removed.

martindevans:
This would be a good issue for a new contributor to take on. The steps involved would be:

1. All our builds are made with [this GitHub action](https://github.com/SciSharp/LLamaSharp/blob/master/.github/workflows/compile.yml). First up remove CUDA11 from there and replace it with CUDA13.

2. LL# works out with DLLs to load at runtime, replace CUDA11 with CUDA13 there. Start [here](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/Native/Load/NativeLibraryWithCuda.cs).

3. Create a nuspec file for CUDA13 (see [CUDA12](https://github.com/SciSharp/LLamaSharp/blob/master/LLama/runtimes/build/LLamaSharp.Backend.Cuda12.nuspec) nuspec), so that a nuget package gets published.

(these can be tackled as individual PRs, by different people if necessary)

modelpath-dev:
I'll take this.

[...]

PR #1438, opened 2026-09-06
Disable SSE4.2 in noavx native builds

modelpath-dev:
## Summary

- The published `noavx` CPU backend still compiles with `-msse4.2`. ggml turns `GGML_SSE42` on for a normal (non-cross) compile even when `GGML_NATIVE=OFF`.
- That is an illegal instruction on CPUs that stop at SSSE3, which is the crash in #1422. `lscpu` there has no `avx`, `sse4_1`, or `sse4_2`.
- `noavx` jobs on Linux, musl, Windows, and the macOS x64 Rosetta build now pass `-DGGML_SSE42=OFF`. gcc/clang jobs also pin `-march=x86-64` so auto-vectorization cannot put SSE4 back in. MSVC only gets the CMake flag.

## Test plan

- Configure current llama.cpp with the old noavx flags. CMake prints `Adding CPU backend variant ggml-cpu: -msse4.2 GGML_SSE42`.
- Configure with the new flags. CMake prints an empty CPU variant and `GGML_SSE42=OFF`.
- `objdump` of `libggml-cpu.so`: old build contains `pblendvb` / `pinsrd` (SSE4.1). New build does not.
- After the next Update Binaries run and a new `LLamaSharp.Backend.Cpu` publish, #1422 should load without SIGILL. Until then the FAQ notes the self-compile path.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -27,8 +27,10 @@ jobs:
       fail-fast: true
       matrix:
         include:
+          # SSE4.2 stays on unless we disable it; that SIGILLs pre-Nehalem CPUs (#1422).
+          # -march=x86-64 stops gcc from emitting SSE4 via auto-vectorization.
           - build: 'noavx'
-            defines: '-DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF -DGGML_F16C=OFF -DGGML_BMI2=OFF'
+            defines: '-DGGML_AVX=OFF -DGGML_AVX2=OFF -DGGML_FMA=OFF -DGGML_F16C=OFF -DGGML_BMI2=OFF -DGGML_SSE42=OFF -DCMAKE_C_FLAGS=-march=x86-64 -DCMAKE_CXX_FLAGS=-march=x86-64'
             os: ubuntu-22.04
             arch: x64
           - build: 'avx2'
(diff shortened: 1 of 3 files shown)


martindevans:
Thanks for putting this together, it all looks good to me.

Could you please trigger a test action run on your fork (I can't do it in this repo until it's merged).

To do that:
1. Go to "Actions" tab in your fork
2. Select "Update Binaries" in the list on the left
3. Click "Run Workflow" (top right of the action panel)
4. Link the action run here

modelpath-dev:
Triggered Update Binaries on the fork from this branch (`fix/1422-noavx-sse42`, llama.cpp `master`):

https://github.com/modelpath-dev/LLamaSharp/actions/runs/34667855880

modelpath-dev:
Update Binaries completed successfully on the fork from this branch:

https://github.com/modelpath-dev/LLamaSharp/actions/runs/34669867568

llama.cpp `d3146f2` (master at dispatch). Every compile job and Gather Binaries passed, including Linux/Windows/musl noavx with SSE4.2 off.

The first run failed on Windows ARM64 only. ClangCL + CMake PCH disagrees on C17 for ggml-cpu, so `llama.dll` was never produced. That is unrelated to the noavx flags. I pinned that job to `windows-2022` and set `CMAKE_DISABLE_PRECOMPILE_HEADERS=ON` so the ARM64 artifact builds.

martindevans (approved):
Thanks very much for working on this! Ready to merge once the CI passes :)

modelpath-dev:
@martindevans 

Hi Martin, thanks again for the thorough review on this one. Walking through the noavx native builds and the fork CI runs taught me a lot about how the binary pipeline fits  together. I would like to keep contributing to LLamaSharp, especially around native builds and packaging. 

I am currently looking for a role, so if there is ever build or maintenance work you would like a hand with, I would be glad to help. Either way, I will keep an eye on the open issues.

martindevans:
That's great, we're always in need of more contributors! If you're looking for something else to work on would you be interested in looking into https://github.com/SciSharp/LLamaSharp/issues/1360 - first of all adding a CUDA13 build, then CUDA 13 support into LL# itself?

modelpath-dev:
Thanks Martin — I took #1360 and opened https://github.com/SciSharp/LLamaSharp/pull/1439.

`compile.yml` on master was already CUDA 12.4.0 only, so I added a CUDA 13.0.0 job next to it rather than swapping 12 out, plus the loader and `Backend.Cuda13` packages. Update Binaries still has to produce the `cu13.0.0` artifacts before those packages have anything to ship. Happy to split if you'd rather land the compile job first.

PR #1439, opened 2026-09-16
feat(cuda): add CUDA 13.0.0 jobs to the compile workflow

### LINKED ISSUE #1360: [Feature]: Cuda 13 support

modelpath-dev:
## Why

Martin asked to split the CUDA 13 work so the compile workflow can land first. Maintainers can then run a Compile build and refresh the development binaries package before the managed loader or NuGet packages change.

This is part 1 of 3 for #1360.

## Scope

- `.github/workflows/compile.yml`: add `13.0.0` to the cublas matrix, use `Jimver/cuda-toolkit@v0.2.36` for non-12.4.0 installs, and gather `cu13.0.0` Windows/Linux artifacts next to `cu12.4.0`.

Out of scope here: `NativeLibraryWithCuda`, `LLamaSharp.Backend.Cuda13` nuspecs, `LLamaSharp.Runtime.targets`, and docs.

## Tradeoffs

Kept the existing `12.4.0` toolkit pin on `@v0.2.15` and only moved newer matrix values to `@v0.2.36`, so the CUDA 12 path stays unchanged while 13 can install.

## Blast Radius

Only the Compile workflow and its published native artifact layout change. No managed package or runtime load path changes until parts 2 and 3.

## Verification

- Patch-id of the full three-part tip matches the previous combined #1439 tip (`757b5dd6…`).
- This PR's diff is only `.github/workflows/compile.yml` (+33/−4).
- `dotnet` is not installed on this machine, so Compile CI on GitHub is the live check for the new matrix jobs.

Stack:

1. This PR (build)
2. Loader (opens next)
3. NuGet packaging (opens after loader)


martindevans:
Thanks for putting this together so quickly!

Would you mind spitting the PR into 3 parts:
1. Just the build changes. That way we can merge those, run a build, and update the binaries package used for development
2. The new loader. We can merge this once the binaries are available for it to use.
3. The nuget changes. We have a separate issue with our package management/ownership at the moment, we need to get that resolved before we can create new nuget packages.

Sorry for the extra work.

modelpath-dev:
Ok @martindevans, working on it

modelpath-dev:
@martindevans Split as you asked. Same code as before, three PRs:

1. This PR — compile workflow only (build / Gather Binaries for `cu13.0.0`)
2. https://github.com/SciSharp/LLamaSharp/pull/1440 — loader + unit tests
3. https://github.com/SciSharp/LLamaSharp/pull/1441 — Backend.Cuda13 nuspecs, Runtime.targets, docs

Please merge in that order. I will rebase 1440 onto master after this one lands, then 1441 after 1440.

.github/workflows/compile.yml
```diff
@@ -417,21 +417,38 @@ jobs:
...
+          use-github-cache: false
+
+      - uses: Jimver/cuda-toolkit@v0.2.36
+        if: runner.os == 'Linux' && matrix.cuda != '12.4.0'
+        id: cuda-toolkit-linux-13
+        with:
+          cuda: ${{ matrix.cuda }}
+          method: 'network'
+          linux-local-args: '["--toolkit"]'
+
```
martindevans, line 451: Is there a reason these have been split into separate stages from the existing ones? The cuda version is parameterised.

The differences I can see are:
 - `Jimver/cuda-toolkit@v0.2.15` vs `Jimver/cuda-toolkit@v0.2.36`
 - Added `use-github-cache: false`

But I'm not sure what the motivation is for these changes?

modelpath-dev: No good reason, you're right. The only real constraint was that v0.2.15 doesn't know about CUDA 13 at all (its version table stops at 12.4.1), so I kept the old pin for 12.4.0 to avoid touching that path. That was overcautious since v0.2.36 still supports 12.4.0, so it's now one step per OS on v0.2.36 for every matrix value, and I dropped `use-github-cache: false` because there was no need for it.

One thing that is actually CUDA 13 specific: on Windows the 13.x installer split `crt` and `nvvm` out of `nvcc`, so without them `cuda_runtime.h` can't find `crt/host_config.h`. I handled that with a per-version `win-sub-packages` entry in the matrix `include` so it stays a single parameterised job.

